PR #9, opened 2016-04-25
replica: switched to O_DIRECT

kp6:
based on https://github.com/rancher/sparse-tools/pull/13

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -54,10 +54,10 @@ RUN cd /usr/src/tcmu-runner && \
     cp scsi_defs.h /usr/local/include && \
     cp libtcmu_static.a /usr/local/lib/libtcmu.a
 
-# Install ssync
-RUN go get -d github.com/kp6/alphorn/ssync && \
-    cd $GOPATH/src/github.com/kp6/alphorn && \
-    git checkout c7705144deb2d398820fd248c261d2279b27f45d && \
+# Install ssync, etc
+RUN go get -d github.com/rancher/sparse-tools/ssync && \
+    cd $GOPATH/src/github.com/rancher/sparse-tools && \
+    git checkout 84fc96db3ee359d3e5675df414494c09c7d0c0c2 && \
     cd ssync && \
     go build -o /usr/local/bin/ssync
 
--- a/replica/diff_disk.go
+++ b/replica/diff_disk.go
@@ -5,7 +5,7 @@ import (
 	"os"
 
 	"github.com/frostschutz/go-fibmap"
-    fio "github.com/rancher/sparse-tools/directfio"
+	fio "github.com/rancher/sparse-tools/directfio"
 )
 
 type diffDisk struct {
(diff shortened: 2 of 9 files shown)


replica/diff_disk.go
```diff
@@ -42,7 +43,7 @@ func (d *diffDisk) WriteAt(buf []byte, offset int64) (int, error) {
 	startSector := offset / d.sectorSize
 	sectors := int64(len(buf)) / d.sectorSize
 
-	c, err := d.files[target].WriteAt(buf, offset)
+	c, err := fio.WriteAt(d.files[target], buf, offset)
```
ibuildthecloud: Can we avoid the copy and just make the original `[]byte` aligned?  The buffer is originally allocated at https://github.com/rancher/longhorn/blob/master/rpc/wire.go#L68

replica/replica.go
```diff
@@ -223,7 +224,7 @@ func (r *Replica) nextFile(parsePattern *regexp.Regexp, pattern, parent string)
 
 func (r *Replica) openFile(name string, flag int) (*os.File, error) {
 	// TODO: need to turn on O_DIRECT
-	return os.OpenFile(path.Join(r.dir, name), os.O_RDWR|os.O_CREATE|flag, 0666)
+	return os.OpenFile(path.Join(r.dir, name), fio.O_DIRECT|os.O_RDWR|os.O_CREATE|flag, 0666)
```
ibuildthecloud: I'd prefer to just use `syscall.O_DIRECT`, no reason to abstract it.

ibuildthecloud:
@kp6 Since CI is broken can you just locally run `make ci` and you'll see there are some errors.

replica/diff_disk.go
kp6: Yeah, that's what I was planning to do first but then suspected it was not worth it, tested and saw no difference as I/O masks the copy overhead completely. 
I like that all the alignment stuff is confined to directfio

replica/replica.go
kp6: fixed

Dockerfile.dapper
```diff
@@ -54,12 +54,15 @@ RUN cd /usr/src/tcmu-runner && \
     cp scsi_defs.h /usr/local/include && \
     cp libtcmu_static.a /usr/local/lib/libtcmu.a
 
-# Install ssync
-RUN go get -d github.com/kp6/alphorn/ssync && \
-    cd $GOPATH/src/github.com/kp6/alphorn && \
-    git checkout c7705144deb2d398820fd248c261d2279b27f45d && \
+# Install ssync, etc
```
ibuildthecloud: I don't think any of these changes are necessary.

Dockerfile.dapper
```diff
@@ -54,10 +54,10 @@ RUN cd /usr/src/tcmu-runner && \
...
+RUN go get -d github.com/rancher/sparse-tools/ssync && \
```
kp6: what about these?

PR #112, opened 2016-05-28
Controller: added stats instrumentation

kp6:
@ibuildthecloud @sheng-liang 
- there's no flush command support for the stats yet
- instead the stats are flushed to controller stdout automatically upon the ping error

### DIFF
--- a/backend/remote/remote.go
+++ b/backend/remote/remote.go
@@ -184,6 +184,7 @@ func (r *Remote) monitorPing(client *rpc.Client) error {
 			if err := r.Ping(); err != nil {
 				logrus.Errorf("Failed to get ping response: %v", err)
 				client.SetError(err)
+				stats.Print()
 				return err
 			}
 		}
@@ -196,7 +197,12 @@ func (r *Remote) Ping() error {
 		timeStart := time.Now()
 		resp, err := r.httpClient.Get(r.pingURL)
 		timeElapsed := time.Now().Sub(timeStart)
-		stats.Sample(timeStart, timeElapsed, 0, stats.OpPing, 0)
+		{
+			var prefix string
+			var port int
+			fmt.Sscanf(r.replicaURL, "%17s%d", &prefix, &port)
+			stats.Sample(timeStart, timeElapsed, port, stats.OpPing, 0)
+		}
 		if err != nil {
 			ret <- err
 			return
@@ -213,7 +219,6 @@ func (r *Remote) Ping() error {
 	case err := <-ret:
 		return err
 	case <-time.After(pingTimeout):
-		stats.Print()
 		return ErrPingTimeout
 	}
 }
--- a/backend/remote/remote.go
+++ b/backend/remote/remote.go
@@ -196,7 +196,7 @@ func (r *Remote) Ping() error {
 		timeStart := time.Now()
 		resp, err := r.httpClient.Get(r.pingURL)
 		timeElapsed := time.Now().Sub(timeStart)
-		stats.Sample(timeStart, timeElapsed, index, stats.OpRead, len(buf))
+		stats.Sample(timeStart, timeElapsed, 0, stats.OpPing, 0)
 		if err != nil {
 			ret <- err
 			return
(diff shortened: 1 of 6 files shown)


sheng-liang:
Can I also see all pending IO operations? You'll need to track all pending operations anyway in order to enforce timeout.

ibuildthecloud:
@kp6 I'd prefer we use https://github.com/rcrowley/go-metrics.  Also can we avoid locking.  There are points in the code where it's non-parallel like the beginning of the frontend or the rpc layer.

kp6:
OK, but what would you like to do with that info? What's the use case?

-----Original Message-----

ibuildthecloud:
https://github.com/rcrowley/go-metrics is not an answer for tracking timeouts, that must be separate.  I'd probably just do that in that as a function of the RPC layer.

sheng-liang:
For troubleshooting, it is good to know not only what operations have recently completely, but also what are the on-going operations (which might be stuck, for example.)

kp6:
@ibuildthecloud I looked the `go-metrics` but they don't seem to match what I thought we tried to achieve here:
- we want a lightweight flight data recorder (FDR) that tracks stats and overwrites the old entries 
- the FDR logs nothing unless asked after the fact
- the FDR is naturally customizable for whatever we need like pending I/O tracking

Re: the locking
- the locks are taken very briefly but I'm sure it can be improved
- I don't expect any noticeable contention or serialization effect there as the samples are typically at least 100us apart and putting the sample into the FDR takes less than 100ns (1000 times less)
- could you point me to the non-parallel code you were referring to in the frontend and the rpc layer?
 - what's the concern there?

kp6:
@sheng-liang I'll add the pending I/O tracking. I'll report current pending I/O along with the stats when they are flushed

sheng-liang:
Thanks. Tracking pending IO is probably required to implement read/write timeout anyway.

kp6:
@sheng-liang I added pending ops

ibuildthecloud:
@kp6 I guess I would say we don't need a FDR.  There's two objectives I've heard.  Handle operation timeouts and collect statistics.  For metrics, a metric library is obviously better.  For timeouts I'm not terribly sure what you have here is the best to handle that because this is out of band.  All you know is that some read/write timeout, but not the exact one.  We end up being limited on how intelligently we can deal with the timeout.

ibuildthecloud:
@kp6 I think we are going down the wrong path here.  The pending I/O commit now adds a separate data structure.  If the original sample code is not sufficient for pending I/O then should we delete it?  What does it accomplish.  The new code seems like an insert is on average `O(n/2)`.  Now I'm sure you'd say that N is always small.

I think the hang up I have here is that I personally only care about one thing.  We want to move to per operation timeouts and not just rely on the ping.  I don't understand where this code is heading to see how we can effectively handle that.

Any code around metrics handling is a waste of time because this is well address subject in the community already.  All one needs to do is use a proper library like go-metrics and instrument your code to invoke gauges, meters, counters, timers, histographs, etc.

kp6:
@ibuildthecloud the FDR is only for logging/reporting of recent and pending operations. It was not meant to serve any purpose in timeout handling except reporting upon timeout or on-demand when invoked via command line. The latter could be of form `longhorn --report-ops` or something like that. I do see your point on using something like go-metrics for the longer term. But not necessarily as a starting point. The use of go-metrics would imply somewhat different usage pattern: 
- a command line or configuration driven logging on/off control of the metrics regardless of timeout events or other errors
- is that what you have in mind?

So how about this plan:
- let me finish testing the non-ping based read/write timeout handling
- let me address your other concerns:
 - I'll move the stats instrumentation into the rpc/client
 - I'll remove the redundant `Sample` API invocation as it is a subset of the new pending op tracking
 - O(n/2) bothers me too, and it's trivial to fix
- then let's review the new PR and discuss switching to `go-metrics` as a separate issue?

would this plan work?

kp6:
this PR was replaced by #118

PR #102, opened 2016-05-24
Revert to snapshot

ibuildthecloud:

PR #110, opened 2016-05-27
Backup/restore and revert support

yasker:

PR #113, opened 2016-05-31
Revert support, includes #102

yasker:
On top of #102 
1. Add revert cmdline
2. Fix TCMU fail to restart case.

Also, I didn't clean up all the snapshots in the chain after revert, because I found out it won't be shown in `longhorn snapshot ls`, because they're not a part of live chain. If rancher is using the same mechanism to detect existing snapshots, we can safely ignore them for now - and rancher's delete snapshot commands should be no-op.


yasker:
Merged into latest #110 , for restore support.

PR #116, opened 2016-06-01
controller: read/write timeout

kp6:
- implemented "per request" read/write timeout
- replica ping is disabled

### DIFF
--- a/app/replica.go
+++ b/app/replica.go
@@ -83,16 +83,9 @@ func startReplica(c *cli.Context) error {
 	go func() {
 		server := rest.NewServer(s)
 		router := http.Handler(rest.NewRouter(server))
-		loggingHander := handlers.LoggingHandler(os.Stdout, router)
-		wrappedRouter := http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
-			if req.URL.Path == "/ping" {
-				router.ServeHTTP(rw, req)
-			} else {
-				loggingHander.ServeHTTP(rw, req)
-			}
-		})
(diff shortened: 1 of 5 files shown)


yasker:
Use sleep to count the timeout seems a little weird to me.

I think we should able to replace the wait group [here]( https://github.com/rancher/longhorn/blob/master/rpc/client.go#L59 ) for a channel, then implement the typical [go time out pattern](https://blog.golang.org/go-concurrency-patterns-timing-out-and). This way seems more straightforward to me.

kp6:
@yasker you may be right, let me check if that's the case

kp6:
@yasker let me know if that's what you had in mind, I'll squash/rebase then

yasker:
@kp6 

Yes, that's exactly what I have in mind.

Looks good to me. Thanks.

kp6:
squashed/rebased

yasker:
@kp6 

You can just get rid of all ping timeout code in the second patch. If anything goes wrong we still can easily revert back.

kp6:
@yasker per your suggestion I amended the last commit

yasker:
@kp6 

You missed a bunch of code for ping. Better take a look at https://github.com/rancher/longhorn/commit/90f7a6cedf0a3a245574ea0c6cf346926f934356 , which added ping logic.

kp6:
looking...

kp6:
@yasker amended the ping removal; thanks for catching the un-reverted pieces

yasker:
Merged, thanks.

PR #120, opened 2016-06-03
Snapshot and backup

cjellick:


cjellick:
NOTE to self: need to squash

PR #123, opened 2016-06-03
Ping retries; R/W ops with timeout retries

kp6:
This should be reviewed after 
PR https://github.com/rancher/longhorn/pull/124 - ignore the first commit for now

or use this PR unmergeable instead: https://github.com/kp6/longhorn/pull/1

### DIFF
--- a/backend/remote/remote.go
+++ b/backend/remote/remote.go
@@ -187,7 +187,7 @@ func (r *Remote) monitorPing(client *rpc.Client) error {
 				retry = 0 // reset on success
 				break
 			} else {
-				if err == ErrPingTimeout && retry < pingRetries {
+				if _, isNetErr := err.(net.Error); (isNetErr || err == ErrPingTimeout) && retry < pingRetries {
 					retry++
 					logrus.Errorf("Ping retry %v on error: %v", retry, err)
 					break
--- a/backend/remote/remote.go
+++ b/backend/remote/remote.go
@@ -19,8 +19,10 @@ import (
 )
 
 var (
-	pingTimeout    = 3 * time.Second
-	pingInveral    = 2 * time.Second
+	pingRetries = 6
+	pingTimeout = 3 * time.Second
+	pingInveral = 2 * time.Second
+
 	timeout        = 30 * time.Second
 	requestBuffer  = 1024
 	ErrPingTimeout = errors.New("Ping timeout")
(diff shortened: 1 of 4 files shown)


backend/remote/remote.go
```diff
@@ -175,12 +177,21 @@ func (r *Remote) monitorPing(client *rpc.Client) error {
...
 		select {
 		case <-r.closeChan:
 			return nil
 		case <-ticker.C:
-			if err := r.Ping(); err != nil {
+			if err := r.Ping(); err == nil {
+				retry = 0 // reset on success
+				break
+			} else {
+				if err == ErrPingTimeout && retry < pingRetries {
```
ibuildthecloud: What is the specific advantage of looking for `ErrPingTimeout`.  It we got something like a `net.Error` I think we would probably should not immediately fail.

kp6: you think of retrying on the whole net.Error class? 
- some of them are not transitive 
- but then we give up after retrying anyway...
Yeah, makes sense to me

kp6: here: c1d803073325ea199bc7a0a26b8808e811267c87

kp6: just realized that you might have meant not checking the error type at all and retrying the Ping blindly. Is that what you actually meant?

kp6: bb51738cbaee31da8dcf151b0d7555be44119cd6

kp6:
@ibuildthecloud take a look when you get a chance. 
Specifically I'd like to know whether you think it worth retrying on errors other than timeout (e.g. net.Error) for read/write ops? 

I'll squash the PR a bit after we are done

kp6:
one more thing: 
- what do you think on never timing out on the last replica? 
 - this is mostly for resource starvation scenarios like Sheng's stress test where we may end up dropping some replicas but probably should not drop the last
- and may be even including other error classes like net.Error into the "never give up last" 
- or may be just retrying more times on the last one?
- the question here is how and should we propagate an alert to the upper stack?

kp6:
@ibuildthecloud you mentioned that replicas don't support overlapping writes so I put 5b2a219ad0ebb348c74416e02f47f9e1608900a5 for now

But let's think if it makes sense to provide `write` idempotence guarantee inside replicas? 
What's your take?

PR #128, opened 2016-06-04
RW timeout retry

kp6:
Notes:
- writes are retried
- but direct block write I/O is supposed to be atomic on a block level. So concurrent 'same write' ops idempotence should still hold...

### DIFF
--- a/rpc/client.go
+++ b/rpc/client.go
@@ -10,11 +10,15 @@ import (
 )
 
 var (
+	//ErrRWTimeout r/w operation timeout
+	ErrRWTimeout = errors.New("r/w timeout")
+
+	opRetries      = 4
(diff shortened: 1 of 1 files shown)


kp6:
another thought on coexistence of ping and r/w timeouts:

-    should we disable the pings in the presence of I/O activity so that it activates only in the absence of I/O in say 2..15s period?

ibuildthecloud:
@kp6 I don't think retries should be handled in the rpc layer.  It seem like it would be far simpler to do at the replica or controller layer.  For writes we should do it at the replica https://github.com/rancher/longhorn/blob/master/replica/replica.go#L570 and for reads we can retry at the controller layer https://github.com/rancher/longhorn/blob/master/controller/replicator.go#L83.

PR #129, opened 2016-06-05
FDR: Controller R/W instrumentation

kp6:
isolated FDR from other changes

### DIFF
--- a/app/stats.go
+++ b/app/stats.go
@@ -9,9 +9,15 @@ func LsStats() cli.Command {
 	return cli.Command{
 		Name:      "ls-stats",
 		ShortName: "stats",
+		Flags: []cli.Flag{
+			cli.IntFlag{
+				Name:  "limit",
+				Value: 0,
+			},
+		},
 		Action: func(c *cli.Context) {
 			controllerClient := getCli(c)
-			err := controllerClient.ListStats()
+			err := controllerClient.ListStats(c.Int("limit"))
 			if err != nil {
 				logrus.Fatalln("Error running stats command:", err)
 			}
--- a/app/stats.go
+++ b/app/stats.go
@@ -1,0 +1,20 @@
+package app
+
+import (
+	"github.com/Sirupsen/logrus"
+	"github.com/codegangsta/cli"
+)
+
+func LsStats() cli.Command {
+	return cli.Command{
+		Name:      "ls-stats",
+		ShortName: "stats",
+		Action: func(c *cli.Context) {
+			controllerClient := getCli(c)
+			err := controllerClient.ListStats()
+			if err != nil {
+				logrus.Fatalln("Error running stats command:", err)
+			}
+		},
+	}
+}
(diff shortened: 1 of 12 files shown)


backend/remote/remote.go
```diff
@@ -168,29 +169,38 @@ func (rf *Factory) Create(address string) (types.Backend, error) {
...
+	go r.monitorPing(rpc, dataAddress /*use replica data addr for ping tracking*/)
 
 	return r, nil
 }
 
-func (r *Remote) monitorPing(client *rpc.Client) error {
+func (r *Remote) monitorPing(client *rpc.Client, targetAddr string) error {
 	ticker := time.NewTicker(pingInveral)
 	defer ticker.Stop()
+	{
```
ibuildthecloud: what's the code block for?

backend/remote/remote.go
```diff
@@ -168,29 +169,38 @@ func (rf *Factory) Create(address string) (types.Backend, error) {
...
+		// resolve target addr to match tracking of r/w ops
+		tcpAddr, err := net.ResolveTCPAddr("", targetAddr)
```
ibuildthecloud: Why resolve the backend?

controller/rest/router.go
```diff
@@ -32,5 +32,8 @@ func NewRouter(s *Server) *mux.Router {
 	router.Methods("DELETE").Path("/v1/replicas/{id}").Handler(f(schemas, s.DeleteReplica))
 	router.Methods("PUT").Path("/v1/replicas/{id}").Handler(f(schemas, s.UpdateReplica))
 
+	// Stats
+	router.Methods("POST").Path("/v1/stats").Handler(f(schemas, s.ListStats))
```
ibuildthecloud: GET is more fitting here.

rpc/client.go
```diff
@@ -126,13 +127,20 @@ func (c *Client) nextSeq() uint32 {
...
 func (c *Client) replyError(req *Message) {
+	stats.RemovePendingOp(req.ID, false)
 	delete(c.messages, req.Seq)
 	req.Type = TypeError
 	req.Data = []byte(c.err.Error())
 	req.Complete <- struct{}{}
 }
 
 func (c *Client) handleRequest(req *Message) {
+	switch req.Type {
```
ibuildthecloud: The RPC layer is all async, but higher level code like replica and control are all synchronous.  It makes more sense to annotate the code for this ops tracking at a higher level.  Then the code will be straight forward, `InsertPendingOp(); ReadAt(); RemovePendingOp()`

app/stats.go
```diff
@@ -0,0 +1,26 @@
+package app
+
+import (
+	"github.com/Sirupsen/logrus"
+	"github.com/codegangsta/cli"
+)
+
+func LsStats() cli.Command {
+	return cli.Command{
+		Name:      "ls-stats",
```
ibuildthecloud: What about journal, log, or telemetry.  Seems like we will add stats later and it will be key/values, not this log data.

kp6:
@ibuildthecloud  rebased

PR #141, opened 2016-06-09
Controller: minor tweaks 

kp6:
- incorporated Darren's feedback: journal
- and missing commit from the yesterdays merge

### DIFF
--- a/app/stats.go
+++ b/app/stats.go
@@ -5,10 +5,10 @@ import (
 	"github.com/codegangsta/cli"
 )
 
-func LsStats() cli.Command {
+// Journal flush operations since last flush
+func Journal() cli.Command {
 	return cli.Command{
-		Name:      "ls-stats",
-		ShortName: "stats",
+		Name: "journal",
 		Flags: []cli.Flag{
 			cli.IntFlag{
 				Name:  "limit",
@@ -17,9 +17,9 @@ func LsStats() cli.Command {
 		},
 		Action: func(c *cli.Context) {
 			controllerClient := getCli(c)
-			err := controllerClient.ListStats(c.Int("limit"))
+			err := controllerClient.ListJournal(c.Int("limit"))
 			if err != nil {
-				logrus.Fatalln("Error running stats command:", err)
+				logrus.Fatalln("Error running journal command:", err)
 			}
 		},
 	}
--- a/backend/remote/remote.go
+++ b/backend/remote/remote.go
@@ -16,6 +16,7 @@ import (
 	"github.com/rancher/longhorn/rpc"
 	"github.com/rancher/longhorn/types"
 	"github.com/rancher/longhorn/util"
+	journal "github.com/rancher/sparse-tools/stats"
 )
 
 var (
@@ -183,14 +184,16 @@ func (r *Remote) monitorPing(client *rpc.Client) error {
 		case <-r.closeChan:
 			return nil
 		case <-ticker.C:
-			if err := r.Ping(); err == nil {
+			if err := r.Ping(client); err == nil {
 				retry = 0 // reset on success
 			} else {
 				if retry < pingRetries {
 					retry++
-					logrus.Errorf("Ping retry %v on error: %v", retry, err)
+					logrus.Errorf("Ping retry %v on replica %v; error: %v", retry, r.replicaURL, err)
+					journal.PrintLimited(1000) //flush automatically upon retry
 				} else {
-					logrus.Errorf("Failed to get ping response: %v", err)
+					logrus.Errorf("Failed to get ping response from replica %v; error: %v", r.replicaURL, err)
+					journal.PrintLimited(1000) //flush automatically upon error
 					client.SetError(err)
 					return err
 				}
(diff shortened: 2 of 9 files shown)


backend/remote/remote.go
```diff
@@ -168,29 +169,35 @@ func (rf *Factory) Create(address string) (types.Backend, error) {
 		return nil, err
 	}
 
-	go r.monitorPing(rpc)
+	go r.monitorPing(rpc, dataAddress /*use replica data addr for ping tracking*/)
```
yasker: Weird place for comment. Move it up?

app/stats.go
```diff
@@ -5,10 +5,10 @@ import (
 	"github.com/codegangsta/cli"
 )
 
-func LsStats() cli.Command {
+func LsJournal() cli.Command {
 	return cli.Command{
-		Name:      "ls-stats",
-		ShortName: "stats",
+		Name:      "ls-journal",
```
yasker: Are we going to have any other operation on journal? Otherwise I think `ls-` prefix is not necessary.

backend/remote/remote.go
kp6: technically, the comment is just for one parameter... I can move it if you think it's better

yasker: It's better to move it IMO.

backend/remote/remote.go
```diff
@@ -168,29 +169,35 @@ func (rf *Factory) Create(address string) (types.Backend, error) {
...
 
 	return r, nil
 }
 
-func (r *Remote) monitorPing(client *rpc.Client) error {
+func (r *Remote) monitorPing(client *rpc.Client, targetAddr string) error {
```
yasker: In fact targetAddr is information available with client right? You don't need the extra parameter here.

app/stats.go
kp6: `ls-journal` is simply to match similar commands like `ls-replicas`. It's likely to be the only command for a while. But who knows?.. 
If you think it makes it's not helping than I can remove it?

backend/remote/remote.go
```diff
@@ -169,7 +169,8 @@ func (rf *Factory) Create(address string) (types.Backend, error) {
 		return nil, err
 	}
 
-	go r.monitorPing(rpc, dataAddress /*use replica data addr for ping tracking*/)
+	// use replica data addr for ping tracking
+	go r.monitorPing(rpc, dataAddress)
```
yasker: Please modify on the original commit... It's easy to do `git rebase -i`, or you can use tool like `stgit` to do it.

app/stats.go
yasker: I can't think of any other operation we will have for journal now. I suppose flush can be done automatically with you call `journal`. Unlike replica, or snapshot, we're not going to remove/add journal etc. Should be fine to remove it.

backend/remote/remote.go
kp6: It makes sense, let me see how a patch would look like

kp6:
@yasker I addressed you comments; squashed and rebased

app/stats.go
```diff
@@ -5,10 +5,11 @@ import (
 	"github.com/codegangsta/cli"
 )
 
-func LsStats() cli.Command {
+// JournalFlush flush operations since last flush
+func JournalFlush() cli.Command {
```
yasker: Why we want to call it JournalFlush()? Flush is an implicit action of calling journal. Just do Journal() sounds better because it's the main purpose of this command - showing the journal.

app/stats.go
```diff
@@ -5,10 +5,11 @@ import (
...
 	return cli.Command{
-		Name:      "ls-stats",
-		ShortName: "stats",
+		Name:      "journal-flush",
```
yasker: Just call it "journal"...

backend/remote/remote.go
```diff
@@ -162,6 +164,7 @@ func (rf *Factory) Create(address string) (types.Backend, error) {
 	}
 
 	rpc := rpc.NewClient(conn)
+	r.replicaDataAddr = conn.RemoteAddr().String()
```
yasker: In fact I was thinking in https://github.com/rancher/longhorn/blob/master/rpc/client.go#L23 , add a field for peer addr, instead of in replica.

As long as we have conn, we should able to figure peer's address. Also the same field can be by e.g. https://github.com/rancher/longhorn/blob/master/rpc/client.go#L99 .

yasker: In fact it should also be fine here. This way Ping() won't need a extra addr.

backend/remote/remote.go
```diff
@@ -37,11 +38,12 @@ type Factory struct {
...
 	types.ReaderWriterAt
-	name       string
-	pingURL    string
-	replicaURL string
-	httpClient *http.Client
-	closeChan  chan struct{}
+	name            string
+	pingURL         string
+	replicaURL      string
+	replicaDataAddr string // used as operation traget ID in journal
```
yasker: Don't need the comment here. The usage of variable can be easily checked. Also, there is a typo in the comment...

kp6:
take a look at the recent patches. If that's what you meant I'll squash/rebase

yasker:
LGTM.

kp6:
rebased

rpc/client.go
```diff
@@ -21,32 +21,39 @@ var (
...
-	wire      *Wire
-	err       error
+	end             chan struct{}
+	requests        chan *Message
+	send            chan *Message
+	responses       chan *Message
+	seq             uint32
+	messages        map[uint32]*Message
+	wire            *Wire
+	replicaDataAddr string
```
yasker: Sorry, just noticed. It's improper to call this field `replicaDataAddr`. `peerAddr` should be fine.

Will merge after this got fixed.

kp6:
renamed

PR #143, opened 2016-06-10
Coalescing fix using snapshot tree

yasker:
See https://github.com/rancher/longhorn/wiki/Snapshot-design for design.

Ready for merge


yasker:
One remaining issue need to be fixed, as well as need clean code.

PR #153, opened 2016-06-12
Ensure sync-agent up before adding replica

cjellick:
I noticed that if we do the add replica before the sync agent is up, it results in unnecessary adds and removes of replicas to the controller which was also generating lots of snapshot on the good replica


yasker:
LGTM.

Issue #121, opened 2016-06-03
Get rid of excess amount of ping log in the controller


yasker:
Logs like this floods the real useful logs.

```
::1 - - [02/Jun/2016:20:51:59 +0000] "GET /v1/volumes HTTP/1.1" 200 686
::1 - - [02/Jun/2016:20:51:59 +0000] "GET /v1/volumes HTTP/1.1" 200 686
10.42.101.144 - - [02/Jun/2016:20:51:59 +0000] "GET /v1/replicas HTTP/1.1" 200 588
10.42.101.144 - - [02/Jun/2016:20:52:00 +0000] "GET /v1/replicas HTTP/1.1" 200 588
::1 - - [02/Jun/2016:20:52:01 +0000] "GET /v1/replicas HTTP/1.1" 200 584
::1 - - [02/Jun/2016:20:52:01 +0000] "GET /v1/replicas HTTP/1.1" 200 584
::1 - - [02/Jun/2016:20:52:01 +0000] "GET /v1/replicas HTTP/1.1" 200 584
10.42.175.66 - - [02/Jun/2016:20:52:01 +0000] "GET /v1/replicas HTTP/1.1" 200 588
10.42.175.66 - - [02/Jun/2016:20:52:04 +0000] "GET /v1/replicas HTTP/1.1" 200 588
::1 - - [02/Jun/2016:20:52:04 +0000] "GET /v1/volumes HTTP/1.1" 200 686
::1 - - [02/Jun/2016:20:52:04 +0000] "GET /v1/volumes HTTP/1.1" 200 686
10.42.101.144 - - [02/Jun/2016:20:52:05 +0000] "GET /v1/replicas HTTP/1.1" 200 588
10.42.101.144 - - [02/Jun/2016:20:52:05 +0000] "GET /v1/replicas HTTP/1.1" 200 588
::1 - - [02/Jun/2016:20:52:06 +0000] "GET /v1/replicas HTTP/1.1" 200 584
::1 - - [02/Jun/2016:20:52:06 +0000] "GET /v1/replicas HTTP/1.1" 200 584
::1 - - [02/Jun/2016:20:52:06 +0000] "GET /v1/replicas HTTP/1.1" 200 584
```

kp6:
@yasker I don't seem to find where those logs are edited from... Do you happen to know by any chance?

yasker:
No, I haven't looked into them myself. Can be in the dependency package like go-rancher etc. @ibuildthecloud may knows more.

[...]

PR #176, opened 2016-06-15
controller: bypass http logging for frequent control plane verbs/URLs

kp6:
- issue #121


yasker:
Have you run an end-to-end test using Rancher for this?

Also, we have similar issue with replica too. We can fix that too.

kp6:
I've used Sheng's stress test to validate this. I can look into the replica logs too. Will put a separate PR

yasker:
Just put it in the same PR, there are small changes and similar usage.

And please setup a Rancher end-to-end test for this. I don't think stress test has called to rancher agent which caused the flooding log.

Issue #131, opened 2016-06-06
Frontend access exceed the end of volume result in replica disconnect


yasker:
Easily use rest frontend to reproduce.

Specify length = 1 and offset = volume_size.

yasker:
https://github.com/rancher/longhorn/blob/master/integration/data/test_basic_ops.py#L19

yasker:
Fix at #177

PR #177, opened 2016-06-16
controller: handle read/write beyond volume condition inside the cont…

kp6:
…roller to prevent replicas from being kicked out

- issue #131

### DIFF
--- a/controller/control.go
+++ b/controller/control.go
@@ -234,6 +234,11 @@ func (c *Controller) Start(addresses ...string) error {
 
 func (c *Controller) WriteAt(b []byte, off int64) (int, error) {
 	c.RLock()
+	if off < 0 || off+int64(len(b)) > c.size {
+		err := fmt.Errorf("EOF: Write of %v bytes at offset %v is beyond volume size %v", len(b), off, c.size)
+		c.RUnlock()
+		return 0, err
+	}
 	n, err := c.backend.WriteAt(b, off)
 	c.RUnlock()
 	if err != nil {
@@ -244,6 +249,11 @@ func (c *Controller) WriteAt(b []byte, off int64) (int, error) {
 
 func (c *Controller) ReadAt(b []byte, off int64) (int, error) {
 	c.RLock()
+	if off < 0 || off+int64(len(b)) > c.size {
+		err := fmt.Errorf("EOF: Read of %v bytes at offset %v is beyond volume size %v", len(b), off, c.size)
+		c.RUnlock()
+		return 0, err
+	}
 	n, err := c.backend.ReadAt(b, off)
 	c.RUnlock()
 	if err != nil {
--- a/integration/data/test_basic_ops.py
+++ b/integration/data/test_basic_ops.py
@@ -16,9 +16,13 @@ def test_basic_rw(dev):  # NOQA
         common.verify_data(dev, offset, data)
 
 
-# TODO BUG: https://github.com/rancher/longhorn/issues/131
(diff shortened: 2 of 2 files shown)


controller/control.go
```diff
@@ -234,6 +234,11 @@ func (c *Controller) Start(addresses ...string) error {
 
 func (c *Controller) WriteAt(b []byte, off int64) (int, error) {
 	c.RLock()
+	if off < 0 || off+int64(len(b)) > c.size {
```
yasker: Should be `>=`...

yasker:
We should also enable the integration test for this. In the same commit.

controller/control.go
kp6: yep

kp6: hold on, no, why `>=`? 
Reading 100 bytes at offset 0 is fine for a 100 byte volume

yasker: Oh yeah, you're right...

yasker: Btw, we can add this case to boundary test case as well.

kp6: ok, what's the simplest way to launch just one selected integration test?

yasker: `./scripts/integration-test -- -k test_beyond_bounary`(found I wrote a typo here)

After `make build`, you need to refresh you local binary for this to run correctly.

kp6: could you elaborate?
refresh what binary? `bin/longhorn`? or which? what do you mean by refresh?

yasker: `bin/longhorn` I meant.

if `make` failed, you won't get new `bin/longhorn` binary with your latest change. So you may need to run `make build` to get binary, and skip all the tests to get the latest binary.

yasker: Because the command I provided above is running in the host, rather inside container.

kp6: I see... so I can only ran it on rancher os then?

yasker: ... Why? How you normally build?

kp6: I build on ubuntu: `make` or `make build`

kp6: here's what I get:
```
kirill@ubuntu:~/go/src/github.com/rancher/longhorn$ ./scripts/integration-test -- -k test_beyond_bounary
longhorn: no process found
ssync: no process found
INFO[0000] Listening on localhost:9501                  
INFO[0000] Listening on control localhost:9505          
INFO[0000] Listening on data localhost:9506             
INFO[0000] Listening on data localhost:9503             
INFO[0000] Listening on control localhost:9502          
INFO[0000] Listening on sync localhost:9507             
INFO[0000] Listening on sync localhost:9504             
./scripts/integration-test: line 28: qemu-img: command not found

Comment (yasker) [inline controller/control.go]: Yeah, so why you need to run it on RancherOS?

Comment (yasker) [inline controller/control.go]: Just install the missing tools...

Comment (kp6) [inline controller/control.go]: ok, but what does this mean:
```
longhorn: no process found
ssync: no process found

kp6: the test seems to run fine despite the above though....

kp6:
enabled slightly amended test

integration/data/test_basic_ops.py
```diff
@@ -16,16 +16,20 @@ def test_basic_rw(dev):  # NOQA
         common.verify_data(dev, offset, data)
 
 
-# TODO BUG: https://github.com/rancher/longhorn/issues/131
```
yasker: Better to keep the bug here for future reference.

kp6:
amended

PR #199, opened 2016-07-11
Update vendor to latest sparse-tool commit

jimengliu:
@yasker


yasker:
I just realized this one doesn't have vendor library change attached. Please update.

PR #200, opened 2016-07-11
Update share-mnt to v0.1.2

cjellick:
Addresses #201

Had to update share-mnt to work with docker 1.11. The previous version only worked with docker 1.10 and down. The new version is also backwards compatible.


cjellick:
NOT LGTM. I need to update the launch script

Issue #204, opened 2016-07-15
Fuse frontend cannot enable inside container of Ubuntu 16.04


yasker:
It's a kernel issue.

The fuse frontends runs well on Linux v4.2, but not on v4.4 of Ubuntu 16.04

yasker:
It reports:

```
[   25.194951] blk_update_request: I/O error, dev loop2, sector 2097024
[   25.196573] blk_update_request: I/O error, dev loop2, sector 2097024
[   25.197846] Buffer I/O error on dev loop2, logical block 262128, async page read
[   25.199975] blk_update_request: I/O error, dev loop2, sector 2097136
[   25.201162] blk_update_request: I/O error, dev loop2, sector 2097136
[   25.202436] Buffer I/O error on dev loop2, logical block 262142, async page read
[   25.204683] blk_update_request: I/O error, dev loop2, sector 0
[   25.205945] blk_update_request: I/O error, dev loop2, sector 0
[   25.207117] Buffer I/O error on dev loop2, logical block 0, async page read
[   25.208463] blk_update_request: I/O error, dev loop2, sector 8
[   25.209283] Buffer I/O error on dev loop2, logical block 1, async page read
[   25.210328] blk_update_request: I/O error, dev loop2, sector 2097144
[   25.211195] Buffer I/O error on dev loop2, logical block 262143, async page read
[   25.213236] blk_update_request: I/O error, dev loop2, sector 2097144
[   25.214148] Buffer I/O error on dev loop2, logical block 262143, async page read
[   25.215448] blk_update_request: I/O error, dev loop2, sector 2097144
[   25.216219] Buffer I/O error on dev loop2, logical block 262143, async page read
[   25.217519] Buffer I/O error on dev loop2, logical block 262143, async page read
[   25.218768] Buffer I/O error on dev loop2, logical block 262143, async page read
[   25.220074] Buffer I/O error on dev loop2, logical block 262143, async page read
```

Also any operations on the device returns e.g.:
```
cat: /dev/longhorn/test-vol: Input/output error
```

yasker:
Narrow down between Ubuntu-4.4.0-5.20 and Ubuntu-4.4.0-6.21

Finally `git bisect`, found the first bad commit:
```
commit a166e6726c6e12e28ac8489ff4e2faff7065a856
Author: Seth Forshee <seth.forshee@canonical.com>
Date:   Wed Jul 2 16:29:19 2014 -0500

[...]

PR #205, opened 2016-07-15
Fuse frontend

yasker:
It's not done yet. Linux v4.4/Ubuntu 16.04 broke the mount loopback inside container for fuse(see #204).

Also since truncate hasn't been implemented, write directly to Fuse file won't work.


yasker:
Also there is a racy between delete replica and stop processing need to be fixed.

yasker:
Replica deletion racy fixed.

PR #214, opened 2016-08-05
add a test case for sfolding empty snapshot to its parent

jimengliu:
@yasker 

add a test case for sfolding empty snapshot to its parent

https://github.com/rancher/longhorn/issues/211

### DIFF
--- a/integration/core/test_cli.py
+++ b/integration/core/test_cli.py
@@ -339,6 +339,45 @@ def test_snapshot_rm(bin, controller_client, replica_client, replica_client2):
     assert chain[2] == new_chain[1]
 
 
+def test_snapshot_rm_empty(
+        bin, controller_client, replica_client, replica_client2):
+    open_replica(replica_client)
+    open_replica(replica_client2)
+
+    v = controller_client.list_volume()[0]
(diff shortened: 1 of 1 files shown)


integration/core/test_cli.py
```diff
@@ -322,21 +322,30 @@ def test_snapshot_rm(bin, controller_client, replica_client, replica_client2):
...
+    # remove the first snapshot(empty), it will fold second snapshot(empty)
+    # to the first snapshot(empty) and rename it to second snapshot
+    cmd = [bin, 'snapshot', 'rm', output1]
     subprocess.check_call(cmd)
-
     new_chain = replica_client.list_replica()[0].chain
     assert len(new_chain) == 2
     assert chain[0] == new_chain[0]
-    assert chain[2] == new_chain[1]
+    assert chain[1] == new_chain[1]
```
yasker: You need to add a new test case, not replacing the old one.

The new tests should have "more" coverage than old one, not "different" coverage.

jimengliu:
Now I separated the 2 rm snapshot tests, so it will be an additional. Please take a quick look and merge.

Issue #207, opened 2016-07-22
Direct IO support on fuse


yasker:
Currently we cannot enable `direct_io` option for fusermount, otherwise it will reports `/bin/fusermount: mount failed: Invalid argument`.

yasker:
This bug also result in sometime longhorn build process failure on `test_snapshot_revert` etc. cases. Because of the data wasn't written to the backend when snapshot was taken, which result in lost of data when revert.

yasker:
loopback directIO support only added after Linux v4.4, in:
```
commit 2e5ab5f379f96a6207c45be40c357ebb1beb8ef3
Author: Ming Lei <ming.lei@canonical.com>
Date:   Mon Aug 17 10:31:49 2015 +0800

    block: loop: prepare for supporing direct IO

    This patches provides one interface for enabling direct IO
    from user space:

        - userspace(such as losetup) can pass 'file' which is
        opened/fcntl as O_DIRECT

    Also __loop_update_dio() is introduced to check if direct I/O
    can be used on current loop setting.

    The last big change is to introduce LO_FLAGS_DIRECT_IO flag
    for userspace to know if direct IO is used to access backing
    file.

    Cc: linux-api@vger.kernel.org
    Signed-off-by: Ming Lei <ming.lei@canonical.com>
    Reviewed-by: Christoph Hellwig <hch@lst.de>
    Signed-off-by: Jens Axboe <axboe@fb.com>
```
Change for losetup as follow is in v2.28, which is NOT in Ubuntu 16.04(which contains v2.27):
```
commit 64c3bb3cb7ead2faa97aa6d400d65f392df3dba8
Author: Ming Lei <tom.leiming@gmail.com>
Date:   Tue Nov 17 22:32:47 2015 +0800

    losetup: support ioctl cmd of LOOP_SET_DIRECT_IO

    From v4.4, linux kernel starts to support direct I/O and
    AIO to backing file for loop driver, so allow losetup to
    enable the feature by using LOOP_SET_DIRECT_IO ioctl cmd.

    Signed-off-by: Ming Lei <tom.leiming@gmail.com>
```

Which means direct_io on loopback device cannot be assumed as system requirement.
1. It doesn't support before Linux v4.4
2. No distribution so far contains required losetup version to make it work.

We may able to pack the new losetup in the container, but will need kernel newer than 4.4 to support it.

yasker:
In conclusion:
[...]

PR #216, opened 2016-08-25
Fuse direct_io

yasker:
Here is the part of work done for #207, but in the end we are not able to get good enough performance on FUSE. See detail in #207. 

There is still value to merge this because of test case updated with device rw with O_DIRECT, as well we may come back in the future for revisit.

The integration test works fine with fuse now, but I don't want to risk the build on it since we stopped working on it. If we want, we can easily flip the switch to enable test with fuse.


yasker:
I will merge this PR, since it won't affect current code path. The failure of test case is a bug in master, and need to be address separately.

PR #224, opened 2016-10-21
Put log in /tmp/longhorn for build test and better teardown policy

yasker:
Which can deal with leftover from last run, if necessary.


yasker:
Close this one, will open another one without log change.

PR #267, opened 2016-12-28
Move longhorn agent and docker volume driver out to longhorn-orc

imikushin:
And update some build scripts in the process


package/Dockerfile
```diff
@@ -1,12 +1,7 @@
...
-RUN sed -i "s/http:\/\/archive.ubuntu.com\/ubuntu\//mirror:\/\/mirrors.ubuntu.com\/mirrors.txt/g" /etc/apt/sources.list && \
-	apt-get update && apt-get install -y kmod curl nfs-common fuse \
-        libibverbs1 librdmacm1 libconfig-general-perl libaio1
+RUN apt-get update && apt-get install -y kmod curl nfs-common fuse libibverbs1 librdmacm1 libconfig-general-perl libaio1
 
-RUN curl -sSL -o share-mnt https://github.com/rancher/runc/releases/download/share-mnt-v0.1.2/share-mnt && \
-    chmod u+x share-mnt && mv share-mnt /usr/bin
-
-COPY longhorn longhorn-agent launch launch-driver docker-longhorn-driver copy-binary launch-with-vm-backing-file launch-simple-longhorn /usr/bin/
+COPY longhorn launch copy-binary launch-with-vm-backing-file launch-simple-longhorn /usr/bin/
```
yasker: Need to check about `copy-binary`

package/Dockerfile
```diff
@@ -1,12 +1,7 @@
 FROM ubuntu:16.04
-RUN sed -i "s/http:\/\/archive.ubuntu.com\/ubuntu\//mirror:\/\/mirrors.ubuntu.com\/mirrors.txt/g" /etc/apt/sources.list && \
-	apt-get update && apt-get install -y kmod curl nfs-common fuse \
-        libibverbs1 librdmacm1 libconfig-general-perl libaio1
+RUN apt-get update && apt-get install -y kmod curl nfs-common fuse libibverbs1 librdmacm1 libconfig-general-perl libaio1
```
yasker: It's better to keep the first line to speed up the new build.

package/Dockerfile
imikushin: It's just a bash script used by the volume stack

PR #269, opened 2017-01-17
docker image: copy binaries to /usr/local/bin and make it a volume

imikushin:
Allows to use Longhorn binaries in Rancher storage driver via `volumes_from` compose key, vs
inheriting the binaries from Longhorn image (used as the base image). This way, Rancher storage driver and Longhorn 
can be developed in parallel.


yasker:
Where is the correlated change in longhorn-orc? I suppose this change only apply to work with the base image. 

Also could you elaborate how this change can help with development in the future?

imikushin:
@yasker this is used in the [volume stack](https://github.com/imikushin/longhorn-orc/blob/master/driver/template.go#L154)

imikushin:
This change helps avoid rebuilding longhorn-orc every time longhorn is updated

PR #300, opened 2018-05-23
Enable S3 support

JacieChao:
To support S3 backup in longhorn-engine, we need to add aws credential parameters.
- Add aws credential parameter for backup command.
- Add validate s3 command for credential validation.

### DIFF
--- a/app/backup.go
+++ b/app/backup.go
@@ -2,6 +2,7 @@ package app
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/pkg/errors"
@@ -10,6 +11,7 @@ import (
 	"github.com/rancher/backupstore/cmd"
 
 	"github.com/rancher/longhorn-engine/sync"
+	"github.com/rancher/longhorn-engine/types"
 	"github.com/rancher/longhorn-engine/util"
 )
 
@@ -83,7 +85,25 @@ func createBackup(c *cli.Context) error {
 		}
 	}
 
-	backup, err := task.CreateBackup(snapshot, dest, labels)
+	credential := map[string]string{}
+	backupType, err := util.CheckBackupType(dest)
+	if err != nil {
+		return err
+	}
+	if backupType == "s3" {
+		accessKey := os.Getenv(types.AWSAccessKey)
+		if accessKey == "" {
+			return fmt.Errorf("Missing environment variable AWS_ACCESS_KEY_ID for s3 backup")
+		}
+		secretKey := os.Getenv(types.AWSSecretKey)
+		if secretKey == "" {
+			return fmt.Errorf("Missing environment variable AWS_SECRET_ACCESS_KEY for s3 backup")
+		}
+		credential[types.AWSAccessKey] = accessKey
+		credential[types.AWSSecretKey] = secretKey
+	}
+
+	backup, err := task.CreateBackup(snapshot, dest, labels, credential)
 	if err != nil {
 		return err
 	}
--- a/sync/agent/model.go
+++ b/sync/agent/model.go
@@ -8,15 +8,16 @@ import (
 
 type Process struct {
 	client.Resource
-	ProcessType string    `json:"processType"`
-	SrcFile     string    `json:"srcFile"`
-	DestFile    string    `json:"destfile"`
-	Host        string    `json:"host"`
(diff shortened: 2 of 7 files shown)


app/backup.go
```diff
@@ -83,7 +94,20 @@ func createBackup(c *cli.Context) error {
 		}
 	}
 
-	backup, err := task.CreateBackup(snapshot, dest, labels)
+	accessKey := ""
+	secretKey := ""
+	if strings.Contains(dest, "s3") {
```
yasker: Wrong. What if the dest is `nfs://<server>/as3f`?

S3 is prefix with `s3://`

app/config.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/Sirupsen/logrus"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/urfave/cli"
+)
+
+func ValidateCmd() cli.Command {
```
yasker: Why we need this new command? Can we just send a backup list instead?

sync/agent/process.go
```diff
@@ -103,6 +103,13 @@ func (s *Server) CreateProcess(rw http.ResponseWriter, req *http.Request) error
 
 	p.ExitCode = -2
 	go func() {
+		// set aws credential
```
yasker: Put in `launch` for backup related processes. Other commands don't need those env variables.

Also enforce the parameter validity check there as well, since it's API.

app/backup.go
```diff
@@ -11,12 +11,23 @@ import (
...
 	"github.com/rancher/longhorn-engine/sync"
 	"github.com/rancher/longhorn-engine/util"
+	"strings"
 )
 
 func BackupCmd() cli.Command {
 	return cli.Command{
 		Name:      "backups",
 		ShortName: "backup",
+		Flags: []cli.Flag{
```
yasker: I don't think we need those new parameters. We should able to pick them up in the environment variable instead.

types/types.go
```diff
@@ -9,6 +9,9 @@ const (
 
 	StateUp   = State("Up")
 	StateDown = State("Down")
+
+	AccessKey = "AWS_ACCESS_KEY_ID"
```
yasker: `AWSAccessKey`, `AWSSecretKey`.

Keep in mind that we can support different protocols later, so the naming of variables shouldn't be too generic.

app/backup.go
```diff
@@ -83,7 +86,21 @@ func createBackup(c *cli.Context) error {
...
+		if accessKey == "" {
+			return fmt.Errorf("Missing environment variable AWS_ACCESS_KEY_ID for s3 backup")
+		}
+		secretKey = os.Getenv(types.SecretKey)
+		if secretKey == "" {
+			return fmt.Errorf("Missing environment variable AWS_SECRET_ACCESS_KEY for s3 backup")
+		}
+	}
+
+	backup, err := task.CreateBackup(snapshot, dest, labels, accessKey, secretKey)
```
yasker: Instead of passing two variables (and potentially more in the future), we can pass a `credential map[string]string`, and the callee can perform according to which kind of protocol is it.

e.g. 
```
credential = map[string]string{}
if s3 {
  credential[types.AWSAccessKey] = xxx
  credential[types.AWSSecretKey] = xxx
}
```

Same for the `Process` struct below.

sync/agent/model.go
```diff
@@ -17,6 +17,8 @@ type Process struct {
 	Output      string    `json:"output"`
 	Created     time.Time `json:"created"`
 	Labels      []string  `json:"labels"`
+	AccessKey   string    `json:"accessKey"`
```
yasker: See above.

app/backup.go
```diff
@@ -83,7 +85,26 @@ func createBackup(c *cli.Context) error {
 		}
 	}
 
-	backup, err := task.CreateBackup(snapshot, dest, labels)
+	var credential map[string]string
```
yasker: Just do:

```
credential := map[string]string{}
```
And remove the `credential = make(map[string]string)` later.

PR #304, opened 2018-06-06
Base Image Support

LLParse:


yasker:
So far LGTM. Waiting for the integration test update

integration/data/test_snapshot.py
```diff
@@ -85,6 +85,11 @@ def test_snapshot_revert_with_backing_file(backing_dev):  # NOQA
     before = read_dev(dev, offset, length)
     assert before != ""
 
+    info = cmd.snapshot_info()
```
yasker: Add the test to the same commit that made the change.

The other one is fine since 1. it has the unit test 2. it needs one additional patch to test correctly.

package/launch-with-vm-backing-file
```diff
@@ -1,50 +0,0 @@
-#!/bin/bash
```
yasker: Don't touch the file in the first commit then.

yasker:
Also don't touch the unused script in the first commit since we're going to remove it anyway.

LLParse:
@yasker rebased

util/util.go
```diff
@@ -236,3 +238,32 @@ func CheckBackupType(backupTarget string) (string, error) {
...
+	if fileOrDirInfo.IsDir() {
+		files, err := fileOrDir.Readdir(-1)
+		if err != nil {
+			return "", err
+		}
+		if len(files) != 1 {
+			return "", fmt.Errorf("expected exactly one file, found %d files/subdirectories", len(files))
+		}
+		if files[0].IsDir() {
+			return "", errors.New("expected exactly one file, found a subdirectory")
```
yasker: I think you missed the comment above. You don't need to import `errors` just for `errors.New` since you can use `fmt.Errorf`.

LLParse: yes, my bad

yasker:
LGTM. Merged, thanks.

PR #313, opened 2018-08-03
Change validation regexp for label values

LLParse:


util/util.go
```diff
@@ -23,6 +23,8 @@ import (
 var (
 	MaximumVolumeNameSize = 64
 	parsePattern          = regexp.MustCompile(`(.*):(\d+)`)
+	validVolumeName       = regexp.MustCompile(`^[a-zA-Z0-9][a-zA-Z0-9_.-]+$`)
+	validLabelValue       = regexp.MustCompile(`^[^\t\n\v\f\r =]+$`)
```
yasker: It's better to have a permitted list rather than a prohibited list.

How about `^[a-zA-Z0-9_.-/:]+$`? I remembered one of the concern was it wasn't going to play well if inputted as a part of a URL(so no`?&`), or on the bash command line (no ` \=`).

LLParse: Good idea. That character list appears to cover everything needed for a Docker tag.

PR #317, opened 2018-09-10
replica: error out replica if underlying file system doesn't support extent mapping

utkarshmani1997:
Not all file systems support file extent mapping, So replica
should error out if the underlying filesystem does not support
extent mapping.

This commit add check for the extent mapping support.

This can be verified if the given path to the replica is formated
into the file-system which does not give file extent map. Few of
such file-system are NFS, ZFS etc.

Signed-off-by: Utkarsh Mani Tripathi <utkarsh.tripathi@mayadata.io>

### DIFF
--- a/replica/replica.go
+++ b/replica/replica.go
@@ -14,6 +14,8 @@ import (
 	"syscall"
 
 	"github.com/Sirupsen/logrus"
+	fibmap "github.com/frostschutz/go-fibmap"
+	"github.com/pkg/errors"
 	"github.com/rancher/longhorn-engine/types"
 	"github.com/rancher/longhorn-engine/util"
 	"github.com/rancher/sparse-tools/sparse"
@@ -179,6 +181,10 @@ func construct(readonly bool, size, sectorSize int64, dir, head string, backingF
 		}
 	}
 
+	if err := r.isExtentSupported(); err != nil {
+		return nil, errors.Wrap(err, "file extent is unsupported")
+	}
+
 	r.info.Parent = r.diskData[r.info.Head].Parent
 
 	r.insertBackingFile()
@@ -213,6 +219,26 @@ func (r *Replica) diskPath(name string) string {
 	return path.Join(r.dir, name)
 }
 
+func (r *Replica) isExtentSupported() error {
+	filePath := r.diskPath(r.info.Head + metadataSuffix)
+	fileInfo, err := os.Stat(filePath)
+	if err != nil {
+		return err
+	}
+
+	file, err := os.OpenFile(filePath, os.O_RDONLY, 0666)
+	if err != nil {
+		return err
+	}
+
+	defer file.Close()
+
+	fiemapFile := fibmap.NewFibmapFile(file)
+	if _, errno := fiemapFile.Fiemap(uint32(fileInfo.Size())); errno != 0 {
+		return errno
+	}
+	return nil
+}
 func (r *Replica) insertBackingFile() {
 	if r.info.BackingFile == nil {
 		return


replica/replica.go
```diff
@@ -182,7 +182,10 @@ func construct(readonly bool, size, sectorSize int64, dir, head string, backingF
 	r.info.Parent = r.diskData[r.info.Head].Parent
 
 	r.insertBackingFile()
-
+	if err := preload(&r.volume); err != nil {
```
yasker: This error is too generic and has a side effort of preloading the volume. We can think if it's proper to preload the volume, but if we only need to deal with no file extent map support, we should try to solve that issue only.

utkarshmani1997: @yasker, i was thinking of other approach also, which is to create a temporary file and send ioctl (FIEMAP and FIBMAP) and if returns no error we can go ahead and start replica else we can error out. i'm thinking to add this code in `server.create()` function.
Shall i go with the above approach ?

utkarshmani1997: roger that.

replica/extents.go
```diff
@@ -62,3 +66,47 @@ func (u *UsedGenerator) findExtents(c chan<- int64) {
 		}
 	}
 }
+
+func createTempFile(path string) error {
```
yasker: If the file is empty but fs wasn't supported, will `Fiemap` return error? If so, you don't need to create another file. Just open `head` disk in RO and test it. https://github.com/rancher/longhorn-engine/blob/master/replica/replica.go#L177

replica/extents.go
```diff
@@ -62,3 +66,47 @@ func (u *UsedGenerator) findExtents(c chan<- int64) {
...
+	}
+	file, err := os.Open(path)
+	if err != nil {
+		return nil
+	}
+	defer file.Close()
+	fiemapFile := fibmap.NewFibmapFile(file)
+	if _, err := fiemapFile.Fiemap(uint32(fileInfo.Size())); err != 0 {
+		// verify is FIBMAP is supported in case FIEMAP failed
+		if _, err := fiemapFile.FibmapExtents(); err != 0 {
```
yasker: I don't think we use `FibmapExtents?`

utkarshmani1997:
@yasker we should check for this in `server.Start()` function because if the replica get restarted on the same node `server.Status()` will return `state: Closed` and it will not reach to the `construct()` function. So replica will not be exited.

replica/extents.go
utkarshmani1997: `Fiemap` also work on empty files, it just requires `size` to generate extents.

replica/extents.go
utkarshmani1997: ok

utkarshmani1997:
@yasker shouldn't we fatal the replica for this case?

yasker:
@utkarshmani1997 I prefer not to panic deep in the logic. The error can be passed to the upper layer and eventually caused panic. But some errors may be gracefully handled but the caller won't even know there is an error if the function just decided to panic by itself. And catch all kinds of panics in the out layer is troublesome IMO. So as a general rule here, using error instead of panic deep in the logic.

And please squash your commits into one, which should make logic sense by itself.

replica/replica.go
```diff
@@ -213,6 +218,29 @@ func (r *Replica) diskPath(name string) string {
...
+	filePath := r.diskPath(r.info.Head + metadataSuffix)
+	fileInfo, err := os.Stat(filePath)
+	if err != nil {
+		return err
+	}
+
+	file, err := os.OpenFile(filePath, os.O_RDONLY, 0666)
+	if err != nil {
+		if os.IsPermission(err) {
+			logrus.Errorf("failed to read %s, error: Read permission denied", filePath)
```
yasker: Don't need to single this out? `err` should contain the message already.

replica/replica.go
```diff
@@ -213,6 +218,29 @@ func (r *Replica) diskPath(name string) string {
...
+		}
+		return err
+	}
+
+	defer file.Close()
+
+	fiemapFile := fibmap.NewFibmapFile(file)
+	if _, err := fiemapFile.Fiemap(uint32(fileInfo.Size())); err != 0 {
```
yasker: `err != 0` implies that err is a number. Then what's `err.Error()`?

Also, if `err` is indeed a number, use `errno` instead of `err`.

utkarshmani1997: thanks for pointing this out.

replica/replica.go
utkarshmani1997: roger that

replica/replica.go
utkarshmani1997: it returns `operation not supported` and `0` if there is no error.

yasker: Dig around for a bit. Seems Fiemap should box `errno` to `error` before returning, using something similar to `errnoErr` in https://golang.org/src/syscall/syscall_unix.go. I guess we need to live with `errno != 0` for now.

replica/replica.go
```diff
@@ -179,6 +180,10 @@ func construct(readonly bool, size, sectorSize int64, dir, head string, backingF
 		}
 	}
 
+	if err := r.isExtentSupported(); err != nil {
+		return nil, err
```
yasker: Use something like: `return nil, errors.Wrap(err, "file extent is unsupported")`. So we can see what the error is clearly in the log.

Other than this, LGTM.

yasker:
It's a very good commit message, just needs a little tweak.

Can you remove following from the commit message? It's unnecessary. 
```
On branch fix-replica
Changes to be committed:
    modified:   replica/replica.go
```

Also, change the title to:
```
replica: error out replica if the underlying file system doesn't support extent mapping 
```

Thanks.

utkarshmani1997:
Thank you so much for the review as well as your guidance, learned a lot from this PR.

PR #319, opened 2018-10-13
fix(golint): update go get url of golint

utkarshmani1997:
This commit update the url of golint to be installed while running
make.

Ref: https://stackoverflow.com/questions/52784861/cannot-install-golint-package-wrong-import-path

Signed-off-by: Utkarsh Mani Tripathi <utkarsh.tripathi@mayadata.io>


yasker:
LGTM. Merged, thanks.

PR #325, opened 2018-12-05
Add command to produce coalesced base image

LLParse:


app/restore_to.go
```diff
@@ -0,0 +1,145 @@
...
+			},
+			cli.StringFlag{
+				Name:  "image-format",
+				Usage: "format of image to produce",
+				Value: DefaultImageFormat,
+			},
+			cli.StringFlag{
+				Name:  "output-file",
+				Usage: "filepath to write the resulting image to",
+				Value: "/image/base.qcow2",
```
yasker: Output to the current directory by default. `volume.qcow2`.

Check and make sure `raw` works as well.

app/restore_to.go
```diff
@@ -0,0 +1,145 @@
...
+		return err
+	}
+	s := replica.NewServer(dir, backingFile, 512)
+
+	backupURL := c.String("backup-url")
+	if backupURL == "" {
+		return fmt.Errorf("backup-url must be provided")
+	}
+
+	snapshotName := "coalesced"
```
yasker: It's a bit weird to use the word `coalesced` here. Just call it `backup`.

app/restore_to.go
```diff
@@ -0,0 +1,145 @@
...
+	if c.NArg() != 1 {
+		return fmt.Errorf("directory name is required")
+	}
+	dir := c.Args()[0]
+
+	backingFile, err := openBackingFile(c.String("backing-file"))
+	if err != nil {
+		return err
+	}
+	s := replica.NewServer(dir, backingFile, 512)
```
yasker: You should only need [this](https://github.com/rancher/longhorn-engine/pull/325/commits/df56c7e4735890aceb5351da93fef08569b3bf65#diff-09f0fd9605d3f3cf6757ff3d4d30fc9aR358)(along with some precheck and unescape, see the logic) directly from backupstore API. You don't need to use replica server.

app/restore_to.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/Sirupsen/logrus"
+	"github.com/rancher/longhorn-engine/replica"
+	"github.com/rancher/longhorn-engine/util"
+	"github.com/urfave/cli"
+)
+
+const DefaultImageFormat = "qcow2"
+
+var SupportedImageFormats = []string{
+	"rbd",
```
yasker: It would be a nightmare to test again so many formats.

I say we only support `qcow2`, and `raw` for now. We can expand the list when there are use cases.

yasker:
Also, we should add integration test for this command. Restore to raw and qcow2, with or without base image, and verify the data.

app/restore_to.go
```diff
@@ -0,0 +1,137 @@
...
+			},
+			cli.StringFlag{
+				Name:  "image-format",
+				Usage: "format of image to produce",
+				Value: DefaultImageFormat,
+			},
+			cli.StringFlag{
+				Name:  "output-file",
+				Usage: "filepath to write the resulting image to",
+				Value: "volume.qcow2",
```
yasker: "volume" + DefaultImageFormat

app/restore_to.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		return err
+	}
+
+	backupURL := c.String("backup-url")
+	if backupURL == "" {
+		return fmt.Errorf("backup-url must be provided")
+	}
+
+	backupFilepath := "backup.qcow2"
```
yasker: It should be a raw image file, as the result of `RestoreDeltaBlockBackup`

app/restore_to.go
```diff
@@ -0,0 +1,137 @@
...
+
+	if err := backupstore.RestoreDeltaBlockBackup(backupURL, backupFilepath); err != nil {
+		return err
+	}
+
+	outputFile := c.String("output-file")
+	if backingFile == nil {
```
yasker: `backingFile` wasn't used? Seems you only use the path.

replica/server.go
```diff
@@ -374,5 +374,6 @@ func (s *Server) Restore(url, name string) (err error) {
 	if err != nil {
 		return err
 	}
+
```
yasker: Extra line.

app/restore_to.go
LLParse: It became unnecessary after the last code review changes, I forgot to remove it.

yasker:
LGTM. If there isn't anything else need to be updated, can you clean the code? So I can merge it.

Thanks.

yasker:
Close in favor of #335

PR #344, opened 2019-03-22
util: Add ExecuteWithoutTimeout(). app: Fixed log issue for command restore-to-file

shuo-wu:


app/restore_to_file.go
```diff
@@ -126,11 +126,12 @@ func outputFormatSupported(desiredFormat string) bool {
 }
 
 func CheckBackingFileFormat(backingFilePath string) error {
-	output, err := exec.Command(QEMUImageBinary, "info", backingFilePath).Output()
+	output, err := exec.Command(QEMUImageBinary, "info", backingFilePath).CombinedOutput()
+	outStr := string(output)
 	if err != nil {
-		return err
+		return errors.Wrapf(fmt.Errorf(outStr), "Failed CheckBackingFileFormat %s", backingFilePath)
```
yasker: There are a few issues with this statement (and below):
1. `err` was ignored instead of passing to the caller.
2. If you need to generate an error from scratch, use `fmt.Errorf` instead of `errors.Wrapf/Wrap`.
3. The change from `cmd.Output()` to `cmd.CombinedOutput()` has side effects. It's will contain the output to stderr in the second case. If that's what's needed, it's better to be done similar to https://github.com/rancher/longhorn-manager/blob/master/util/util.go#L216

To make it easier, you can create `util.ExecuteWithoutTimeout()` and use it for the command executions here (and below), e.g. (untested)
```
func ExecuteWithoutTimeout(binary string, args ...string) (string, error) {
       var err error
       cmd := exec.Command(binary, args...)

       var output, stderr bytes.Buffer
       cmd.Stdout = &output
       cmd.Stderr = &stderr

       if err = cmd.Run(); err != nil {
               return "", fmt.Errorf("Failed to execute: %v %v, output %s, stderr, %s, error %v",
                       binary, args, output.String(), stderr.String(), err)
        }
       return output.String(), nil
}
```

app/restore_to_file.go
```diff
@@ -145,64 +146,64 @@ func CopyFile(backingFilepath, outputFile string) error {
 }
 
 func CleanupTempFiles(outputFile string, files ...string) error {
+	defer logrus.Debugf("Cleanup Temp Files %v", files)
 	outputFilePath, err := filepath.Abs(outputFile)
 	if err != nil {
-		logrus.Errorf("find absolute path for output file=%s failed: %v", outputFile, err)
-		return err
+		return errors.Wrapf(err, "Failed find absolute path for output file=%s", outputFile)
```
yasker: Lower case for the error in the `fmt.Errorf()` or `errors.Wrapf()` (`failed` vs `Failed`)

Also, `failed to find`, not `failed find`.

app/restore_to_file.go
```diff
@@ -145,64 +146,64 @@ func CopyFile(backingFilepath, outputFile string) error {
...
-			return err
+			return errors.Wrapf(err, "Failed find absolute path for tmp file=%s", file)
 		}
 		if filePath == outputFilePath {
 			continue
 		}
 		if err := os.Remove(filePath); err != nil {
 			return err
 		}
+		if _, err := os.Stat(filePath); !os.IsNotExist(err) {
```
yasker: Why adding this? After the previous line, the file should be gone.

shuo-wu: To make sure that file has been removed

yasker: Is there any case you saw the file wasn't removed and the line above didn't error out?

util/util.go
```diff
@@ -80,7 +81,7 @@ func FilteredLoggingHandler(filteredPaths map[string]struct{}, writer io.Writer,
 	return filteredLoggingHandler{
 		filteredPaths:  filteredPaths,
 		handler:        router,
-		loggingHandler: handlers.CombinedLoggingHandler(writer, router),
+		loggingHandler: handlers.LoggingHandler(writer, router),
```
yasker: Why?

util/util.go
```diff
@@ -236,6 +237,21 @@ func ExecuteWithTimeout(timeout time.Duration, binary string, args ...string) (s
 	return string(output), nil
 }
 
+func ExecuteWithoutTimeout(binary string, args ...string) (string, error) {
```
yasker: Make this a separate commit.

app/restore_to_file.go
```diff
@@ -147,62 +150,60 @@ func CopyFile(backingFilepath, outputFile string) error {
 func CleanupTempFiles(outputFile string, files ...string) error {
```
yasker: Remove the `error` as return variable since it's not useful.

app/restore_to_file.go
```diff
@@ -147,62 +150,60 @@ func CopyFile(backingFilepath, outputFile string) error {
...
-		return err
+		return errors.Wrapf(err, "failed MergeSnapshotsToBackingFile %s -> %s", snapFilepath, backingFilepath)
 	}
+	logrus.Infof("Done MergeSnapshotsToBackingFile %s -> %s", snapFilepath, backingFilepath)
 	return commitSnapshot(snapFilepath)
 }
 
 func rebaseSnapshot(snapFilepath, backingFilepath string) error {
-	output, err := exec.Command(QEMUImageBinary, "rebase", "-u", "-b", backingFilepath, snapFilepath).Output()
+	output, err := util.ExecuteWithoutTimeout(QEMUImageBinary, "rebase", "-u", "-b", backingFilepath, snapFilepath)
```
yasker: Add the logs for the starting of rebaseSnapshot and commitSnapshot as well.

app/restore_to_file.go
```diff
@@ -147,62 +150,60 @@ func CopyFile(backingFilepath, outputFile string) error {
...
-			logrus.Errorf("find absolute path for tmp file=%s failed: %v", file, err)
-			return err
+			logrus.Errorf("failed to find absolute path for tmp file=%s: %v", file, err)
+			return nil
 		}
 		if filePath == outputFilePath {
 			continue
 		}
 		if err := os.Remove(filePath); err != nil {
-			return err
```
yasker: And log the error before exit.

PR #348, opened 2019-05-01
Added new feature: restore incrementally

shuo-wu:


integration/data/utils.py
```diff
@@ -38,20 +38,20 @@ def create_backup(backup_target, snap):
     return backup
 
 
-def rm_backups(*backups):
+def rm_backups(url, *backups):
```
yasker: We can change *backups to an array and put `url` after it. It looks weird to have the empty inserted into every function.

integration/data/cmd.py
```diff
@@ -17,69 +21,83 @@ def _bin():
...
+def info(url=CONTROLLER):
+    cmd = [_bin(), '--url', url]
+    cmd = cmd + ['--debug', 'info']
     return json.loads(subprocess.check_output(cmd))
 
 
-def snapshot_create():
-    cmd = [_bin(), '--debug', 'snapshot', 'create']
+def snapshot_create(url=CONTROLLER):
+    cmd = [_bin(), '--url', url]
```
yasker:
```
def longhorn(url):
    return [_bin(), '--url', url]

def info(url=CONTROLLER):
    cmd = longhorn(url) + ['--debug', 'info']
    return json.loads(subprocess.check_output(cmd))
```

integration/data/common.py
```diff
@@ -31,9 +31,15 @@
 UPGRADE_REPLICA2 = 'tcp://localhost:9515'
 UPGRADE_REPLICA2_SCHEMA = 'http://localhost:9515/v1/schemas'
 
-CONTROLLER = "http://localhost:9501"
+STANDBY_REPLICA1 = 'tcp://localhost:9522'
+STANDBY_REPLICA1_SCHEMA = 'http://localhost:9522/v1/schemas'
+STANDBY_REPLICA2 = 'tcp://localhost:9525'
+STANDBY_REPLICA2_SCHEMA = 'http://localhost:9525/v1/schemas'
+
+STANDBY_REPLICA1_PATH = '/tmp/fix1/'
```
yasker: `fix1` is weird. How about `/tmp/standby_vol1`?

And this should be in the next commit?

integration/data/cmd.py
```diff
@@ -108,3 +108,10 @@ def restore_to_file(backup_url, backing_file='', output_file='', format='',
         cmd.append('--output-format')
         cmd.append(format)
     return subprocess.check_output(cmd)
+
+
+def restore_incre(backup_url, last_restored, url=CONTROLLER):
```
yasker: `restore_inc` is better.

integration/data/cmd.py
```diff
@@ -108,3 +108,10 @@ def restore_to_file(backup_url, backing_file='', output_file='', format='',
...
+    cmd = [_bin(), '--url', url]
```
yasker: Same as before.

integration/data/common.py
shuo-wu: vendor file commit is the 1st first. met github bug again...
In engine command, the short name of flag `--incrementally` is `-i`. Is that OK?

yasker: I think you made it `-I`, but either is OK.

yasker: Double check.

PR #349, opened 2019-05-01
Updating the Permissions of the mounted block device 

shreesha21:
The following changes entail creating the block device on the node with 0660 permission. Further, if there is an instance where the code is creating the directory that holds these block devices(/dev/longhorn), then the permissions for this directory file is set to 0755


yasker:
1. Please squash into one commit. There is only one thing for this PR to do. And change the commit name. Don't include anything like `Test permission changes` or temporary debug code.
2. Give a brief explanation why mknod 0660 doesn't work.
3. CI failed on `go fmt`.

yasker:
Dockerfile change needs to be submitted in another PR, after the PR for longhorn-engine-launcher merged.

PR #354, opened 2019-05-10
util: FindAncestorByName() check both 'dockerd' and 'containerd'

shuo-wu:
https://github.com/rancher/longhorn/issues/501

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -30,8 +30,7 @@ var (
 
 	cmdTimeout = time.Minute // one minute by default
 
-	HostProc       = "/host/proc"
-	DockerdProcess = "dockerd"
+	HostProc = "/host/proc"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


yasker:
This function is used by manager, engine and engine-launcher. Can you check if we can put it in the go-iscsi-helper then use the same function across the board?

util/util.go
```diff
@@ -304,10 +306,16 @@ func ResolveBackingFilepath(fileOrDirpath string) (string, error) {
 
 func GetInitiatorNS() string {
 	pf := iutil.NewProcessFinder(HostProc)
+	var ps *linuxproc.ProcessStatus
 	ps, err := pf.FindAncestorByName(DockerdProcess)
 	if err != nil {
-		logrus.Warnf("Failed to find dockerd in the process ancestors, fall back to use pid 1: %v", err)
-		return fmt.Sprintf("%s/1/ns/", HostProc)
+		logrus.Warnf("Failed to find dockerd in the process ancestors, then try to find containerd: %v", err)
```
yasker: Don't show warning here. Only show warning if we failed to find both.

util/util.go
```diff
@@ -304,10 +306,16 @@ func ResolveBackingFilepath(fileOrDirpath string) (string, error) {
...
+		ps, err = pf.FindAncestorByName(ContainerdProcess)
+		if err != nil {
+			logrus.Warnf("Failed to find containerd in the process ancestors, fall back to use pid 1: %v", err)
+			return fmt.Sprintf("%s/1/ns/", HostProc)
+		}
 	}
+	logrus.Infof("Found proc %v in the process ancestors", ps.Name)
```
yasker: Don't need this log

PR #358, opened 2019-05-22
Fix issue HasSnapshot() cannot find last snapshot

shuo-wu:
https://github.com/rancher/longhorn/issues/568

### DIFF
--- a/replica/backup.go
+++ b/replica/backup.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/sirupsen/logrus"
+
 	"github.com/rancher/backupstore"
 )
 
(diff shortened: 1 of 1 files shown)


replica/backup.go
```diff
@@ -77,7 +73,7 @@ func (rb *Backup) OpenSnapshot(snapID, volumeID string) error {
 
 func (rb *Backup) assertOpen(id, volumeID string) error {
```
shuo-wu: Why do we need `assertOpen()`? This function restricts that we can only operate one snapshot for each `Backup` object. 
e.g. If I use the backup object (it's `deltaOps` in backupstore) to open snapshot A (`OpenSnapshot()`), then later I could read snapshot A only (`ReadSnapshot()`), or check the existence of snapshot A only (`HasSnapshot()`). It means we cannot use this backup object to check/read snapshot B.

yasker:
Patches are LGTM. Need to find a way to test this, make sure it won't happen again.

yasker:
@shuo-wu We can test it by adding a parameter e.g. `IncrementalOnly` into DeltaBackupConfig, which makes the function errors out if that's not the case. This way we can know if a full backup or incremental backup in our tests.

yasker:
Merged this for now. The testing is tracking at #592.

PR #363, opened 2019-06-03
 Reporting Backup Progress in Longhorn Engine

shreesha21:
The following code changes modify the backup creation workflow. Earlier, backup creation was a blocking operation that happens synchronously. However, if it takes longer, the caller had to wait for longer time in order to get the response. 
Now, the create backup will only initiate the backup creation process and returns a backupID as a response upon initiating successfully. A new backup command "backup status" is introduced in longhorn engine, which will return the current progress of the backup creation in percentage. When it reaches 100%, backup URL is displayed in the output. However, if there are any errors while creating backup, the error is displayed as well.
The backup status command takes backupID as input. 
`backup status <backupID>`
The output of backup status would be in the JSON format with following fields: 
`<progress> , <backupURL> , <backupError>`

Github Issue: https://github.com/rancher/longhorn/issues/45


replica/backup.go
```diff
@@ -77,7 +93,8 @@ func (rb *Backup) OpenSnapshot(snapID, volumeID string) error {
 
 func (rb *Backup) assertOpen(id, volumeID string) error {
 	if rb.volumeID != volumeID || rb.snapshotID != id {
-		return fmt.Errorf("Invalid state volume [%s] and snapshot [%s] are open, not [%s], [%s]", rb.volumeID, rb.snapshotID, id, volumeID)
+		return fmt.Errorf("Invalid state volume [%s] and snapshot [%s] are open, not [%s], [%s]",
```
yasker: Don't need this. #363 already have a similar fix. Also, the order of snapshot ID and volume ID is wrong.

sync/rpc/rpc.proto
```diff
@@ -10,6 +10,7 @@ service SyncAgentService {
     rpc BackupRemove(BackupRemoveRequest) returns (Empty) {}
     rpc BackupRestore(BackupRestoreRequest) returns (Empty) {}
     rpc BackupRestoreIncrementally(BackupRestoreIncrementallyRequest) returns (Empty) {}
+    rpc BackupFetchProgress(BackupProgressRequest) returns (BackupProgressReply) {}
```
yasker: `BackupGetStatus`

And separate the GRPC generation code into another commit. Stub the function with `not implemented`. See #364

replica/backup.go
```diff
@@ -34,6 +38,18 @@ func NewBackup(backingFile *BackingFile) *Backup {
 	}
 }
 
+func (rb *Backup) UpdateBackupStatus(snapID, volumeID string, progress int, url string, errString string) error {
+	id := GenerateSnapshotDiskName(snapID)
```
yasker: We need a lock to access rb fields now.

sync/backup.go
```diff
@@ -71,11 +74,28 @@ func (t *Task) createBackup(replicaInController *rest.Replica, snapshot, dest, v
...
-	backup, err := repClient.CreateBackup(snapshot, dest, volumeName, labels, credential)
+	backupID, err := repClient.CreateBackup(snapshot, dest, volumeName, labels, credential)
 	if err != nil {
 		logrus.Errorf("Failed backing up %s on %s to %s", snapshot, replicaInController.Address, dest)
 		return "", err
 	}
+
+	var backup, backupErr string
+
+	for i := 0; i < BackupStatusRetryAttempts; i++ {
```
yasker: We should convert to asynchronized process, right? And only very small backups can finish in 300 seconds.

The caller needs to change the way of making backup. It will use the backupID to query the backup status from time to time. `longhorn snapshot backup` shouldn't be a blocking call.

sync/rpc/server.go
```diff
@@ -180,7 +182,96 @@ func (s *SyncAgentServer) ReceiverLaunch(ctx context.Context, req *ReceiverLaunc
...
+}
+
+func openBackingFile(file string) (*replica.BackingFile, error) {
+	if file == "" {
+		return nil, nil
+	}
+	return nil, fmt.Errorf("backing file is not supported")
+}
+
+func doCreateBackup(req *BackupCreateRequest) (string, *replica.Backup, error) {
```
yasker: Don't copy the code. Refactor the code in the `/backup/` and make a function call there.

sync/rpc/server.go
```diff
@@ -194,41 +285,44 @@ func (*SyncAgentServer) BackupCreate(ctx context.Context, req *BackupCreateReque
 			os.Setenv(types.AWSSecretKey, credential[types.AWSSecretKey])
 			os.Setenv(types.AWSEndPoint, credential[types.AWSEndPoint])
 		} else if os.Getenv(types.AWSAccessKey) == "" || os.Getenv(types.AWSSecretKey) == "" {
-			return nil, errors.New("Could not backup to s3 without setting credential secret")
+			return nil, errors.New("could not backup to s3 without setting credential secret")
```
yasker: Separate commit with other typo fixes.

sync/rpc/server.go
```diff
@@ -194,41 +285,44 @@ func (*SyncAgentServer) BackupCreate(ctx context.Context, req *BackupCreateReque
...
+		return nil, fmt.Errorf("bad request: empty backup name")
 	}
 
-	logrus.Infof("Done running %s %v, returns %v", "sbackup", cmd.Args, reply.Backup)
+	obj, present := s.backupList.Load(req.Backup)
+	if present == false {
+		return nil, fmt.Errorf("backup name[%v] not found", req.Backup)
+	}
+
+	progress := obj.(*replica.Backup).BackupProgress
```
yasker: Check for type assertion result. Otherwise, it can trigger NPE.

replica/backup.go
shreesha21: I didn't understand. Are you saying we don't need this function?
I just inserted a new line for arguments

yasker: I meant #358 . Yeah, we don't need this since #358 already has a similar fix.

replica/backup.go
```diff
@@ -87,7 +90,8 @@ func (rb *Backup) ReadSnapshot(snapID, volumeID string, start int64, data []byte
 	if err := rb.assertOpen(id, volumeID); err != nil {
 		return err
 	}
-
+	rb.lock.Lock()
```
yasker: Why is the lock not covering `rb.assertOpen`?

vendor.conf
```diff
@@ -19,7 +19,7 @@ github.com/go-ini/ini                   afbd495e5aaea13597b5e14fe514ddeaa4d76fc3
 github.com/jmespath/go-jmespath         c01cf91b011868172fdcd9f41838e80c9d716264
 github.com/rancher/go-iscsi-helper      789435a8b666193489219424de752e0727e44664
 github.com/c9s/goprocinfo               0010a05ce49fde7f50669bc7ecda7d41dd6ab824
-github.com/rancher/backupstore          cf505f7d66af5197923e0cfd11ae2a8ea1270e55
+github.com/rancher/backupstore          dc98ada1e67ef972872a0d2937357c157ec5418d
```
yasker: Update the backupstore to the latest version

sync/rpc/server.go
```diff
@@ -232,6 +232,11 @@ func (*SyncAgentServer) BackupCreate(ctx context.Context, req *BackupCreateReque
 	return reply, nil
 }
 
+func (s *SyncAgentServer) BackupGetStatus(ctx context.Context, req *BackupProgressRequest) (*BackupProgressReply, error) {
+	//not implemented
+	return nil, nil
```
yasker: Don't return nil error if it's not implemented.

replica/backup.go
shreesha21: Earlier I had locks inside assertOpen and I had made this change thinking that I should not have a lock inside a lock. Later I removed locks inside assertOpen as it will be eventually removed from shuo's commit. 

As you pointed out, for this commit, we need the lock to cover rb.assertOpen. Will fix this

controller/rest/volume.go
```diff
@@ -237,3 +238,47 @@ func (s *Server) FinishRestoreVolume(rw http.ResponseWriter, req *http.Request)
...
+
+	id := mux.Vars(req)["backupID"]
+
+	ip, err := s.c.FetchReplicaIP(id)
+	if err != nil {
+		logrus.Errorf("Failed to create backup:%v", err)
+		rw.WriteHeader(http.StatusNotFound)
+		return err
+	}
+	logrus.Infof("Successfully fetched Replica IP:%v for backup %v", ip, id)
```
yasker: Remove the log. It's too trivial.

controller/rest/router.go
```diff
@@ -43,6 +43,10 @@ func NewRouter(s *Server) *mux.Router {
 	router.Methods("POST").Path("/v1/volumes/{id}").Queries("action", "preparerestore").Handler(f(schemas, s.PrepareRestoreVolume))
 	router.Methods("POST").Path("/v1/volumes/{id}").Queries("action", "finishrestore").Handler(f(schemas, s.FinishRestoreVolume))
 
+	//Backups
+	router.Methods("POST").Path("/v1/backupReplicaMap").Handler(f(schemas, s.CreateBackupMap))
```
yasker: `/v1/backupreplicamaps`
`/v1/backupreplicamaps/{backupID}`

controller/rest/model.go
```diff
@@ -98,6 +98,16 @@ type FinishRestoreInput struct {
 	CurrentRestored string `json:"currentRestored"`
 }
 
+type BackupMapping struct {
```
yasker: BackupReplicaMap

replica/backup.go
```diff
@@ -98,19 +125,22 @@ func (rb *Backup) ReadSnapshot(snapID, volumeID string, start int64, data []byte
...
 	id := GenerateSnapshotDiskName(snapID)
-	if rb.volumeID == "" {
-		return nil
-	}
-
 	if err := rb.assertOpen(id, volumeID); err != nil {
 		return err
 	}
 
+	rb.lock.Lock()
```
yasker: You cannot do this. assertOpen() needs to hold the lock. Between assertOpen and getting the lock again, you would have a race condition, e.g. rb.volumeID or snapshotID can change then makes the assertOpen check useless.

sync/rpc/server.go
```diff
@@ -232,10 +232,17 @@ func (s *SyncAgentServer) BackupGetStatus(ctx context.Context, req *BackupProgre
 		return nil, fmt.Errorf("invalid return value from sync.map")
 	}
 
+	snapshotName, err := replica.GetSnapshotNameFromDiskName(replicaObj.SnapshotID)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't get snapshot name: %v", err)
+	}
+	fmt.Println("Shreesha: Fetching Snapshot from replica Object", snapshotName)
```
shreesha21: Accidentally got added. Will remove this and raise another PR

controller/rest/volume.go
```diff
@@ -237,3 +238,62 @@ func (s *Server) FinishRestoreVolume(rw http.ResponseWriter, req *http.Request)
...
+	apiContext := api.GetApiContext(req)
+	var input BackupReplicaMap
+
+	if err := apiContext.Read(&input); err != nil {
+		logrus.Errorf("Failed to read input:%v", err)
+		return err
+	}
+
+	if err := s.c.CreateBackupMapping(input.BackupID, input.ReplicaIP); err != nil {
+		logrus.Errorf("Failed to create backup:%v", err)
```
shuo-wu: wrong log message. same below

app/backup.go
```diff
@@ -53,6 +54,46 @@ func BackupCreateCmd() cli.Command {
...
+
+	//Fetch backupObject using the replicaIP
+	task := sync.NewTask(c.GlobalString("url"))
+
+	backupStatus, err := task.FetchBackupStatus(backupID, replicaIP)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println(string(backupStatus))
```
shuo-wu: remove it or use `logrus`

integration/data/cmd.py
```diff
@@ -57,10 +58,25 @@ def snapshot_purge(url=CONTROLLER):
     return subprocess.check_call(cmd)
 
 
+def backup_status(backupID, url=CONTROLLER):
+    output = ""
+    cmd = [_bin(), '--url', url, 'backup', 'status', backupID]
+    for x in range(3600):
```
shuo-wu: Can we use `common.RETRY_COUNTS`? The backup size in our integration test is limited hence we don't need to retry 3600 times.

sync/rpc/rpc.proto
```diff
@@ -10,6 +10,7 @@ service SyncAgentService {
     rpc BackupRemove(BackupRemoveRequest) returns (Empty) {}
     rpc BackupRestore(BackupRestoreRequest) returns (Empty) {}
     rpc BackupRestoreIncrementally(BackupRestoreIncrementallyRequest) returns (Empty) {}
+    rpc BackupGetStatus(BackupProgressRequest) returns (BackupProgressReply) {}
```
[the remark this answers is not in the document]
shreesha21: BackupGetStatus was suggested by Sheng in one of the previous comments and hence I modified it.

app/backup.go
shreesha21: This is required as the output of the command executed by engine will be read by manager periodically every 5 seconds.

controller/rest/volume.go
shreesha21: Removed

controller/rpc/pb/controller.proto
```diff
@@ -23,6 +23,11 @@ service ControllerService {
     rpc ReplicaPrepareRebuild(ReplicaAddress) returns (ReplicaPrepareRebuildReply);
     rpc ReplicaVerifyRebuild(ReplicaAddress) returns (ControllerReplica);
 
+    rpc BackupReplicaMapCreate(BackupReplicaMap) returns (google.protobuf.Empty);
+    rpc ReplicaAddressGet(ReplicaAddressRequest) returns (ReplicaAddressResponse);
```
yasker: `BackupReplicaMapGet`, change request and response name as well. Same above.

Please pay close attention to the naming pattern.

controller/client/controller_client.go
```diff
@@ -473,3 +473,84 @@ func (c *ControllerClient) Check() error {
 
 	return nil
 }
+
+func (c *ControllerClient) StoreBackupReplicaMapping(backupID string, replicaAddress string) error {
```
yasker: This is totally a random name. We cannot know which RPC function this one going to call by looking at the name. Please keep the name consistent. Same below. Use the same name as the server defined in the client.

controller/rpc/pb/controller.proto
```diff
@@ -23,6 +23,11 @@ service ControllerService {
...
+    rpc BackupGet(google.protobuf.Empty) returns (BackupGetResponse);
```
yasker: It's a list, not get.

Also, it's not `get backup`, it just `get backupID`.

You probably just do `BackupReplicaMapList` and return a map contains all the backup IDs and replica IPs.

controller/rpc/pb/controller.proto
```diff
@@ -23,6 +23,11 @@ service ControllerService {
...
+    rpc BackupDelete(BackupDeleteRequest) returns (google.protobuf.Empty);
```
yasker: It's not `delete a backup`. It deletes a backup ID from controller's backup replica address map.

`BackupReplicaMapDelete`.

sync/rpc/server.go
```diff
@@ -26,6 +31,12 @@ type SyncAgentServer struct {
 	endPort         int
 	processesByPort map[int]string
 	backupList      sync.Map
+	backupIDs       BackupIDList
+}
+
+type BackupIDList struct {
+	completedBackups ring.Ring
+	onGoingBackups   ring.Ring
```
yasker: onGoingBackups shouldn't be a ring buffer.

yasker: The backupStatus (backupList + backupIDs) should have a defined CRUD interface to keep at most 5 completed backups + unlimited ongoing backup status.

```
Create()
Get()
Update()
Delete()
```

And this struct needs to be unit tested separately to make sure it's implemented correctly. You should able to implement it using a slice (array). No need for `golang-ring`.

The slice has the ordered elements. In fact, since the array should be small, we don't even need a map. Just iterate the array to find the object. You would need an RWMutex for the array.

sync/rpc/server.go
```diff
@@ -18,14 +18,29 @@ import (
...
 type SyncAgentServer struct {
 	sync.Mutex
 
 	currentPort     int
 	startPort       int
 	endPort         int
 	processesByPort map[int]string
-	backupList      sync.Map
+
+	backup *Backup
```
yasker: BackupList, not backup. It's not only for a single backup.

sync/rpc/server.go
```diff
@@ -209,9 +228,11 @@ func (s *SyncAgentServer) BackupCreate(ctx context.Context, req *BackupCreateReq
 	reply := &BackupCreateReply{
 		Backup: backupID,
 	}
-	fmt.Fprintf(os.Stdout, reply.Backup)
 
-	s.backupList.Store(backupID, replicaObj)
+	s.backup.BackupCreate(backupID, replicaObj)
+	if err := s.backup.BackupUpdate(); err != nil {
```
yasker: Caller shouldn't need to remember calling `BackupUpdate` after BackupCreate.  It should be done automatically since there is no situation that you won't call `BackupUpdate` after BackupCreate.

Also, call it `backuplist.BackupAdd()`, since you're not creating a new object. You just pass some information in.

And call BackupUpdate() `Refresh()` since you're not updating the object. It's updating the list.

And the `Refresh()` function shouldn't return any error because it's all in memory. Unless there is a bug, there shouldn't be any error.

sync/rpc/server.go
```diff
@@ -209,9 +228,11 @@ func (s *SyncAgentServer) BackupCreate(ctx context.Context, req *BackupCreateReq
...
+		logrus.Errorf("failed to cleanup old backups:%v", err)
```
yasker: Space after `:`.

sync/rpc/server.go
```diff
@@ -308,3 +328,91 @@ func (*SyncAgentServer) BackupRestoreIncrementally(ctx context.Context, req *Bac
 	logrus.Infof("Done running %s %v", "sbackup", cmd.Args)
 	return &Empty{}, nil
 }
+
+//The following APIs implement the CRUD interface for the backup object
```
yasker: Space after `//`. And this should fail the `go lint`? Because the comment of one function should start with `// BackupCreate xxxx`.

sync/rpc/server.go
```diff
@@ -308,3 +328,91 @@ func (*SyncAgentServer) BackupRestoreIncrementally(ctx context.Context, req *Bac
...
+// The slice backuplist is implemented similar to a FIFO queue.
+// Insertion of new BackupInfo object happens in the rear end of backuplist
+// BackupUpdate deletes all the old completed backups from the front
+// BackupDelete will delete the entry with the corresponding backupID
+
```
yasker: Remove the empty line.

sync/rpc/server.go
```diff
@@ -308,3 +328,91 @@ func (*SyncAgentServer) BackupRestoreIncrementally(ctx context.Context, req *Bac
...
+		backupID:     backupID,
+		backupStatus: backup,
+	})
+}
+
+func (b *Backup) BackupGet(backupID string) (*replica.Backup, error) {
+	if backupID == "" {
+		return nil, fmt.Errorf("empty backupID")
+	}
+	b.Lock()
```
yasker: RLock().

sync/rpc/server.go
```diff
@@ -308,3 +328,91 @@ func (*SyncAgentServer) BackupRestoreIncrementally(ctx context.Context, req *Bac
...
+	defer b.Unlock()
+
+	b.backupList = append(b.backupList, &BackupInfo{
+		backupID:     backupID,
+		backupStatus: backup,
+	})
+}
+
+func (b *Backup) BackupGet(backupID string) (*replica.Backup, error) {
+	if backupID == "" {
```
yasker: Inconsistent. You didn't check backupID for create but check it for get.

sync/rpc/server.go
```diff
@@ -308,3 +328,91 @@ func (*SyncAgentServer) BackupRestoreIncrementally(ctx context.Context, req *Bac
...
+}
+
+func (*Backup) remove(b []*BackupInfo, index int) ([]*BackupInfo, error) {
+	if index >= len(b) || index < 0 {
+		return nil, fmt.Errorf("index out of range")
+	}
+	return append(b[:index], b[index+1:]...), nil
+}
+
+func (b *Backup) BackupUpdate() error {
```
yasker: As I mentioned before, no need to return an error, and there is no error return by the function.

shreesha21: Error is being returned inside the function at line 392. 
Should we account for the fact that if there are any issue encountered, say indices go out of range when I call to delete from slice ?
From manual testing, I didn't see any such issues, but it's necessary to have them in the code right?

sync/rpc/server.go
shreesha21: This line is removed as refresh doesn't return error

sync/rpc/server.go
```diff
@@ -308,3 +324,98 @@ func (*SyncAgentServer) BackupRestoreIncrementally(ctx context.Context, req *Bac
...
+}
+
+// remove deletes the object present at slice[index] and returns the remaining elements of slice yet maintaining
+// the original order of elements in the slice
+func (*Backup) remove(b []*BackupInfo, index int) []*BackupInfo {
+	if b == nil {
+		return nil
+	}
+	if index >= len(b) || index < 0 {
+		logrus.Errorf("BUG: index out of range")
```
yasker: You're right. I missed error handling here. Yes, we need error for out of range operation though it would be treated as a bug. Make sure you provide enough information in the error message for that. This one doesn't have enough information.

sync/rpc/server.go
```diff
@@ -18,14 +18,29 @@ import (
...
 	currentPort     int
 	startPort       int
 	endPort         int
 	processesByPort map[int]string
-	backupList      sync.Map
+
+	BackupList *Backup
+}
+
+type Backup struct {
```
yasker: Structure name change to backuplist too.

sync/rpc/server.go
```diff
@@ -18,14 +18,29 @@ import (
...
+	sync.RWMutex
+	backupList []*BackupInfo
```
yasker: backups []*BackupInfo

sync/rpc/server.go
shreesha21: So should I change back the return value of Refresh function to error? Or should we just report it in logrus.Errorf?

yasker: Change it back to error I meant. Log is not enough and ignore this will cause trouble down the line.

PR #369, opened 2019-06-18
Add Lock for Purging Snapshots

ttpcodes:
This PR adds a new state to the `Controller` called `isPurging` that keeps tracks of whether or not there is currently a `longhorn snapshots purge` command in progress. The `gRPC` API definitions have been updated accordingly, and the `longhorn` CLI commands have been updated to have to call `PreparePurge` and `FinishPurge` functions when running `longhorn snapshot purge`.

If the state on the `Controller` doesn't allow for the client to run another concurrent `longhorn snapshots purge` or is otherwise inconsistent, the CLI will report an error now instead of attempting to simultaneously purge snapshots, which was the cause of the race condition described in longhorn/longhorn#581.


controller/rpc/pb/controller.proto
```diff
@@ -38,8 +40,9 @@ message Volume {
     string endpoint = 3;
     string frontend = 4;
     string frontendState = 5;
-    bool isRestoring = 6;
-    string lastRestored = 7;
+    bool isPurging = 6;
```
yasker: Add the new variable to the end. Same for the other places.

ttpcodes: Didn't see the notification for the review until now. Will fix variable ordering.

ttpcodes:
All of the new definitions related to the `Snapshot` purge lock have been moved after existing variables instead of in between them. Additionally, the branch has been rebased onto `master` to resolve the conflicts in the `gRPC` definitions.

yasker:
This is not enough. If the purge failed for any reason, there is no way to restart the purge.

yasker:
Revert this PR since we're going to moving the purging operation to the sync agent server in the future.

PR #372, opened 2019-06-20
Configure Drone to publish longhornio/longhorn-engine image at push event.

meldafrawi:


Dockerfile.dapper
```diff
@@ -10,7 +10,7 @@ ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH}
 ENV PATH /go/bin:$PATH
 ENV DAPPER_DOCKER_SOCKET true
 ENV DAPPER_ENV TAG REPO
-ENV DAPPER_OUTPUT bin
+ENV DAPPER_OUTPUT package
```
yasker: No, you cannot change this. All the binaries generated by building the software is stored in `./bin`, and there are scripts depends on this assumption.

Issue #263, opened 2016-12-21
Sometime snapshot cannot be created during rebuild_test.py


yasker:
it seems due to a false remaining snapshot count = 0

```
time="2016-12-21T23:36:28Z" level=info msg="Starting with replicas [\"tcp://172.18.1.3:9502\" \"tcp://172.18.2.3:9502\"]" 
time="2016-12-21T23:36:28Z" level=info msg="Connecting to remote: 172.18.1.3:9502" 
time="2016-12-21T23:36:28Z" level=info msg="Opening: 172.18.1.3:9502" 
time="2016-12-21T23:36:28Z" level=info msg="Adding backend: tcp://172.18.1.3:9502" 
time="2016-12-21T23:36:28Z" level=info msg="Connecting to remote: 172.18.2.3:9502" 
time="2016-12-21T23:36:28Z" level=info msg="Opening: 172.18.2.3:9502" 
time="2016-12-21T23:36:28Z" level=info msg="Adding backend: tcp://172.18.2.3:9502" 
time="2016-12-21T23:36:32Z" level=info msg="SCSI device /dev/sdd created" 
time="2016-12-21T23:36:32Z" level=info msg="Device /dev/longhorn/vol3 is ready" 
time="2016-12-21T23:36:32Z" level=info msg="Listening on localhost:9501" 
time="2016-12-21T23:36:32Z" level=error msg="Error reading from wire: Error with read Seq: EOF" 
time="2016-12-21T23:36:33Z" level=error msg="Error in request: Too many snapshots created"
```

PR #374, opened 2019-06-25
Restore Progress in Engine

shreesha21:
The following code changes implement the asynchronous way of Restoring backup. 
Restore is initiated first and then the progress is checked with the help of Restore Backup status command. The commits individually detail the code changes in them.

Issue: https://github.com/longhorn/longhorn/issues/49


replica/server.go
```diff
@@ -354,10 +357,37 @@ func (s *Server) Restore(url, name string) (err error) {
...
+		DeltaOps:     restoreObj,
+		SnapshotName: toFile,
+	}
+
+	if err := backupstore.RestoreDeltaBlockBackup(config); err != nil {
+		logrus.Errorf("failed to initiate Restore backup for url[%v]: %v", backupURL, err)
 		return err
 	}
 
+	for j := 0; j < RetryAttempts; j++ {
```
yasker: There are two issues with this approach.

1. This operation should be an async process. The caller shouldn't be blocked once the restoration started.
2. The polling mechanism cannot be time-based. It's the biggest issue of the previous approach. Because we cannot know how long a backup or restoration will take, so we cannot say for sure it would be done in 1800 seconds. With progress, what we can do instead, is there should be updated in e.g. 5 minutes. It shouldn't be just percentage change because it's 100 percent at most. Seems we would need a timestamp on when is last time status was updated, as long as it has been updated at least once in e.g. 5 minutes, we know we're making progress.

app/restore_to_file.go
```diff
@@ -59,6 +61,53 @@ func RestoreToFileCmd() cli.Command {
...
+		DeltaOps:     restoreObj,
+		SnapshotName: BackupFilePath,
+	}
+
+	err := backupstore.RestoreDeltaBlockBackup(config)
+	if err != nil {
+		return err
+	}
+
+	for j := 0; j < 600; j++ {
```
yasker: Same here as in the previous case.

app/restore_to_file.go
```diff
@@ -59,6 +61,53 @@ func RestoreToFileCmd() cli.Command {
...
+	r.Lock()
+	defer r.Unlock()
+
+	logrus.Errorf("Updating progress[%v], snapshot[%v], error[%v]", rp, snapshot, re)
+	r.SnapshotName = snapshot
+	r.RestoreProgress = rp
+	r.RestoreError = re
+}
+
+func checkRestore(backupURL string) error {
```
yasker: It's still `restore`, not `checkRestore`.

integration/data/cmd.py
```diff
@@ -75,6 +75,26 @@ def backup_status(backupID, url=CONTROLLER):
     return output
 
 
+def restore_status(backup, url=CONTROLLER):
```
yasker: It's restore and wait for complete, not just `restore_status`.

app/restore_to_file.go
shreesha21: Restore-to-file calls backupstore.RestoreDeltaBlockBackup(config) directly. I tried to make it asynchronous but when the restore is happening in the background, the process that initiated it terminates. 
```
$ docker exec 1d8d057cb2b8 longhorn --url http://localhost:9501 backup restore-to-file vfs:///data/backupstore?backup=backup-f571a76c907d4463\u0026volume=vol-name
time="2019-06-27T00:19:31Z" level=info msg="Running restore to file command: backup-url=vfs:///data/backupstore?backup=backup-f571a76c907d4463u0026volume=vol-name backing-file= output-file= output-format=qcow2"
time="2019-06-27T00:19:31Z" level=info msg="Output file path=/volume.qcow2"
time="2019-06-27T00:19:31Z" level=info msg="Start to restore vfs:///data/backupstore?backup=backup-f571a76c907d4463u0026volume=vol-name to backup.img"
time="2019-06-27T00:19:31Z" level=info msg="Done running restore to file command. Produced image: "
```

Do we have to keep it waiting, in that case, why not keep this a blocking operation?

yasker: The restore to file is a blocking operation, but we need to show the user the progress and don't timeout based on the time only. We need to timeout based on the progress as in the other cases.

We can use e.g. https://github.com/cheggaaa/pb to show the progress bar in the console.

shreesha21: This progress bar package seems to have lots of issues. Couldn't compile it as the package has many dependency errors. Will just use the timer logic in the restore_to_file.

yasker: You need to add those dependency in the vendor.conf and regenerate dependency use trash.

yasker: and use v1 for now. v2 is in beta.

sync/rpc/server.go
```diff
@@ -282,27 +283,77 @@ func (*SyncAgentServer) BackupRemove(ctx context.Context, req *BackupRemoveReque
 	return &Empty{}, nil
 }
 
-func (*SyncAgentServer) BackupRestore(ctx context.Context, req *BackupRestoreRequest) (*Empty, error) {
-	cmd := reexec.Command("sbackup", "restore", req.Backup, "--to", req.SnapshotFileName)
-	cmd.SysProcAttr = &syscall.SysProcAttr{
-		Pdeathsig: syscall.SIGKILL,
+func (s *SyncAgentServer) anyActiveRestoreInProgress() bool {
```
yasker: `isRestorationInProgress`

sync/rpc/server.go
```diff
@@ -282,27 +283,77 @@ func (*SyncAgentServer) BackupRemove(ctx context.Context, req *BackupRemoveReque
...
+	}
 
-	logrus.Infof("Done running %s %v", "sbackup", cmd.Args)
+	rObj, err := backup.DoBackupRestore(req.Backup, req.SnapshotFileName)
+	if err != nil {
+		return nil, fmt.Errorf("error initiating restore [%v]", err)
+	}
+	logrus.Infof("Successfully initiated restore for snapshot [%v]", rObj.SnapshotName)
+
+	s.Lock()
```
yasker: The check for existing and set have to be in the same critical area, otherwise there are things can happen between check and set.

sync/backup.go
```diff
@@ -150,16 +153,99 @@ func (t *Task) restoreBackup(replicaInController *types.ControllerReplicaInfo, b
...
+		if err != nil {
+			return nil, err
+		}
+
+		// call to controller to revert to sfile
+		if err := t.client.VolumeRevert(snapshotID); err != nil {
+			return nil, fmt.Errorf("Fail to revert to snapshot %v: %v", snapshotID, err)
+		}
+
+		//TODO: Call for sync restore object delete
```
yasker: We can leave the previous restoration information there. I don't think we need to hurry to delete it since if later it happens again, the old one will be overridden.

app/backup.go
```diff
@@ -240,15 +253,53 @@ func doRestoreBackup(c *cli.Context) error {
...
+	if rsList == nil {
+		//No active restore happening
+		return nil
+	}
+
+	if task.IsRestorationComplete(rsList) {
+		cli := getCli(c)
+		//Current Restored Flag empty for regular Restore backup
+		if err = cli.VolumeFinishRestore(""); err != nil {
+			return errors.Wrapf(err, "failed to finsish incrementally restoring")
```
yasker: `finish`.

replica/server.go
```diff
@@ -21,6 +22,8 @@ const (
 	Dirty      = State("dirty")
 	Rebuilding = State("rebuilding")
 	Error      = State("error")
+
+	Retry = 5
```
yasker: `ProgressBasedTimeoutInMinute`.

integration/core/test_cli.py
```diff
@@ -643,6 +643,28 @@ def get_backup_url(bin, backupID):
...
+        for status in rs:
+            if status['progress'] == 100:
+                completed += 1
+                continue
+            if 'restoreError' in status.keys():
+                assert status['restoreError'] == ""
+        if completed == len(rs):
+            return
+        time.sleep(1)
+    assert x != RETRY_COUNTS
```
yasker: `assert completed == len(rs)`

controller/control.go
```diff
@@ -584,7 +584,7 @@ func (c *Controller) PrepareRestore(lastRestored string) error {
 	if c.isRestoring {
 		return fmt.Errorf("volume %v is restoring", c.Name)
 	}
-	if c.lastRestored != "" && c.lastRestored != lastRestored {
+	if c.lastRestored != "" && c.lastRestored != lastRestored && lastRestored != "" {
```
yasker: OK, I cannot understand this commit without looking at the function using it. Merge it with the commit that using it.

app/restore_to_file.go
```diff
@@ -5,10 +5,13 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
+	"gopkg.in/cheggaaa/pb.v2"
```
yasker: You haven't checked in the vendor change for this dependency.

app/restore_to_file.go
```diff
@@ -21,6 +24,8 @@ const (
 	BackupFilePath        = "backup.img"
 	BackupFileConverted   = "backup.img.converted"
 	BackingFileCopy       = "backing.img.cp"
+
+	Retry = 5
```
yasker: Still the same name...

Also, it should be in minutes. e.g. `ProgressBasedTimeout = 5 * time.Minute`

5 seconds is too soon. TCP retry is longer than that.

app/restore_to_file.go
```diff
@@ -59,6 +64,78 @@ func RestoreToFileCmd() cli.Command {
...
+			restoreObj.RUnlock()
+			bar.Finish()
+			periodicChecker.Stop()
+			return restoreObj.RestoreError
+		}
+		now := time.Now()
+		diff := now.Sub(restoreObj.LastUpdatedAt)
+		restoreObj.RUnlock()
+
+		if diff.Seconds() > Retry {
```
yasker: Separate the periodicChecker with progress based timeout. We can refresh every 5 seconds for the progress bar but we won't want to error out just because it didn't finish restoring one block in 5 seconds.

app/restore_to_file.go
```diff
@@ -59,6 +64,78 @@ func RestoreToFileCmd() cli.Command {
...
+
+	bar := pb.StartNew(100)
+	periodicChecker := time.NewTicker(Retry * time.Second)
+	progress := 0
+
+	for _ = range periodicChecker.C {
+		restoreObj.RLock()
+		if restoreObj.RestoreProgress == 100 {
+			restoreObj.RUnlock()
+			for ; progress < 100; progress++ {
```
yasker: Seems instead of sleep 2ms (which isn't observable), you can use bar.Set().

app/restore_to_file.go
```diff
@@ -59,6 +64,78 @@ func RestoreToFileCmd() cli.Command {
...
+	if err != nil {
+		return err
+	}
+
+	bar := pb.StartNew(100)
+	periodicChecker := time.NewTicker(Retry * time.Second)
+	progress := 0
+
+	for _ = range periodicChecker.C {
+		restoreObj.RLock()
```
yasker: Get the values from restoreObj then release the lock. It's easier this way since you're only reading from the object, not writing into it.

app/restore_to_file.go
```diff
@@ -59,6 +64,78 @@ func RestoreToFileCmd() cli.Command {
...
+		now := time.Now()
+		diff := now.Sub(restoreObj.LastUpdatedAt)
+		restoreObj.RUnlock()
+
+		if diff.Seconds() > Retry {
+			bar.Finish()
+			periodicChecker.Stop()
+			return fmt.Errorf("no restore update happened since %v seconds. Returning failure", Retry)
+		}
+		for ; progress < restoreObj.RestoreProgress; progress++ {
```
yasker: Same here.

sync/rpc/server.go
```diff
@@ -282,27 +283,75 @@ func (*SyncAgentServer) BackupRemove(ctx context.Context, req *BackupRemoveReque
...
+		// validate environment variable first, since CronJob has set credential to environment variable.
+		if credential != nil && credential[types.AWSAccessKey] != "" && credential[types.AWSSecretKey] != "" {
+			os.Setenv(types.AWSAccessKey, credential[types.AWSAccessKey])
+			os.Setenv(types.AWSSecretKey, credential[types.AWSSecretKey])
+			os.Setenv(types.AWSEndPoint, credential[types.AWSEndPoint])
+		} else if os.Getenv(types.AWSAccessKey) == "" || os.Getenv(types.AWSSecretKey) == "" {
+			return nil, errors.New("could not backup to s3 without setting credential secret")
+		}
+	}
+	s.Lock()
```
yasker: You can use `defer s.Unlock() here`

sync/rpc/server.go
```diff
@@ -282,27 +283,75 @@ func (*SyncAgentServer) BackupRemove(ctx context.Context, req *BackupRemoveReque
...
+	logrus.Infof("Successfully initiated restore for snapshot [%v]", restoreObj.SnapshotName)
+	s.Unlock()
 
-	logrus.Infof("Done running %s %v", "sbackup", cmd.Args)
 	return &Empty{}, nil
 }
 
+func (s *SyncAgentServer) BackupRestoreStatus(ctx context.Context, req *Empty) (*BackupRestoreStatusReply, error) {
+	s.Lock()
+	restoreStatus := s.RestoreList
```
yasker: You need a copy of the object. Otherwise you're still operating the object (because it's a pointer) directly.

sync/backup.go
```diff
@@ -150,16 +153,132 @@ func (t *Task) restoreBackup(replicaInController *types.ControllerReplicaInfo, b
...
+
+	// Not required to revert the snapshot for incremental full backup as it's renamed and replica is reloaded
+	if !isIncrementalFullRestore && t.IsRestorationComplete(restoreStatusList) {
+		snapshotID, err := replica.GetSnapshotNameFromDiskName(snapshotFile)
+		if err != nil {
+			return nil, err
+		}
+
+		// call to controller to revert to sfile
+		if err := t.client.VolumeRevert(snapshotID); err != nil {
```
yasker: You cannot do functional change in a function suppose only to get the status. GetRestoreStatus is not required to be called at all. You need to have a thread (goroutine) waiting for the restoration completed then call the revert or rename. It needs to be run on the server side (since the client would be gone).

replica/server.go
```diff
@@ -354,25 +379,53 @@ func (s *Server) Restore(url, name string) (err error) {
...
+					ProgressBasedTimeoutInMinute)
+				periodicChecker.Stop()
+				return
+			}
+		}
+		periodicChecker.Stop()
+
+		if restoreObj.RestoreProgress == 100 {
+			logrus.Infof("Successfully completed backup restore")
+			postRestoreOperations(snapName, r)
```
yasker: hmm, it's done right here... Not sure why it's done differently in the previous commit.

app/backup.go
```diff
@@ -265,6 +265,10 @@ func doRestoreBackup(c *cli.Context) error {
 		return err
 	}
 	if err := task.RestoreBackup(backupURL); err != nil {
+		//Finish Restore as initiation failed
+		if err = cli.VolumeFinishRestore(""); err != nil {
```
yasker: Do we need to call cli.VolumeFinishRestore() in this case? Normally it should be handled automatically by the task.RestoreBackup().

sync/rpc/server.go
```diff
@@ -352,24 +355,43 @@ func (s *SyncAgentServer) BackupRestoreStatus(ctx context.Context, req *Empty) (
...
+			return nil, errors.New("could not backup to s3 without setting credential secret")
+		}
+	}
 
-	logrus.Infof("Running %s %v", cmd.Path, cmd.Args)
-	if err := cmd.Wait(); err != nil {
-		logrus.Infof("Error running %s %v: %v", "sbackup", cmd.Args, err)
-		return nil, err
+	s.Lock()
+	if s.isRestorationInProgress() {
```
yasker: Can use `defer s.Lock()` here.

Check how `defer` works in Golang.

app/restore_to_file.go
shreesha21: As restore_to_file is a blocking operation, I thought of keeping it smaller. Do you suggest to keep it 5 minutes here as well?
The only downside would be to wait for 5 minutes even for shorter restores.

app/backup.go
shreesha21: Yes. This function is reverting the flag (c.IsRestoring in controller server) which was set by VolumePrepareRestore in line #263. I believe, this is required.

sync/backup.go
shreesha21: Yes, Removing and Renaming activity can be done on the server side and I've making those changes and testing them. 

However, Volume Revert and Replica Reload definitions are implemented in Controller Server and Replica Server respectively. I don't know how I can call those functions from SyncAgentServer as we don't have controller and replica client. Even to create those clients, I'd need some information which requires to be sent to SyncAgentServer (which I don't think is the right thing to do)

So is it okay to keep it in sync backup?

yasker: No, you cannot do this. This status check shouldn't have any side effect. Also, just imagine if the user never calls to check the status of restoration, then the restoration will never be completed.

But the point you raised is valid. There is a way out. During the restoration, if no frontend is enabled, the revert won't be needed in the controller since it's mainly used to refresh the cache of the block device (frontend). Also, the currently logic makes replicas automatically revert to the restored snapshot, which will cause problems anyway if the controller's frontend is still up.

So we can add a few constraints to make it work:
1. If the controller's frontend is up, no restoration is allowed.
2. If a restoration is in progress for any replica, `start-frontend` of the controller is not allowed. For this purpose, replicas need to reports if it's restoring or not.

Since replicas revert to the newly restored snapshot automatically, after the restoration is completed, the frontend can be started again and the volume should contain the restored volume content.

sync/rpc/server.go
```diff
@@ -31,7 +36,7 @@ type SyncAgentServer struct {
 	processesByPort map[int]string
 
 	BackupList  *BackupList
-	RestoreList *replica.Restore
+	RestoreInfo *replica.Restore
```
yasker: Merged with previous commit.

replica/client/client.go
```diff
@@ -523,7 +523,8 @@ func (c *ReplicaClient) BackupRestoreStatus() (*syncagentrpc.BackupRestoreStatus
 	return reply, nil
 }
 
-func (c *ReplicaClient) RestoreBackupIncrementally(backup, deltaFile, lastRestored string) error {
+func (c *ReplicaClient) RestoreBackupIncrementally(backup, deltaFile, lastRestored, snapshotDiskName string,
```
yasker: This commit should merge with the previous one.

replica/server.go
```diff
@@ -354,25 +380,60 @@ func (s *Server) Restore(url, name string) (err error) {
...
+			if diff.Minutes() > ProgressBasedTimeoutInMinutes {
+				logrus.Errorf("no restore update happened since %v minutes. Returning failure",
+					ProgressBasedTimeoutInMinutes)
+				periodicChecker.Stop()
+				return
+			}
+		}
+		periodicChecker.Stop()
+
+		if restoreProgress == 100 {
```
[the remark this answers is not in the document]
shreesha21: https://github.com/longhorn/backupstore/blob/master//deltablock.go#L378
I'm setting the progress to 100 only after successful completion of restore. The round error you might be referring to wouldn't occur in this scneario because I have split the progress into 95%(for restore completion) and 5% for post restore operation. If the progress is 100, it implies restoration is complete

replica/server.go
```diff
@@ -354,25 +380,60 @@ func (s *Server) Restore(url, name string) (err error) {
...
-		UserCreated: false,
-		Created:     util.Now(),
+	config := &backupstore.DeltaRestoreConfig{
+		BackupURL: backupURL,
+		DeltaOps:  restoreObj,
+		Filename:  toFile,
 	}
 
-	if err := r.encodeToFile(newDisk, snapName+metadataSuffix); err != nil {
+	if err := backupstore.RestoreDeltaBlockBackup(config); err != nil {
```
yasker: The restoration cannot be done in the replica since it's not in the data path. It should be done in the SyncAgentServer.

The sync agent server would need to call the replica server to perform revert. 

This restore function is for replica command restoration only, so it should be removed when this is merged in.

PR #375, opened 2019-07-09
Defer VolumeFinishPurge call

ttpcodes:
This PR makes changes to the implementation of the lock on `Snapshot` purging that was introduced in #369. Instead of releasing the lock on the purge `Task` at the end of the operation (meaning that the lock will only be released if the `Task` was successful), the release is now deferred after the call to `t.client.VolumePreparePurge`, meaning that even if some error occurs during the purge process, the lock will still be released and the client will still be able to make another purge request.

### DIFF
--- a/sync/sync.go
+++ b/sync/sync.go
@@ -54,10 +54,20 @@ func (t *Task) DeleteSnapshot(snapshot string) error {
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


yasker:
Close this PR since we're going to moving the purging operation to the sync agent server in the future.

PR #379, opened 2019-07-12
Sync Agent Server Performing Incremental Restore

shreesha21:
The following code changes implements the following:
1. Sync Agent server implements the restore without the dependency of backup process (to perform incremental backup) and controller server(to maintain state changes)
2. Adds a new command backup restore-status to query current status of engine (restoring or not) and to fetch the latest backup in the replicas.
3. Implements sync-agent-server-cleanup command to reset the state values of sync agent server.


sync/rpc/server.go
```diff
@@ -282,7 +326,23 @@ func (*SyncAgentServer) BackupRemove(ctx context.Context, req *BackupRemoveReque
 	return &Empty{}, nil
 }
 
-func (*SyncAgentServer) BackupRestore(ctx context.Context, req *BackupRestoreRequest) (*Empty, error) {
+func (s *SyncAgentServer) BackupRestore(ctx context.Context, req *BackupRestoreRequest) (*Empty, error) {
+	if strings.HasSuffix(req.SnapshotFileName, ".snap_tmp") {
```
yasker: Is this for incremental restore fall back? I don't think we need to check this since normal full restore should able to update `restoring` and `lastRestored` as well.

sync/rpc/server.go
```diff
@@ -282,7 +326,23 @@ func (*SyncAgentServer) BackupRemove(ctx context.Context, req *BackupRemoveReque
...
+		if err := s.PrepareRestore(""); err != nil {
+			logrus.Errorf("failed to preprare backup restore for incremental full restore: %v", err)
```
yasker: Typo, `prepare`

sync/rpc/server.go
```diff
@@ -282,7 +326,23 @@ func (*SyncAgentServer) BackupRemove(ctx context.Context, req *BackupRemoveReque
...
+			return nil, err
+		}
+	}
+	finished := false
```
yasker: We can use `err` for `defer func()` if we add it to the function's signature as a return variable.

sync/rpc/server.go
```diff
@@ -299,18 +359,44 @@ func (*SyncAgentServer) BackupRestore(ctx context.Context, req *BackupRestoreReq
 		return nil, err
 	}
 
+	// Successfully finish incremental full restore
+	if strings.HasSuffix(req.SnapshotFileName, ".snap_tmp") {
+		if backupName, err := backupstore.GetBackupFromBackupURL(util.UnescapeURL(req.Backup)); err != nil {
+			return nil, err
+		} else if err := s.FinishRestore(backupName); err != nil {
```
yasker: Hmm, even we failed the restore we are still going to set the FinishRestore with backupName?

app/sync_agent.go
```diff
@@ -37,6 +38,17 @@ func SyncAgentCmd() cli.Command {
 	}
 }
 
+func SyncAgentServerCleanupCmd() cli.Command {
+	return cli.Command{
+		Name: "sync-agent-server-cleanup",
```
yasker: Sorry, `sync-agent-server-reset` is better.

backup/main.go
```diff
@@ -53,9 +41,7 @@ func Main() {
 
 	app := cli.NewApp()
 	app.Version = VERSION
-	app.Commands = []cli.Command{
-		backupRestoreCmd,
-	}
+	app.Commands = []cli.Command{}
```
yasker: We can remove the unnecessary codes for cmdline as well.

app/backup.go
```diff
@@ -27,6 +27,7 @@ func BackupCmd() cli.Command {
 			BackupStatusCmd(),
 			BackupRestoreCmd(),
 			RestoreToFileCmd(),
+			BackupRestoreStatusCmd(),
```
yasker: `RestoreStatusCmd()`

integration/data/test_restore_inc.py
```diff
@@ -137,8 +137,8 @@ def restore_inc_test(grpc_controller,  # NOQA
 
     assert not path.exists(STANDBY_REPLICA1_PATH + delta_file1)
     assert not path.exists(STANDBY_REPLICA2_PATH + delta_file1)
-    volume_info = cmd.info(CONTROLLER_NO_FRONTEND)
-    assert volume_info['lastRestored'] == backup1_name
+    volume_info = cmd.restore_status(CONTROLLER_NO_FRONTEND)
```
yasker: it's no longer `volume_info`.

Also, need to check all the replicas instead of just the first one.

sync/rpc/server.go
shreesha21: When the RestoreBackup function starts running in the context of Sync agent(instead of backup process), this issue is addressed then.

yasker: I don't understand..

shreesha21: In case if the restore failed, only the isRestoring flag should be turned back to false but the lastRestored flag should not be updated in this scenario. Is that right?

Or are you saying that this code will be hit, even if the restore is failed? 
Then what I was trying to say is that - In that scenario, when we move add the `DoRestoreBackup` function in the next commit, that will make sure this code is not hit in case of failure.

sync/rpc/server.go
shreesha21: Yes, this is for incremental restore fall back. The normal full restore wasn't updating `restoring` and `lastRestored` even earlier. These two fields are exclusively for incremental restore. If the incremental restore is successful, we are updating `lastRestored` and `isRestoring` below. Otherwise, we are only updating `isRestoring`, just like how it was done before.

backup/main.go
shreesha21: Do you mean to say that removing the entire main function from this file is better?

sync/rpc/server.go
yasker: OK.

Also, don't use `else` when you suppose to execute the next line of code when the previous code succeed.

sync/rpc/server.go
shreesha21: I quite didn't understand this. Could you please elaborate?

backup/main.go
yasker: Yeah.

sync/rpc/server.go
yasker: it looks like `finished == (err == nil)`, so we should be able to use something like this https://github.com/longhorn/longhorn-manager/blob/master/controller/volume_controller.go#L206

sync/rpc/server.go
```diff
@@ -299,18 +359,44 @@ func (*SyncAgentServer) BackupRestore(ctx context.Context, req *BackupRestoreReq
...
 	return &Empty{}, nil
 }
 
-func (*SyncAgentServer) BackupRestoreIncrementally(ctx context.Context, req *BackupRestoreIncrementallyRequest) (*Empty, error) {
+func (s *SyncAgentServer) BackupRestoreIncrementally(ctx context.Context, req *BackupRestoreIncrementallyRequest) (*Empty, error) {
+	if err := s.PrepareRestore(req.LastRestoredBackupName); err != nil {
+		logrus.Errorf("failed to prepare incremental restore: %v", err)
+		return nil, err
+	}
+	finished := false
```
yasker: finished is also used here.

sync/rpc/server.go
yasker: Unaddressed?

sync/rpc/server.go
```diff
@@ -389,21 +389,13 @@ func (s *SyncAgentServer) BackupRestoreIncrementally(ctx context.Context, req *B
...
-	}
-
-	logrus.Infof("Running %s %v", cmd.Path, cmd.Args)
-	if err := cmd.Wait(); err != nil {
-		logrus.Infof("Error running %s %v: %v", "sbackup", cmd.Args, err)
-		return nil, err
+	logrus.Infof("Running incremental restore %v to %s with lastRestoredBackup %s", req.Backup,
+		req.DeltaFileName, req.LastRestoredBackupName)
+	if err := backup.DoBackupRestoreIncrementally(req.Backup, req.DeltaFileName, req.LastRestoredBackupName); err != nil {
+		if extraErr := s.FinishRestore(""); extraErr != nil {
```
yasker: This should be covered by the `defer func()` above?

app/sync_agent.go
```diff
@@ -37,6 +38,17 @@ func SyncAgentCmd() cli.Command {
 	}
 }
 
+func SyncAgentServerCleanupCmd() cli.Command {
```
yasker: Change the cmd name as well.

sync/rpc/server.go
```diff
@@ -282,7 +325,21 @@ func (*SyncAgentServer) BackupRemove(ctx context.Context, req *BackupRemoveReque
...
+
+	defer func() {
+		if err != nil {
+			// Reset the isRestoring flag to false
+			if err := s.FinishRestore(""); err != nil {
+				logrus.Errorf("failed to finish backup restore: %v", err)
+				return
+			}
+		}
+	}()
```
yasker: Something seems wrong with this commit. Why it still contains `reexec.Command` below?

shreesha21: This is because the order of commits has varied from my local git history to the one in pull request.
Is it okay if I squash the commits `Moving the RestoreBackup function to sync-agent-server` and `Moving the Incremental Restore function to Sync Agent` into `Sync-Agent server maintaining restore state` ?

PR #380, opened 2019-07-16
 Add credential for restorations and adjust backup restore command for longhorn manager

shuo-wu:


Dockerfile.dapper
```diff
@@ -80,10 +80,10 @@ RUN cd /usr/src && \
 
 # Build longhorn-engine-launcher
 RUN cd /go/src/github.com/longhorn && \
-    git clone https://github.com/shuo-wu/longhorn-engine-launcher.git && \
+    git clone https://github.com/shuo-wu/longhorn-instance-manager.git longhorn-engine-launcher && \
```
yasker: Point to longhorn repo instead.

controller/revert.go
```diff
@@ -61,12 +61,16 @@ func (c *Controller) Revert(name string) error {
 	}
 
 	if err := c.startFrontend(); err != nil {
-		return err
+		if !strings.Contains(err.Error(), "no frontend started since it's not specified") {
```
yasker: If the frontend wasn't enabled, then we shouldn't call `shutdown` and `start` frontend.

PR #381, opened 2019-07-16
Add a command for backup volume deletion

ywei88:
Add a command for backup volume deletion
https://github.com/longhorn/longhorn/issues/598


app/backup.go
```diff
@@ -31,6 +31,7 @@ func BackupCmd() cli.Command {
 			cmd.BackupRemoveCmd(),
 			cmd.BackupListCmd(),
 			cmd.BackupInspectCmd(),
+			cmd.BackupVolumeRemoveCmd(),
```
yasker: Not needed if we add this to the BackupRemoveCmd()

PR #383, opened 2019-07-19
Frontend shutdown before snapshot revert

shreesha21:
The following changes include two things:
1. Integration test to manually shutdown before snapshot revert operation and start the launcher after snapshot revert completion.
2. Modify the controller logic in revert to check if the frontend is disabled before performing snapshot revert. Otherwise, abort the snapshot revert.

### DIFF
--- a/controller/revert.go
+++ b/controller/revert.go
@@ -36,10 +36,8 @@ func (c *Controller) Revert(name string) error {
 		return err
 	}
 
-	if c.Frontend() != "" {
-		if err := c.shutdownFrontend(); err != nil {
-			return err
-		}
+	if c.FrontendState() == "up" {
+		return fmt.Errorf("volume frontend enabled, aborting snapshot revert")
 	}
 
 	c.Lock()
@@ -47,9 +45,9 @@ func (c *Controller) Revert(name string) error {
 
 	minimalSuccess := false
 	now := util.Now()
-	for address, client := range clients {
+	for address, rClient := range clients {
 		logrus.Infof("Reverting to snapshot %s on %s at %s", name, address, now)
-		if err := client.Revert(name, now); err != nil {
+		if err := rClient.Revert(name, now); err != nil {
 			logrus.Errorf("Error on reverting to %s on %s: %v", name, address, err)
 			c.setReplicaModeNoLock(address, types.ERR)
 		} else {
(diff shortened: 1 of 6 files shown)


controller/revert.go
```diff
@@ -62,20 +60,6 @@ func (c *Controller) Revert(name string) error {
...
-
-		if c.launcher != "" {
-			logrus.Infof("Asking the launcher to start the frontend")
-			if err := c.launcherStartFrontend(); err != nil {
-				if !strings.Contains(err.Error(), "no frontend started since it's not specified") {
-					return err
-				}
-			}
-		}
-	}
```
shreesha21: This change is required, without which every single test case related to snapshot revert fails in integration test.
Hence, add the changes similar to the frontend restriction in backup restore.

integration/data/common.py
```diff
@@ -151,6 +152,12 @@ def restore_with_no_frontend(backup, grpc_c):
     assert v.frontendState == "up"
 
 
+def snapshot_revert_with_no_frontend(name):
```
yasker: This is in fact `snapshot_revert_with_frontend`.

integration/data/test_restore_to_file.py
```diff
@@ -140,7 +142,15 @@ def restore_to_file_with_backing_file_test(backing_dev, backup_target):
     os.remove(output_qcow2_path)
     assert not os.path.exists(output_qcow2_path)
 
+    launcher.shutdown_engine_frontend()
```
yasker: Can we put those wrappers around `cmd.snapshot_revert` into a common function as well?

integration/data/common.py
shreesha21: The function disables the engine frontend before running snapshot_revert right?
Are you implying me to change it to `snpashot_revert_with_frontend_disable` ?

yasker: I meant, this function disables the frontend, but it runs with frontend enabled. From the caller point of view, it's running the function when the frontend is enabled. The function knows what to do when the frontend is enabled (which is to disable it first). How to prepare the revert is not the concern of the caller.

shreesha21: Agreed, will change this to `snapshot_revert_with_frontend`

PR #385, opened 2019-07-22
Renames related to backup

shreesha21:
The following code modifications include renaming `backupError` to
`error`. Further, the replica backup object name
is changed from `Backup` to `BackupStatus` and it's attributes
`BackupProgress` and `BackupError` to `Progress` and `Error`
respectively.

### DIFF
--- a/integration/core/test_cli.py
+++ b/integration/core/test_cli.py
@@ -660,8 +660,8 @@ def get_backup_url(bin, backupID):
         if backup['progress'] == 100 and 'backupURL' in backup.keys():
             rValue = backup['backupURL']
             break
-        elif 'backupError' in backup.keys():
-            rValue = backup['backupError']
+        elif 'error' in backup.keys():
+            rValue = backup['error']
             break
         time.sleep(1)
     return rValue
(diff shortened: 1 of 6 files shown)


integration/core/test_cli.py
```diff
@@ -657,11 +657,11 @@ def get_backup_url(bin, backupID):
     for x in range(RETRY_COUNTS):
         output = subprocess.check_output(cmd).strip()
         backup = json.loads(output)
-        if backup['progress'] == 100 and 'backupURL' in backup.keys():
-            rValue = backup['backupURL']
+        if backup['progress'] == 100 and 'url' in backup.keys():
```
yasker: I think we can keep `backupURL` though since it's always used in this way. `backupstatus.url` is a bit misleading.

replica/backup.go
```diff
@@ -56,7 +56,7 @@ func (rr *RestoreStatus) UpdateRestoreStatus(snapshot string, rp int, re error)
 	rr.LastUpdatedAt = time.Now()
 }
 
-type Backup struct {
+type BackupStatus struct {
```
yasker: Can we update the `BackupError` and `BackupProgress` below as well?

PR #387, opened 2019-07-24
Backup and Restore Status

shreesha21:
The following code changes include:
1. Introducing BackupStatus State(complete, incomplete, error)
2. Introducing RestoreStatus State(complete, incomplete, error)
3. Integration test changes to check for the state
4. The datatype of error in restore status is changed from error to string


replica/backup.go
```diff
@@ -11,8 +11,14 @@ import (
 	"github.com/longhorn/backupstore"
 )
 
+type state string
```
yasker: `type ProgressState string`

replica/backup.go
```diff
@@ -11,8 +11,14 @@ import (
...
+
 const (
 	snapBlockSize = 2 << 20 // 2MiB
+
+	StateInComplete = state("incomplete")
```
yasker: `ProgressStateIncomplete = ProgressState("incomplete")` (incomplete is one word).

Same below.

replica/backup.go
```diff
@@ -54,6 +62,11 @@ func (rr *RestoreStatus) UpdateRestoreStatus(snapshot string, rp int, re error)
 	rr.Progress = rp
 	rr.Error = re
 	rr.LastUpdatedAt = time.Now()
+	if rr.Progress == 100 {
+		rr.State = StateComplete
+	} else if rr.Error != nil {
```
yasker: Just noticed the error should be a string too for RestoreStatus.

Also, it's error state then, not incomplete.

integration/core/test_cli.py
```diff
@@ -658,11 +658,15 @@ def get_backup_url(bin, backupID):
         output = subprocess.check_output(cmd).strip()
         backup = json.loads(output)
         if backup['progress'] == 100 and 'backupURL' in backup.keys():
+            assert backup['state'] == "complete"
```
yasker: it's better if use `backup['state']` to check e.g. progress, error etc. instead of using it in a reverse way.

Same below.

replica/backup.go
```diff
@@ -52,7 +60,13 @@ func (rr *RestoreStatus) UpdateRestoreStatus(snapshot string, rp int, re error)
 
 	rr.SnapshotName = snapshot
 	rr.Progress = rp
-	rr.Error = re
+	if re != nil {
+		rr.Error = re.Error()
+		rr.State = ProgressStateError
+	}
```
yasker: `else`

shreesha21: `else if rr.Progress == 100 { ` ?

PR #390, opened 2019-07-25
Update longhorn-instance-manager vendor

ttpcodes:
This PR updates `longhorn-instance-manager` to the latest version (currently `3e573d294759af29c18daae037ba8a2bd0cb85fb`).


ttpcodes:
- Updated `vendor.conf` dependency.

PR #392, opened 2019-07-26
Adding BackupURL into Restoration Status

shreesha21:
The following code changes add backupURL into Restoration status and modifies the integration test to validate the addition of this new field. Further, it also sets the Backup Restoration as complete from the right place.


yasker:
Need to update https://github.com/longhorn/longhorn-engine/blob/master/integration/data/common.py#L236 as well.

integration/data/test_restore_inc.py
```diff
@@ -132,6 +132,7 @@ def restore_inc_test(grpc_engine_manager,  # NOQA
         for i in range(RETRY_COUNTS):
             rs = cmd.restore_status(dr_address)
             for status in rs.values():
+                assert status['backupURL'] != ""
```
yasker: Should be?
```
if status['backupURL'] != backup1:
    break
```
So it can sleep for 1 second then retry in case the restored status wasn't updated yet?

Same for the others.

integration/core/test_cli.py
```diff
@@ -657,6 +657,7 @@ def restore_backup(engine_manager_client,  # NOQA
         rs = json.loads(subprocess.check_output(status_cmd).strip())
         for status in rs.values():
             assert 'state' in status.keys()
+            assert status['backupURL'] != ""
```
yasker: And move `time.sleep(3)` to the bottom of the loop, and update to etc. `time.sleep(RETRY_INTERVAL)`

sync/rpc/server.go
```diff
@@ -480,6 +480,7 @@ func (s *SyncAgentServer) completeBackupRestore() (err error) {
 			logrus.Errorf("failed to complete incremental fallback full restore: %v", err)
 			return err
 		}
+		s.RestoreInfo.FinishRestore()
```
yasker: Feels like we can merge the call with `s.FinishRestore()`

PR #395, opened 2019-08-01
Update for longhorn-instance-manager

shuo-wu:


yasker:
LGTM. Wait for the manager part.

PR #397, opened 2019-08-01
Add a test case for empty backup volume

ywei88:


integration/data/cmd.py
```diff
@@ -101,6 +101,12 @@ def backup_inspect(url, backup):
     return json.loads(subprocess.check_output(cmd))
 
 
+def backup_list(url, name, dest):
```
yasker: `backup_volume_list`

integration/data/test_backup.py
```diff
@@ -288,3 +289,51 @@ def test_backup_volume_deletion(grpc_replica1, grpc_replica2,  # NOQA
...
+
+        snap2_data = random_string(length)
+        verify_data(dev, offset, snap2_data)
+        snap2 = cmd.snapshot_create(address)
+
+        # test if we can recreate backup for this empty volume
+        backup_info2 = create_backup(address, snap2, backup_target)
+        assert backup_info2["VolumeName"] == VOLUME_NAME
+        assert backup_info2["Size"] == BLOCK_SIZE_STR
+        cmd.backup_rm(address, backup_info2["URL"])
```
yasker: Check if volume has one backup before the removal.

integration/data/test_backup.py
```diff
@@ -288,3 +289,51 @@ def test_backup_volume_deletion(grpc_replica1, grpc_replica2,  # NOQA
...
+
+        backup_info1 = create_backup(address, snap1, backup_target)
+        assert backup_info1["VolumeName"] == VOLUME_NAME
+        assert backup_info1["Size"] == BLOCK_SIZE_STR
+
+        cmd.backup_rm(address, backup_info1["URL"])
+        info1 = cmd.backup_list(address, VOLUME_NAME, backup_target)
+
+        assert info1[VOLUME_NAME]["Name"] == VOLUME_NAME
+        with pytest.raises(KeyError) as e:
```
yasker: What's this error? I thought it shouldn't error out.

integration/data/cmd.py
ywei88: this is the cmd that list backups for a backup volume, also we had a function "backup_volume_list"

yasker: in fact, `backup_volume_list_backup`

ywei88: modified

PR #402, opened 2019-08-06
Move SnapshotPurge to SyncAgentServer

ttpcodes:
This PR migrates and modifies the `SnapshotPurge` logic to the `SyncAgentServer` in order to allow for the `SnapshotPurge` to run as an asynchronous `Task`.

The only logic from the previous `SnapshotPurge` implementation that has been retained in the `Task` portion is the marking of system-generated `Snapshots` as `Removed` and the generation of the `Leaves` that represent the `Snapshots` that are to be removed. The `Leaves` are passed to separate calls to the `SyncAgentServer` for each individual `Replica` and will trigger the start of the actual `SnapshotPurge` process on each individual `Replica`.

The actual `SnapshotPurge` in the `SyncAgentServer` now runs as a `goroutine` and updates a `PurgeStatus` variable that represents the status of the `SnapshotPurge`. The user can use the `purge-status` subcommand of `longhorn snapshot` in order to check the status of the purge operation. The `SyncAgentServer` also tracks if there is currently a `SnapshotPurge` running, and will prevent multiple of these operations from attempting to run concurrently.

The integration tests used during the build for `longhorn-engine` have been modified accordingly in order to ensure that the `Snapshot` purge `Task` has completed before continuing.

This PR implements the `longhorn-engine` component of longhorn/longhorn#665 and should fix the race condition described in longhorn/longhorn#581.


ttpcodes:
- Added additional `Progress` updates when running individual `Snapshot` operations.

sync/rpc/rpc.proto
```diff
@@ -96,3 +97,7 @@ message RestoreStatusReply {
     string state = 6;
     string backupURL = 7;
 }
+
+message SnapshotPurgeRequest {
+    repeated string leaves = 1;
```
yasker: Can we let replicas figure out their own leaves?

sync/rpc/server.go
```diff
@@ -47,6 +47,7 @@ type SyncAgentServer struct {
 	replicaAddress  string
 
 	BackupList  *BackupList
+	PurgeStatus *PurgeStatus
```
yasker: After `RestoreInfo`

sync/rpc/server.go
```diff
@@ -778,8 +785,8 @@ func (s *SyncAgentServer) SnapshotPurge(ctx context.Context, req *SnapshotPurgeR
 		var err error
 
 		defer func() {
-			if err != nil {
```
yasker: Just do

```
if err != nil {
    s.PurgeStatus.Error = err
}
```

And you would need a lock for updating and reading PurgeStatus, since read/write can happen at the same time.

sync/rpc/server.go
```diff
@@ -966,6 +981,7 @@ func (s *SyncAgentServer) processRemoveSnapshot(snapshot string) error {
 				return err
 			}
 		}
+		s.PurgeStatus.Progress = int((float32(index)/float32(count) + (float32(opIndex)+1)/float32(opCount)) * 100)
```
yasker: Unnecessary since the most time-consuming one here is `sfold`.

It's better if we can tap into `sfold` and get some progress from https://github.com/longhorn/sparse-tools/blob/master/sparse/sfold.go#L52 . (need to check `sfold` into function call instead fo binary call)

sync/rpc/server.go
```diff
@@ -807,22 +816,25 @@ func (s *SyncAgentServer) SnapshotPurge(ctx context.Context, req *SnapshotPurgeR
...
 				snapshot = info.Parent
 			}
 			// Update snapshotInfo in case some nodes have been removed
-			snapshotsInfo, err = s.getSnapshotsInfo()
-			if err != nil {
+			snapshotsInfo, s.PurgeStatus.Error = s.getSnapshotsInfo()
+			if s.PurgeStatus.Error != nil {
 				return
 			}
+			s.PurgeStatus.Progress = int((float32(index) + 1) / float32(leafCount) * 100)
```
yasker: We only have one leaf normally... 

We can calculate the progress based on how many snapshots we need to remove.

sync/rpc/rpc.proto
```diff
@@ -101,3 +102,9 @@ message RestoreStatusReply {
 message SnapshotPurgeRequest {
     repeated string leaves = 1;
 }
+
+message SnapshotPurgeStatusReply {
+    bool isPurging = 1;
```
yasker: Also, `string state`

Can be:
```
in_progress
completed
error
```

sync/rpc/server.go
```diff
@@ -840,6 +840,21 @@ func (s *SyncAgentServer) SnapshotPurge(ctx context.Context, req *SnapshotPurgeR
 	return &empty.Empty{}, nil
 }
 
+func (s *SyncAgentServer) SnapshotPurgeStatus(ctx context.Context, req *empty.Empty) (*SnapshotPurgeStatusReply, error) {
+	// Check that the error isn't nil before calling Error(), or we could possibly trigger a panic in the
```
yasker: Just record the error message instead of the error then.

app/snapshot.go
```diff
@@ -179,6 +191,22 @@ func purgeSnapshot(c *cli.Context) error {
...
 }
 
+func purgeSnapshotStatus(c *cli.Context) error {
+	task := sync.NewTask(c.GlobalString("url"))
+	statusMap, err := task.PurgeSnapshotStatus()
+	if err != nil {
+		return err
+	}
+
+	purgeStatus, err := json.Marshal(statusMap)
```
yasker: `output, err = json.MarshalIndent(statusMap, "", "\t")`

app/snapshot.go
```diff
@@ -26,6 +26,7 @@ func SnapshotCmd() cli.Command {
 			SnapshotLsCmd(),
 			SnapshotRmCmd(),
 			SnapshotPurgeCmd(),
+			SnapshotPurgeStatusCommand(),
```
yasker: The client needs to check if any replica is purging before proceeding as well.

By default, the purge command should return an error if a purge is in progress. We will need a flag `--skip-if-in-progress` for the purge command return success if purging is happening. The manager would need this flag for all the commands.

ttpcodes:
- Implemented `--skip-if-in-progress` flag.
- Formatted `longhorn snapshot purge-status` output with indentation.
- `PurgeStatus` now stores `string` from `error` during `SnapshotPurge` instead of the entire `error`.
- `Progress` now calculates from additional sources such as number of `Snapshots` that are `Removed` and `Progress` of `sfold`.
- `sfold` is now called by its `method` instead of by `reexec`.
- The `Error` in `PurgeStatus` is now set as part of the `defer` in the `goroutine` of `SnapshotPurge`.
- `PurgeStatus` now has a `RWMutex` for reading and writing data.
- `PurgeStatus` has been reordered to be placed after `RestoreInfo`.
- The `Leaves` of the `SnapshotPurge` operation are now calculated by the `SyncAgentServer`.
- Rebased onto `master`.
- Updated the `vendor` for `sparse-tools` to `19d6588be8e2e7d37a3b9dec9cd415e2f95cb41d`.

ttpcodes:
- Removed `sfold` registration from `reexec` in `main.go`.

sync/rpc/server.go
```diff
@@ -809,12 +825,14 @@ func (s *SyncAgentServer) SnapshotPurge(ctx context.Context, req *empty.Empty) (
...
 
-		snapshotsInfo, err = s.getSnapshotsInfo()
+		snapshotsInfo, markedRemoved, err := s.getSnapshotsInfo()
 		if err != nil {
 			return
 		}
 
 		// We're tracing up from each leaf to the root
+		var removed int
+		progress := [2]float32{0, float32(markedRemoved)}
```
yasker: The progress calculation is a bit too complex. We don't need to calculate in operation level since there are at most 2 (or 3?) operations for each remove snapshot. So  two levels of progress is enough:
1. Upper level: removed / markedAsRemoved * 100
2. Lower level: for each part in level 1, divided by the real-time coalescing progress (if it's a coalescing).

sync/rpc/server.go
```diff
@@ -68,6 +69,43 @@ type PurgeStatus struct {
 	State    types.ProcessState
 }
 
+type CoalesceTask struct {
+	doneChan     chan error
```
yasker: Don't need a channel. Just create a function to hold a lock and update the number. Can we reuse `PurgeStatus` struct? For example, add a private coalescingProgress field to it for calculation.

ttpcodes:
- `progressChan` has been removed from `CoalesceTask`. `PurgeStatus` has been modified to implement `FoldFileOperations` and contains additional information about the `Snapshots` being `Removed` in order to properly calculate the `Progress`.
- `CoalesceTask` has been rewritten to contain the same basic `doneChan` implementation used in `sparse-tools` for use with the `FileCoalesce` and `postIncrementalRestoreOperations` methods.
- The `Progress` calculation based on number of processed `Operations` has been removed.
- The `calculateProgress` method has been removed.

ttpcodes:
- Updated `sparse-tools` dependency. `FoldFile` is now a blocking call, and the logic will simply wait for `FoldFile` to finish instead of attempting to use `channels` to communicate the status of the operation.
- `CoalesceTask` has been removed. `CreateFileFoldOps` can instead be used to create an implementation of `FoldFileOperations` from `PurgeStatus`.
- `FileCoalesce` has been removed from the `gRPC` definition of `SyncAgentServer`.

types/types.go
```diff
@@ -7,6 +7,10 @@ const (
 	RW  = Mode("RW")
 	ERR = Mode("ERR")
 
+	ProcessStateCompleted  = ProcessState("completed")
```
yasker: Just realized backup/restore is using `complete` instead of `completed`.

https://github.com/longhorn/longhorn-engine/blob/master/replica/backup.go#L19

We'd better keep them consistent. And we can clean and merge them later.

sync/rpc/server.go
```diff
@@ -66,6 +67,22 @@ type PurgeStatus struct {
 	Error    string
 	Progress int
 	State    types.ProcessState
+
+	removed  int
+	toRemove int
+}
+
+func CreateFileFoldOps() *PurgeStatus {
```
yasker: `NewPurgeStatus`

sync/rpc/server.go
```diff
@@ -66,6 +67,22 @@ type PurgeStatus struct {
 	Error    string
 	Progress int
 	State    types.ProcessState
+
+	removed  int
+	toRemove int
```
yasker: Need a better name.

ttpcodes:
- Renamed `removed` to `processed` and `toRemove` to `total`.

ttpcodes:
- Renamed `CreateFileFoldOps` to `NewPurgeStatus`.
- Renamed `ProcessStateCompleted` to `ProcessStateComplete`.

yasker:
LGTM.

PR #405, opened 2019-08-12
Sync agent server performing rebuild

shreesha21:
Earlier, controller was performing the rebuild by identifying the `fromReplica` and `toReplica` and managing the file transfer. The following code changes will ensure that the controller will send all the necessary information to the sync agent server running with `toReplica` and sync-agent server will performing the rebuild. Once this is complete, controller will perform the `"postRebuild operations"`. This is still a synchronous rebuild.

Issue: https://github.com/longhorn/longhorn/issues/46


sync/rpc/server.go
```diff
@@ -1214,3 +1214,66 @@ func (b *BackupList) BackupDelete(backupID string) error {
...
+		}
+
+		if err := s.syncFile(disk+".meta", "", fromServer); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (s *SyncAgentServer) syncFile(fromFileName, toFileName string, fromServer string) error {
```
yasker: use `fromServer` as the first argument to keep the consistency.

sync/rpc/server.go
```diff
@@ -1214,3 +1214,66 @@ func (b *BackupList) BackupDelete(backupID string) error {
...
+	if err != s.syncFilesFromHealtyReplica(req.FromServer, req.Disks) {
+		return nil, err
+	}
+
+	return &empty.Empty{}, nil
+}
+
+func (s *SyncAgentServer) syncFilesFromHealtyReplica(fromServer string, disks []string) (err error) {
+	for _, disk := range disks {
+		if strings.Contains(disk, VolumeHeadName) {
```
yasker: It's better if we can validate this earlier so we won't encounter it late in the loop.

sync/rpc/server.go
```diff
@@ -1214,3 +1214,66 @@ func (b *BackupList) BackupDelete(backupID string) error {
 	}
 	return fmt.Errorf("backup not found %v", backupID)
 }
+
+func (s *SyncAgentServer) FileSync(ctx context.Context, req *FileSyncRequest) (e *empty.Empty, err error) {
+	if err != s.syncFilesFromHealtyReplica(req.FromServer, req.Disks) {
```
yasker: `err !=`?

sync/sync.go
```diff
@@ -265,7 +266,12 @@ func (t *Task) AddReplica(replica string) error {
 		return err
 	}
 
-	if err = t.syncFiles(fromClient, toClient, output.Disks); err != nil {
+	fromServer, err := t.getSourceSyncAgentServerURL()
```
yasker: We should have this somewhere already? Otherwise client won't be able to connect to the sync agent server.

replica/client/client.go
```diff
@@ -392,6 +392,27 @@ func (c *ReplicaClient) LaunchReceiver(toFilePath string) (string, int32, error)
...
+	if err != nil {
+		return fmt.Errorf("cannot connect to SyncAgentService %v: %v", c.syncAgentServiceURL, err)
+	}
+	defer conn.Close()
+	syncAgentServiceClient := syncagentrpc.NewSyncAgentServiceClient(conn)
+
+	ctx, cancel := context.WithTimeout(context.Background(), GRPCServiceLongTimeout)
+	defer cancel()
+
+	if _, err := syncAgentServiceClient.FileSync(ctx, &syncagentrpc.FileSyncRequest{
```
yasker: This FileSync RPC call should be async on the serverside and we can monitor the progress later before proceeding. And we don't know how long it will take to complete the sync so even `GRPCServiceLongTimeout` doesn't make sense.

sync/rpc/server.go
```diff
@@ -1214,3 +1214,63 @@ func (b *BackupList) BackupDelete(backupID string) error {
...
+	if err := s.syncFilesFromHealthyReplica(req.FromServer, req.Disks); err != nil {
+		return nil, err
+	}
+
+	return &empty.Empty{}, nil
+}
+
+func (s *SyncAgentServer) syncFilesFromHealthyReplica(fromServer string, disks []string) (err error) {
+	for _, disk := range disks {
+		if err := s.syncFile(fromServer, disk, ""); err != nil {
```
yasker: If we're not using the third parameter, there is no need to keep it.

sync/rpc/server.go
```diff
@@ -1214,3 +1214,63 @@ func (b *BackupList) BackupDelete(backupID string) error {
...
+	myPort := reply.Port
+
+	conn, err := grpc.Dial(fromServer, grpc.WithInsecure())
+	if err != nil {
+		return fmt.Errorf("cannot connect to SyncAgentService %v: %v", fromServer, err)
+	}
+	defer conn.Close()
+	syncAgentFromClient := NewSyncAgentServiceClient(conn)
+
+	ctx, cancel := context.WithTimeout(context.Background(), GRPCServiceCommonTimeout)
```
yasker: We don't know how long it will take to copy a file, it can be TB size. Timeout makes no sense.

replica/client/client.go
shreesha21: ok. Will use `context.Cancel(context.Background())` instead.

sync/rpc/server.go
shreesha21: ok. Will use context.Cancel(context.Background()) instead.

app/add_replica.go
```diff
@@ -31,3 +33,30 @@ func addReplica(c *cli.Context) error {
 	task := sync.NewTask(url)
 	return task.AddReplica(replica)
 }
+
+func RebuildStatusCmd() cli.Command {
```
shuo-wu: Put this command in a separate file

sync/rpc/server.go
```diff
@@ -1265,10 +1320,33 @@ func (s *SyncAgentServer) syncFile(fromServer string, fromFileName string) error
 		Host:         myHost,
 		Port:         int32(myPort),
 	}); err != nil {
-		logrus.Infof("Failed synchronizing %s to %s@%s:%d: %v", fromFileName, toFileName, myHost, myPort, err)
+		logrus.Errorf("Failed synchronizing %s to %s@%s:%d: %v", fromFileName, toFileName, myHost, myPort, err)
```
shuo-wu: Put this modification in the previous commit

integration/core/test_cli.py
```diff
@@ -122,6 +122,42 @@ def test_replica_add_start(bin, grpc_controller_client,  # NOQA
     assert volume.replicaCount == 1
 
 
+def check_rebuild_status(bin, url):
```
shuo-wu: Rename this function since it is actually waiting for rebuild complete.

integration/core/test_cli.py
```diff
@@ -122,6 +122,42 @@ def test_replica_add_start(bin, grpc_controller_client,  # NOQA
...
+    cmd = [bin, '--url', url, 'replica-rebuild-status']
+    completed = 0
+    rebuild_status = json.loads(subprocess.check_output(cmd).strip())
```
shuo-wu: redundant line.

integration/core/test_cli.py
```diff
@@ -122,6 +122,42 @@ def test_replica_add_start(bin, grpc_controller_client,  # NOQA
...
+            assert 'state' in rebuild.keys()
+            assert 'isRebuilding' in rebuild.keys()
+            assert 'error' in rebuild.keys()
+            assert 'progress' in rebuild.keys()
+
+            if rebuild['state'] == "complete":
+                assert rebuild['progress'] == 100
+                assert not rebuild['isRebuilding']
+                completed += 1
+                break
```
shuo-wu: Do we need this `break`? You are not using `switch`. Same below

sync/rpc/server.go
```diff
@@ -1244,6 +1298,7 @@ func (s *SyncAgentServer) syncFile(fromServer string, fromFileName string) error
 		ToFileName: toFileName,
 	})
 	if err != nil {
+		logrus.Errorf("failed to launch receiver at %v", myHost)
```
shuo-wu: Put this modification in the previous commit

sync/rpc/server.go
```diff
@@ -1232,10 +1263,33 @@ func (s *SyncAgentServer) syncFilesFromHealthyReplica(fromServer string, disks [
...
+	s.finishRebuild()
+
 	return nil
 }
 
+func (s *SyncAgentServer) prepareRebuild() {
+	s.RebuildInfo.Lock()
+	defer s.RebuildInfo.Unlock()
+
+	s.RebuildInfo.IsRebuilding = true
```
shuo-wu: Do we need to save this flag in struct `SyncAgentServer`? Just like `PrepareRestore()`, `PreparePurge()`

sync/rpc/server.go
```diff
@@ -1224,6 +1241,20 @@ func (s *SyncAgentServer) FileSync(ctx context.Context, req *FileSyncRequest) (*
...
+		if err != nil {
+			s.RebuildInfo.Lock()
+			defer s.RebuildInfo.Unlock()
+			s.RebuildInfo.IsRebuilding = false
+			s.RebuildInfo.State = types.ProcessStateError
+			s.RebuildInfo.Error = err.Error()
+		}
+	}()
+	total := len(disks)
+	s.RebuildInfo = NewRebuildStatus(total)
```
shuo-wu: Can we init `s.RebuildInfo` in func `NewSyncAgentServer()` and calculate `progress` in the below loop? Like `SnapshotPurge()`

The implementation here is a little bit different from the implementation of `SnapshotPurge()`. You can check which one is better/simpler and decide if you need to modify the current workflow.

app/add_replica.go
shreesha21: Sheng asked me to implement this way.

sync/rpc/server.go
shreesha21: I think in the rebuild scenario, It might not make sense to create a `rebuildStatus` object when the sync agent starts as this scneario may or may not happen. It would be better to create it on-demand in my opinion.

integration/core/test_cli.py
shreesha21: Sure, will use ->`wait_and_fetch_rebuild_status`

replica/client/client.go
yasker: I meant once the call client made to the server should just kick start the server work and return. I don't think `context.Background()` can do that. Can you check if this is indeed a blocking call?

yasker:
Btw, for testing this, we can create a volume with e.g. 10GB or 20GB, full of data, then try a rebuild. If any timeout is wrong, the rebuild won't succeed.

replica/client/client.go
shreesha21: Context.background documentation says: `context.Background() returns a non-nil, empty Context. It is never canceled, has no values, and has no deadline.` 
I tested with 10G and 20G content. Both of them succeeded.

```
root@srao1:/mnt/shr# du -sh 
21G	.

curl http://172.104.18.252:31546/v1/volumes |  jq '.data[0].rebuildStatus'
[
  {
    "actions": null,
    "error": "",
    "isRebuilding": false,
    "links": null,
    "progress": 100,
    "replica": "t1-r-e14e83f3",
    "state": "complete"
  },
  {
    "actions": null,
    "error": "",
    "isRebuilding": false,
    "links": null,
    "progress": 0,
    "replica": "t1-r-d8b7cefc",
    "state": ""
  },
  {
    "actions": null,
    "error": "",
    "isRebuilding": false,
    "links": null,
    "progress": 0,
    "replica": "t1-r-f2ade456",
    "state": ""
  }
]
```

yasker:
LGTM. Please rebase.

yasker:
Close this in favor of https://github.com/longhorn/longhorn-engine/pull/441

PR #409, opened 2019-08-15
Fix issue: tgt fails to access backend data in integration test

shuo-wu:
https://github.com/longhorn/longhorn/issues/646


yasker:
How did those file-based test pass before?

shuo-wu:
Those file-based backend controllers are used in the core test only. And the core test doesn't involve data r/w.

PR #335, opened 2019-03-07
engine: Add command to export the volume

shuo-wu:

PR #412, opened 2019-09-03
Adding Backup type in engine

shreesha21:
The following changes include modifying the backup creation to return the backup type along with the backup ID in longhorn-engine. Further, an integration test is added to test these changes in engine.

Issue: longhorn/longhorn#592


replica/backup.go
```diff
@@ -71,15 +71,16 @@ func (rr *RestoreStatus) FinishRestore() {
...
+	lock                sync.Mutex
+	backingFile         *BackingFile
+	replica             *Replica
+	volumeID            string
+	SnapshotID          string
+	Error               string
+	Progress            int
+	BackupURL           string
+	State               ProgressState
+	IsIncrementalBackup bool
```
yasker: `IsIncremental`

replica/backup.go
```diff
@@ -204,10 +205,11 @@ func (rb *BackupStatus) CloseSnapshot(snapID, volumeID string) error {
 func (rb *BackupStatus) CompareSnapshot(snapID, compareSnapID, volumeID string) (*backupstore.Mappings, error) {
 	id := GenerateSnapshotDiskName(snapID)
 	compareID := ""
+	rb.lock.Lock()
 	if compareSnapID != "" {
+		rb.IsIncrementalBackup = true
```
yasker: I thought we can get `IsIncremental` from the backup object. Not sure why we need to set it manually here?

sync/rpc/rpc.proto
```diff
@@ -51,6 +51,7 @@ message BackupCreateRequest {
 
 message BackupCreateReply {
     string backup = 1;
+    bool isIncrementalBackup = 2;
```
yasker: `isIncremental`

integration/data/test_backup.py
```diff
@@ -288,3 +288,77 @@ def test_backup_volume_deletion(grpc_replica1, grpc_replica2,  # NOQA
...
+        # backup3: 64 zero data + 192 random data in 1st block
+        length3 = 64
+        offset3 = 0
+        verify_data(dev, offset3, zero_string * length3)
+        verify_data(dev, length2, zero_string * offset2)
+        verify_data(dev, block_size, zero_string * length2)
+        snap3 = cmd.snapshot_create(address)
+        backup3 = create_backup(address, snap3, backup_target)
+        backup3_url = backup3["URL"]
+        assert backup3['IsIncremental'] is False
```
yasker: hmm, why this is not incremental?

replica/backup.go
shreesha21: There are basically two ways that I know of, for the engine to fetch the `IsIncremental` field and send as a response for backup creation.
1. Create a new API in backupstore library and fetch from the backup object. In this case, I will have to explicitly make two calls - one for getting backupID(CreateDeltaBlockBackup) and another for getting the backup_type(New Api) by using the backupID from the first call.

2. We can learn the backupType as soon as the backupstore invokes `CompareSnapshot` and record it in the `backupStatus` object. 

I chose option 2 as it's simpler and self contained in the replica/backup.go file.

integration/data/test_backup.py
shreesha21: because of line #335 as we are intentionally deleting the last backup.

replica/backup.go
yasker: We shouldn't need to set `IsIncremental` in two places.

I think it's better for the variable to set during the backup creation process, instead of in the callback, since from the callback we don't know how the caller will deal with the result - for example, some other steps in the later function failed so the caller decides to do a full backup. 

CreateDeltaBlockBackup can return the flag to indicate if it's incremental or not, just like what you did with CreateBackup call.

PR #429, opened 2019-10-18
Merge Longhorn engine code with Longhorn instance manager

yasker:
Those two are too close to be separated, besides we don't have separate integration test for the instance manager.

In the end, there will still be two binaries generated by the build, but those two would be managed in the same repo.


yasker:
Since we're basically moving all the history with the code from Longhorn instance manager repo, we will ignore the DCO warning this one time.

PR #436, opened 2019-11-15
instance manager: Make sure the engine launcher can be cleaned up

shuo-wu:
even if the related process is not found or the frontend cleanup fails

https://github.com/longhorn/longhorn/issues/854
https://github.com/longhorn/longhorn/issues/923


yasker:
This PR seems messed up.

PR #440, opened 2019-12-19
Upgrade Longhorn engine integration tests to use python 3

meldafrawi:


integration/data/snapshot_tree.py
```diff
@@ -63,55 +63,77 @@ def snapshot_tree_verify_relationship(address, snap, strict):
     info = cmd.snapshot_info(address)
 
     assert snap["0a"] in info
-    assert snap["0b"] in info[snap["0a"]]["children"]
+    assert snap["0b"] in \
```
yasker: We no longer need to change those, right?

integration/data/test_basic_ops.py
```diff
@@ -4,16 +4,16 @@
...
 
-import cmd
-from common import (  # NOQA
-    dev, get_dev, read_dev, write_dev,  # NOQA
-    grpc_engine_manager,  # NOQA
-    grpc_controller, grpc_replica1, grpc_replica2,  # NOQA
+import data.cmd as cmd
+from data.common import (  # NOQA
+    get_dev, read_dev, write_dev,  # NOQA
+    #grpc_engine_manager,  # NOQA
```
yasker: unused? Then we can remove them.

meldafrawi: fixed

integration/data/snapshot_tree.py
meldafrawi: fixed

PR #441, opened 2019-12-27
Replica rebuild progress report

shuo-wu:
https://github.com/longhorn/longhorn/issues/46


integration/core/common.py
```diff
@@ -243,3 +244,31 @@ def get_replica_head_file_path(replica_dir):
     cmd = ["find", replica_dir, "-name",
            '*volume-head-*.img']
     return subprocess.check_output(cmd).strip()
+
+
+def wait_for_rebuild_complete(bin, url):
```
shuo-wu: mention that `add-replica` is a sync call

PR #443, opened 2020-01-07
Update gPRC for duplicate proto definition

shuo-wu:
https://github.com/longhorn/longhorn/issues/999

### DIFF
--- a/integration/core/common.py
+++ b/integration/core/common.py
@@ -1,4 +1,5 @@
 import fcntl
+import json
 import struct
 import os
 import grpc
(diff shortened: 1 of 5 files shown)


generate_grpc.sh
```diff
@@ -11,15 +11,15 @@ if [ ! -e ./vendor_proto/protobuf/src/google/protobuf ]; then
 fi
 
 # controller
-protoc -I pkg/engine/controller/rpc/pb -I vendor_proto/protobuf/src/ pkg/engine/controller/rpc/pb/controller.proto --go_out=plugins=grpc:pkg/engine/controller/rpc/pb
+protoc -I pkg/engine/controller/rpc/pb -I vendor_proto/protobuf/src/ pkg/engine/controller/rpc/pb/controller.proto --go_out=plugins=grpc:$GOPATH/src/
```
yasker: Can we avoid using GOPATH? There is no guarantee GOPATH exists. With `go mod`, GOPATH can be empty.

shuo-wu: OK. I modified it to use the relative path. And I added comments in the file `syncagent.proto` to explain the reason.

generate_grpc.sh
```diff
@@ -11,15 +11,15 @@ if [ ! -e ./vendor_proto/protobuf/src/google/protobuf ]; then
 fi
 
 # controller
-protoc -I pkg/engine/controller/rpc/pb -I vendor_proto/protobuf/src/ pkg/engine/controller/rpc/pb/controller.proto --go_out=plugins=grpc:pkg/engine/controller/rpc/pb
+protoc -I pkg/engine/controller/rpc/pb -I vendor_proto/protobuf/src/ pkg/engine/controller/rpc/pb/controller.proto --go_out=plugins=grpc:./../../../
```
yasker: This makes less sense than $GOPATH.

If $GOPATH doesn't exist, how can we know what's at `./../../../`?

shuo-wu: As long as the developers follow the convention of go code organization, `./../../../` should be the parent directory of `github.com/`.

yasker: The convention you mentioned is GOPATH. User doesn't need to put directory there using go mod.

PR #458, opened 2020-02-03
Update gPRC for duplicate proto definition

shuo-wu:
Longhorn #999


yasker:
https://github.com/longhorn/longhorn/issues/999

PR #459, opened 2020-02-10
Introduce InstanceManagerAPIVersion and InstanceManagerAPIMinVersion

shuo-wu:
longhorn/longhorn#880


pkg/engine/meta/version.go
```diff
@@ -2,18 +2,24 @@ package meta
 
 const (
 	// CLIAPIVersion used to communicate with user e.g. longhorn-manager
+	// These 2 version numbers involves in the common volume operation commands,
+	// e.g. snapshot and backup.
```
yasker: Need to clarify when to bump the Min version vs version.

pkg/engine/meta/version.go
```diff
@@ -1,20 +1,34 @@
...
 	CLIAPIVersion    = 3
 	CLIAPIMinVersion = 3
 
-	// ControllerAPIVersion used to communicate with instance-manager
+	// ControllerAPIVersion used to indicate the current live upgrade mechanism.
+	// As long as the live upgrade implementation in longhorn-manager, the process
```
shuo-wu: remove `longhorn-manager` part

shuo-wu: remove `longhorn-manager` part

PR #463, opened 2020-03-06
Fix off-line expansion

shuo-wu:
https://github.com/longhorn/longhorn/issues/1040


pkg/controller/control.go
```diff
@@ -187,16 +187,7 @@ func (c *Controller) Snapshot(name string, labels map[string]string) (string, er
 }
 
 func (c *Controller) Expand(size int64) error {
-	if err := c.startExpansion(); err != nil {
```
yasker: This change is wrong.

Checking on `isExpanding` won't work if you don't hold the lock to update the `expanding = true` immediately.

Consider multiple request coming at the same time, all of them will check `isExpanding` at the same time, and found  `isExpanding == false`, then all of them can go ahead and perform the expansion operation at the same time.

The previous way of implementing `startExpansion` and `finishExpansion` is correct because:
1. The lock was held for a very brief period of time, so no RW should be affected.
2. The value will be updated within the same critical section, so if another thread wanted to expand doesn't have the lock, it will see updated value after getting the lock, then gave up the operation.
3. `canAdd` is protected by the same lock of course.

pkg/controller/control.go
```diff
@@ -208,14 +204,14 @@ func (c *Controller) Expand(size int64) error {
...
 		}
 
-		if size < c.size {
-			logrus.Errorf("Cannot expand to a smaller size %v", size)
-			return
-		} else if size == c.size {
-			logrus.Infof("Controller %v is already expanded to size %v", c.Name, size)
-			return
-		}
+		expanded := false
```
yasker: Move this to the beginning of the function. Otherwise, if the code before this function exits, the lock will never be released (this won't happen in this case but it's better to follow the pattern).

yasker:
LGTM.

PR #466, opened 2020-03-23
Rebuild state is wrongly updated to complete after file sending failure

shuo-wu:
longhorn/longhorn#1103

### DIFF
--- a/pkg/sync/rpc/server.go
+++ b/pkg/sync/rpc/server.go
@@ -891,100 +891,100 @@ func (s *SyncAgentServer) SnapshotPurge(ctx context.Context, req *empty.Empty) (
 		return nil, err
 	}
 
-	go func() {
-		var err error
-
-		defer func() {
-			s.PurgeStatus.Lock()
-			if err != nil {
-				s.PurgeStatus.Error = err.Error()
-				s.PurgeStatus.State = types.ProcessStateError
-			} else {
-				s.PurgeStatus.State = types.ProcessStateComplete
-			}
-			s.PurgeStatus.Unlock()
-
-			if err := s.FinishPurge(); err != nil {
-				logrus.Errorf("could not mark finish purge: %v", err)
-			}
-		}()
-
-		var leaves []string
-
-		snapshotsInfo, _, err := s.getSnapshotsInfo()
-		if err != nil {
-			return
-		}
-
-		for snapshot, info := range snapshotsInfo {
-			if len(info.Children) == 0 {
-				leaves = append(leaves, snapshot)
-			}
-			if info.Name == VolumeHeadName {
-				continue
-			}
-			// Mark system generated snapshots as removed
-			if !info.UserCreated && !info.Removed {
-				if err = s.markSnapshotAsRemoved(snapshot); err != nil {
-					return
-				}
-			}
-		}
-
-		snapshotsInfo, markedRemoved, err := s.getSnapshotsInfo()
-		if err != nil {
-			return
-		}
-
-		s.PurgeStatus.Lock()
-		s.PurgeStatus.total = markedRemoved
-		s.PurgeStatus.Unlock()
-
-		// We're tracing up from each leaf to the root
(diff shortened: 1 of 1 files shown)


pkg/sync/rpc/server.go
```diff
@@ -344,8 +345,8 @@ func (s *SyncAgentServer) FilesSync(ctx context.Context, req *ptypes.FilesSyncRe
 			ops = s.RebuildStatus
 		}
 
-		port, err := s.launchReceiver("FilesSync", info.ToFileName, ops)
-		if err != nil {
+		var port int
```
yasker: It's very easy to make mistakes like this if it's done in this way.

Use the pattern in https://github.com/longhorn/longhorn-manager/blob/d2107aa19465040add77bd2198dd6b9a7cf4a109/api/volume.go#L20 instead.

Define the return valuable name in the function signature, so it can be used in the `defer func()` automatically.

Check other functions and mark the change if necessary.

shuo-wu: For `SnapshotPurge()`, it's better to define the variable `err` inside the goroutine. We don't need to modify this function.

yasker: Yeah, that doesn't apply because it's a goroutine.

pkg/sync/rpc/server.go
```diff
@@ -307,12 +307,11 @@ func (s *SyncAgentServer) launchReceiver(processName, toFileName string, ops spa
 	return port, nil
 }
 
-func (s *SyncAgentServer) FilesSync(ctx context.Context, req *ptypes.FilesSyncRequest) (*empty.Empty, error) {
-	if err := s.PrepareRebuild(req.SyncFileInfoList); err != nil {
+func (s *SyncAgentServer) FilesSync(ctx context.Context, req *ptypes.FilesSyncRequest) (res *empty.Empty, err error) {
+	if err = s.PrepareRebuild(req.SyncFileInfoList); err != nil {
```
yasker: Don't need to remove `:`. The return result will be used as `err`

pkg/sync/rpc/server.go
yasker: In fact for `SnapshotPurge`, it's better to put the logic in the goroutine into a separate function. It's too long and can easily result in mistakes.

pkg/sync/rpc/server.go
```diff
@@ -321,10 +320,11 @@ func (s *SyncAgentServer) FilesSync(ctx context.Context, req *ptypes.FilesSyncRe
 			logrus.Errorf("Sync agent gRPC server failed to rebuild replica/sync files: %v", err)
 		} else {
 			s.RebuildStatus.State = types.ProcessStateComplete
+			logrus.Infof("Sync agent gRPC server finished rebuilding replica/sync files for replica %v", req.ToHost)
 		}
 		s.RebuildStatus.Unlock()
 
-		if err := s.FinishRebuild(); err != nil {
+		if err = s.FinishRebuild(); err != nil {
```
yasker: Don't need to remove `:`. Just `return err` at the end.

pkg/sync/rpc/server.go
```diff
@@ -344,8 +344,8 @@ func (s *SyncAgentServer) FilesSync(ctx context.Context, req *ptypes.FilesSyncRe
 			ops = s.RebuildStatus
 		}
 
-		port, err := s.launchReceiver("FilesSync", info.ToFileName, ops)
-		if err != nil {
+		var port int
+		if port, err = s.launchReceiver("FilesSync", info.ToFileName, ops); err != nil {
```
yasker: Don't need to change this part now.

`err` will be the one updated by the `return` statement.

Should add `:` to the `err := fromClient.SendFile()` below.

pkg/sync/rpc/server.go
shuo-wu: It's impossible to `return err` for this defer function.

PR #471, opened 2020-04-09
Expansion rollback

shuo-wu:
https://github.com/longhorn/longhorn/issues/978


pkg/replica/replica.go
```diff
@@ -117,6 +117,19 @@ const (
 	OpReplace  = "replace"
 )
 
+func (in Info) DeepCopy() Info {
```
yasker, line 121: Unnecessary

pkg/controller/control.go
```diff
@@ -402,6 +402,9 @@ func (c *Controller) StartFrontend(frontend string) error {
 	c.Lock()
 	defer c.Unlock()
 
+	if c.isExpanding {
+		return fmt.Errorf("Cannot start frontend during the enigne expanison")
```
yasker, line 411: Typo

pkg/controller/replicator.go
```diff
@@ -207,18 +207,21 @@ func (r *replicator) Snapshot(name string, userCreated bool, created string, lab
...
 }
 
-func (r *replicator) Expand(size int64) error {
+func (r *replicator) Expand(previousSize, newSize int64) (bool, error) {
 	retErrorLock := sync.Mutex{}
 	retError := &BackendError{
 		Errors: map[string]error{},
 	}
+	rwReplicaCount := 0
+	rolledBack := false
```
shuo-wu, line 216: remove it

pkg/controller/replicator.go
```diff
@@ -207,18 +207,21 @@ func (r *replicator) Snapshot(name string, userCreated bool, created string, lab
 	return nil
 }
 
-func (r *replicator) Expand(size int64) error {
+func (r *replicator) Expand(previousSize, newSize int64) (bool, error) {
```
shuo-wu, line 210: Add comments for this funciton

proto/ptypes/replica.proto
```diff
@@ -42,6 +43,11 @@ message ReplicaExpandRequest {
     int64 size = 1;
 }
 
+message ReplicaExpansionRollbackRequest {
+    int64 previous_size = 1;
```
shuo-wu, line 47: Need a better name

integration/data/test_ha.py
```diff
@@ -512,3 +517,133 @@ def test_expansion_with_rebuild(grpc_controller,  # NOQA
...
+    f = os.open(fake_head_1, os.O_CREAT | os.O_RDWR)
+    os.close(f)
+    f = os.open(fake_head_2, os.O_CREAT | os.O_RDWR)
+    os.close(f)
+
+    # All replicas' expansion will fail
+    # then engine will do rollback automatically
+    grpc_controller.volume_frontend_shutdown()
+    grpc_controller.volume_expand(EXPANDED_SIZE)
+    time.sleep(5)
```
shuo-wu, line 556: Check engine.isExpanding

integration/data/test_ha.py
```diff
@@ -512,3 +517,133 @@ def test_expansion_with_rebuild(grpc_controller,  # NOQA
...
+    grpc_controller.volume_frontend_start(FRONTEND_TGT_BLOCKDEV)
+    wait_and_check_volume_expansion(
```
shuo-wu, line 558: wait_and_check_volume_size()

proto/ptypes/controller.proto
```diff
@@ -42,6 +42,9 @@ message Volume {
     string frontend = 5;
     string frontendState = 6;
     bool isExpanding = 7;
```
yasker, line 44: Just notice the naming convention for Protobuf field is wrong. It should be `is_expanding` instead of `isExpanding`.

Not sure if we can fix it now though.

shuo-wu: I knew. And actually you had told me last year.
I am not sure if we can fix it now, either.

yasker: It's internal GRPC between engine and replica right? It sounds like we can fix it without breaking anything. Need a separate PR though.

pkg/replica/replica.go
```diff
@@ -731,6 +732,13 @@ func (r *Replica) rmDisk(name string) error {
 	if err := os.Remove(r.diskPath(name + metadataSuffix)); err != nil {
 		lastErr = err
 	}
+	// the tmp meta file may not exist
```
yasker, line 735: When does tmp meta file exist?

proto/ptypes/replica.proto
```diff
@@ -42,6 +42,11 @@ message ReplicaExpandRequest {
     int64 size = 1;
 }
 
+message ReplicaExpandReply {
+    Replica replica = 1;
+    bool rollback_failed = 2;
```
yasker, line 47: if `rollback_failed == false`, how did we know if the volume expansion is successful or not?

pkg/backend/remote/remote.go
```diff
@@ -107,25 +107,29 @@ func (r *Remote) Snapshot(name string, userCreated bool, created string, labels
 	return nil
 }
 
-func (r *Remote) Expand(size int64) error {
+func (r *Remote) Expand(size int64) (bool, error) {
```
yasker, line 110: Add comments to explain the return variables.

proto/ptypes/controller.proto
shuo-wu: No. This is the gRPC between manager and engine. But I can fix the naming in replica.proto and syncagent.proto

proto/ptypes/replica.proto
yasker: or
```
bool expansion_result = 2;
string expansion_error_msg = 3;
```
So if expansion failed but roll back succeed, the function return error nil and set expansion_result = false and error message.
If the expansion failed and rollback failed, the function return error fo the rollback, and set the expansion result = false and error message for expansion as well.

pkg/types/error.go
```diff
@@ -0,0 +1,100 @@
+package types
+
+import (
+	"encoding/json"
+	"fmt"
+	"regexp"
+)
+
+type ErrorCode int
```
yasker, line 9: Use string instead of int. We can be more verbose since we're not dealing with C.

pkg/types/error.go
```diff
@@ -0,0 +1,100 @@
...
+
+const (
+	ErrorCodeResultUnknown                   = ErrorCode(0)
+	ErrorCodeFunctionFailedWithoutRollback   = ErrorCode(1)
+	ErrorCodeFunctionFailedRollbackSucceeded = ErrorCode(2)
+	ErrorCodeFunctionFailedRollbackFailed    = ErrorCode(3)
+)
+
+type Error struct {
+	ErrorCode ErrorCode `json:"errorCode"`
```
yasker, line 19: `Code`, `Message` or `Type`, `Message`, or `Info`, `Message`.

pkg/types/error.go
```diff
@@ -0,0 +1,100 @@
...
+
+func NewError(code ErrorCode, msg string) *Error {
+	return &Error{
+		ErrorCode: code,
+		ErrorMsg:  msg,
+	}
+}
+
+func (e *Error) Error() string {
+	return fmt.Sprintf("customized error: code = %v, msg = %s", e.ErrorCode, e.ErrorMsg)
```
yasker, line 31: Don't say `customized error` which `customized` doesn't mean anything.

pkg/types/error.go
```diff
@@ -0,0 +1,100 @@
...
+const (
+	ErrorCodeResultUnknown                   = ErrorCode(0)
+	ErrorCodeFunctionFailedWithoutRollback   = ErrorCode(1)
+	ErrorCodeFunctionFailedRollbackSucceeded = ErrorCode(2)
+	ErrorCodeFunctionFailedRollbackFailed    = ErrorCode(3)
+)
+
+type Error struct {
+	ErrorCode ErrorCode `json:"errorCode"`
+	ErrorMsg  string    `json:"errorMsg"`
```
yasker, line 20: Why do you decide not to store two errors/strings instead?

pkg/types/error.go
```diff
@@ -0,0 +1,100 @@
...
+				retErr = fmt.Errorf("%v; %v", retErr, err)
+			} else {
+				retErr = err
+			}
+		}
+	}
+	return retErr
+}
+
+func GetFinalError(functionErr, rollbackErr error) error {
```
yasker, line 70: `GenerateFunctionErrorWithRollback`

pkg/types/error.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	if rollbackErr != nil {
+		return NewError(ErrorCodeResultUnknown,
+			fmt.Sprintf("function succeeds but there is rollback error: %v", rollbackErr))
+	}
+
+	return nil
+}
+
+func UnmarshalGRPCError(err error) error {
```
yasker, line 87: Can we just create an error with two errors? We can type convert grpc error back instead of depending on the error message format for parsing.

shuo-wu: If we use the customized error as the gRPC server output, the error will be wrapped as a gRPC error automatically and the gRPC client always receives gRPC errors. That's why I convert the error as JSON string then unmarshal it in gRPC client.

yasker: It seems gRPC errors can contain more details. See https://grpc.io/docs/guides/error/ and https://jbrandhorst.com/post/grpc-errors/

integration/data/test_ha.py
```diff
@@ -524,3 +531,159 @@ def test_expansion_with_rebuild(grpc_controller,  # NOQA
...
+    """
+    The test flow:
+    1. Write random data into the block device.
+    2. Create the 1st snapshot.
+    3. Create an empty directory using the tmp meta file path of
+       the expansion disk for each replica.
+       This will fail the following expansion and trigger expansion rollback.
+    4. Try to expand the volume but fails. Then the automatic rollback will
+       be applied implicitly.
+    5. Check the volume status and if there are leftovers of
```
yasker, line 548: Make sure to check the volume size as well.

integration/data/test_ha.py
```diff
@@ -524,3 +531,159 @@ def test_expansion_with_rebuild(grpc_controller,  # NOQA
...
+       the failed expansion.
+    6. Check if the volume is still usable by r/w data,
+       then create the 2nd snapshot.
+    7. Retry expansion. It should succeed.
+    8. Verify the data and try data r/w.
```
yasker, line 553: Same here, check the volume size.

integration/data/test_ha.py
shuo-wu: It's checked in `wait_for_volume_expansion()`

integration/data/test_ha.py
shuo-wu: Same here, it's checked in `wait_for_volume_expansion()`. I will add the comment `Check the volume status` in step 6

PR #474, opened 2020-04-16
Correct gRPC naming for replica and syncagent services

shuo-wu:
https://github.com/longhorn/longhorn/issues/1197

### DIFF
--- a/proto/ptypes/replica.proto
+++ b/proto/ptypes/replica.proto
@@ -5,101 +5,157 @@ package ptypes;
 import "google/protobuf/empty.proto";
 
 service ReplicaService {
-    rpc ReplicaCreate(ReplicaCreateRequest) returns (Replica) {}
-    rpc ReplicaDelete(google.protobuf.Empty) returns (google.protobuf.Empty) {}
-    rpc ReplicaGet(google.protobuf.Empty) returns (Replica) {}
-    rpc ReplicaOpen(google.protobuf.Empty) returns (Replica) {}
-    rpc ReplicaClose(google.protobuf.Empty) returns (Replica) {}
-    rpc ReplicaReload(google.protobuf.Empty) returns (Replica) {}
-    rpc ReplicaRevert(ReplicaRevertRequest) returns (Replica) {}
-    rpc ReplicaSnapshot(ReplicaSnapshotRequest) returns (Replica) {}
-    rpc ReplicaExpand(ReplicaExpandRequest) returns (Replica) {}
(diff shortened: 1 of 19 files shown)


proto/ptypes/replica.proto
```diff
@@ -5,70 +5,70 @@ package ptypes;
...
-    rpc ReplicaRevert(ReplicaRevertRequest) returns (Replica) {}
-    rpc ReplicaSnapshot(ReplicaSnapshotRequest) returns (Replica) {}
-    rpc ReplicaExpand(ReplicaExpandRequest) returns (Replica) {}
-    rpc DiskRemove(DiskRemoveRequest) returns (Replica) {}
-    rpc DiskReplace(DiskReplaceRequest) returns (Replica) {}
-    rpc DiskPrepareRemove(DiskPrepareRemoveRequest) returns (DiskPrepareRemoveReply) {}
-    rpc DiskMarkAsRemoved(DiskMarkAsRemovedRequest) returns (Replica) {}
-    rpc RebuildingSet(RebuildingSetRequest) returns (Replica) {}
-    rpc RevisionCounterSet(RevisionCounterSetRequest) returns (Replica) {}
+    rpc CreateReplica(CreateReplicaRequest) returns (Replica) {}
```
yasker, line 8: Why change the name?

shuo-wu: Since the old name doesn't follow the naming convention:
https://cloud.google.com/apis/design/naming_convention#method_names

shuo-wu: Since the old method name didn't follow the golang naming convention. I was thinking we can correct both together.

yasker:
Is there any protobuf convention checking tool out there? It's better if we can enforce it automatically (and add exception for ones we need to correct later), otherwise we might run into it again.

I feel we might want to push this to v1.0.0 and make sure every proto file follows the convention starting from it. And use tools in build to enforce that.

yasker:
Hmm, the issue is indeed targetted for 1.0.

shuo-wu:
Actually I found 2 available tools for the gRPC convention check:
1. https://github.com/uber/prototool. This one seems powerful, it supports `format` besides `lint`. But it's a little bit complicated, we need to config it to fit our proto files. I fail to run it without any configuration.
2. https://github.com/yoheimuta/protolint. This one is simpler and it doesn't contain the `format` tool. But we can directly run it and get the result after the installation.

yasker:
`prototool` sounds a bit better. But let's do this after the v0.8.1 release.

PR #475, opened 2020-04-18
add .idea folder to .gitignore

joshimoo:
ignore the .idea folder so devs that use goland don't accidentally push their .idea project files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@
 *.pyo
 *.swp
 
-integration/.idea/
 integration/.tox/
 integration/.venv/
 integration/MANIFEST
@@ -15,3 +14,8 @@ integration/MANIFEST
 
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
+
+# ignores all goland project folders and files
+.idea/
+*.iml
+*.ipr


.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea/
```
yasker, line 1: Should it be `/.idea/`?

yasker:
And in general, I don't recommend using `git add` for whole directory (or `git commit -a`). You never know what you might add to the commit.

.gitignore
joshimoo: I used .idea/ to ignore the .idea folders in all subpaths as well.
If I used /.idea/ it would only ignore it in the root.

by using .idea/ to ignore all folders irregardless of depth with the name .idea
we could remove integration/.idea/ from the .gitignore file.

yasker: Make sense.

Then we can remove another `.idea/` line below.

yasker: Also, keep it in a separate section by itself, better add a comment too. So it will not asked about again.

joshimoo: Good idea, will amend the commit & pull request with your suggestion.

yasker: Is it how it works? Seems it doesn't work for me for other directories.

PR #478, opened 2020-04-23
Fix the revision counter for the upgrade engine

shuo-wu:
longhorn/longhorn#1235

### DIFF
--- a/pkg/controller/control.go
+++ b/pkg/controller/control.go
@@ -502,23 +502,29 @@ func (c *Controller) Start(addresses ...string) error {
 		c.setReplicaModeNoLock(address, types.RW)
 	}
 
-	revisionCounters := make(map[string]int64)
-	for _, r := range c.replicas {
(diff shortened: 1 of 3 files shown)


pkg/controller/control.go
```diff
@@ -452,7 +452,7 @@ func (c *Controller) StartFrontend(frontend string) error {
 	return c.startFrontend()
 }
 
-func (c *Controller) Start(addresses ...string) error {
+func (c *Controller) Start(isUpgrade bool, addresses ...string) error {
```
yasker, line 455: `isUpgrade` is already a field of the controller object.

yasker:
Also we need to improve the integration test in the engine for this scenario.

shuo-wu:
Since both data writing and the upgrade call are sync ops for the engine test, we will add the test (keep writing data during live upgrade) in manager integration test or manual validation steps.

PR #479, opened 2020-04-24
Update go-iscsi-helper to get fix for #1223

joshimoo:
Also added an integration test for correct device selection & creation
which was the underlying problem of longhorn/longhorn#1223

### DIFF
--- a/integration/common/core.py
+++ b/integration/common/core.py
@@ -18,7 +18,7 @@ from common.util import read_file, checksum_data
 from common.frontend import blockdev, get_block_device_path
 
 from common.constants import (
-    LONGHORN_BINARY, LONGHORN_UPGRADE_BINARY,
+    LONGHORN_BINARY, LONGHORN_UPGRADE_BINARY, LONGHORN_DEV_DIR,
     VOLUME_NAME, VOLUME_BACKING_NAME,
     SIZE, PAGE_SIZE, SIZE_STR,
     BACKUP_DIR, BACKING_FILE,
@@ -165,7 +165,7 @@ def cleanup_backend_file(paths):
 
 
 def get_dev_path(name):
-    return os.path.join("/dev/longhorn/", name)
+    return os.path.join(LONGHORN_DEV_DIR, name)
 
 
 def get_expansion_snapshot_name():
--- a/integration/data/conftest.py
+++ b/integration/data/conftest.py
@@ -1,5 +1,6 @@
 import pytest
 import os
+import string
 import tempfile
 
 from common.constants import INSTANCE_MANAGER_REPLICA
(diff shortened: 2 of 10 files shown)


integration/data/conftest.py
```diff
@@ -49,6 +50,38 @@ def grpc_controller_no_frontend(request, grpc_controller_client):
...
+    # we find the next available scsi name based on the current
+    # nodes in the /dev filesystem which isn't fool proof
+    device_prefix = "/dev/sd"
+    first_available_device = "does_not_exist"
+    for i in list(string.ascii_lowercase):
+        dev = device_prefix + i
+        if not os.path.exists(dev):
+            first_available_device = dev
+            break
+
```
yasker, line 64: Add `assert first_available_device != ""`

integration/data/conftest.py
```diff
@@ -49,6 +50,38 @@ def grpc_controller_no_frontend(request, grpc_controller_client):
                                   frontend="")
 
 
+@pytest.fixture
+def first_available_device(request):
+    # we find the next available scsi name based on the current
+    # nodes in the /dev filesystem which isn't fool proof
+    device_prefix = "/dev/sd"
+    first_available_device = "does_not_exist"
```
yasker, line 58: Can do `= ""`

integration/data/conftest.py
```diff
@@ -49,6 +50,38 @@ def grpc_controller_no_frontend(request, grpc_controller_client):
...
+        dev = device_prefix + i
+        if not os.path.exists(dev):
+            first_available_device = dev
+            break
+
+    return first_available_device
+
+
+@pytest.fixture
+def grpc_controller_volume_test(request, first_available_device,
```
yasker, line 69: `grpc_controller_device_name_test`

integration/data/conftest.py
```diff
@@ -49,6 +50,38 @@ def grpc_controller_no_frontend(request, grpc_controller_client):
                                   frontend="")
 
 
+@pytest.fixture
```
yasker, line 53: Move those two fixtures to `test_basic_ops.py` (if possible) since it's unlikely they will be shared by others.

integration/data/test_basic_ops.py
```diff
@@ -17,6 +19,17 @@
...
+                         grpc_controller_volume_test,  # NOQA
+                         grpc_replica1, grpc_replica2):  # NOQA
+    assert get_dev(grpc_replica1, grpc_replica2,
+                   grpc_controller_volume_test)
+    check_dev_existence(VOLUME_NAME)
+    longhorn_dev = get_dev_path(VOLUME_NAME)
+    dev_info = os.stat(first_available_device)
+    assert dev_info.st_rdev == os.stat(os.devnull).st_rdev
+    assert dev_info.st_rdev != os.stat(longhorn_dev).st_rdev
+
```
yasker, line 32: add a basic data rw test here, e.g. use `verify_data`

integration/common/core.py
```diff
@@ -18,7 +18,7 @@
 from common.frontend import blockdev, get_block_device_path
 
 from common.constants import (
-    LONGHORN_BINARY, LONGHORN_UPGRADE_BINARY,
+    LONGHORN_BINARY, LONGHORN_UPGRADE_BINARY, LONGHORN_DEV_DIR,
```
yasker, line 21: What this for?

integration/data/conftest.py
joshimoo: I certainly can move them, but it looks like we always use the conftest.py for fixtures.
By moving them we are breaking convention, do you want me to still move them?

yasker: You have a point. If we moved them, the other contributors might get confused on where to put the fixtures. Keep them here then.

integration/data/conftest.py
joshimoo: I did this on purpose, so that if for whatever reason we cannot find an available device we don't try creating a link of "" to /dev/null

integration/data/conftest.py
joshimoo: See below

yasker: Though this `assert` should fail the test immediately.

joshimoo: fair point, will change it.

PR #484, opened 2020-04-30
Remove no longer known replicas from the backup status check

joshimoo:
Only a replica in RW mode can create backups, so there is no point in
checking other replicas for their backup status.

longhorn/longhorn#1230

### DIFF
--- a/app/cmd/backup.go
+++ b/app/cmd/backup.go
@@ -13,6 +13,7 @@ import (
 	"github.com/longhorn/backupstore/cmd"
 
 	"github.com/longhorn/longhorn-engine/pkg/sync"
+	"github.com/longhorn/longhorn-engine/pkg/types"
 	"github.com/longhorn/longhorn-engine/pkg/util"
 )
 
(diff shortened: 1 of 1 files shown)


app/cmd/backup.go
```diff
@@ -82,16 +83,53 @@ func getBackupStatus(c *cli.Context, backupID string, replicaAddress string) (*s
...
+		return fmt.Errorf("failed to get list of backupIDs: %v", err)
+	}
+
+	knownReplicas, err := getKnownReplicas(c)
+	if err != nil {
+		return err
 	}
 
 	for backupID, replicaAddress := range backupReplicaMap {
+		// a replica is only valid in Read-Write state.
```
yasker, line 116: Just put: 
```
// Only a replica in RW mode can create backups.
```

app/cmd/backup.go
```diff
@@ -82,16 +83,53 @@ func getBackupStatus(c *cli.Context, backupID string, replicaAddress string) (*s
 	return backupStatus, nil
 }
 
+func getKnownReplicas(c *cli.Context) (map[string]types.Mode, error) {
```
yasker, line 86: `getReplicaModeMap` sounds like a better name, indicates what's in the result.

app/cmd/backup.go
```diff
@@ -82,16 +83,53 @@ func getBackupStatus(c *cli.Context, backupID string, replicaAddress string) (*s
...
 	backupProgressList := make(map[string]*sync.BackupStatusInfo)
 
 	cli := getCli(c)
 	backupReplicaMap, err := cli.BackupReplicaMappingGet()
 	if err != nil {
-		return fmt.Errorf("failed to get list of backupIDs:%v", err)
+		return fmt.Errorf("failed to get list of backupIDs: %v", err)
+	}
+
+	knownReplicas, err := getKnownReplicas(c)
```
yasker, line 110: just `replicas`, or `replicaModeMap`

app/cmd/backup.go
```diff
@@ -82,16 +83,53 @@ func getBackupStatus(c *cli.Context, backupID string, replicaAddress string) (*s
...
+	if err != nil {
+		return err
 	}
 
 	for backupID, replicaAddress := range backupReplicaMap {
+		// a replica is only valid in Read-Write state.
+		// during a rebuild (Write only) the memory information is lost,
+		// so there is no point in requesting the backup status
+		if mode, known := knownReplicas[replicaAddress]; !known || mode != types.RW {
```
yasker, line 119: Don't need to check `known`. It's a string in the value so it won't be nil.

app/cmd/backup.go
```diff
@@ -82,16 +83,53 @@ func getBackupStatus(c *cli.Context, backupID string, replicaAddress string) (*s
...
+			err := cli.BackupReplicaMappingDelete(backupID)
```
yasker, line 120: I think we might want to revisit our backup status cleanup policy later. Currently, if a backup was interrupted during the process, the error information with it will be gone as well very soon.

app/cmd/backup.go
```diff
@@ -82,16 +83,53 @@ func getBackupStatus(c *cli.Context, backupID string, replicaAddress string) (*s
...
+			if err != nil {
+				return err
+			}
+			backupProgressList[backupID] = &sync.BackupStatusInfo{
+				ReplicaAddress: replicaAddress,
+				State:          "error",
+				Error: fmt.Sprintf("Failed to get backup status on %s for %v: %v",
+					replicaAddress, backupID, "replica is no longer known"),
```
yasker, line 128: Reason: `unknown replica`.

app/cmd/backup.go
joshimoo: Agree but figured it's something for future work.

app/cmd/backup.go
joshimoo: Not sure if you wanted reason to be part of the message, so I left it out because of the formatting.

app/cmd/backup.go
```diff
@@ -144,6 +186,15 @@ func checkBackupStatus(c *cli.Context) error {
 		return fmt.Errorf("replica address is empty")
 	}
 
+	knownReplicas, err := getReplicaModeMap(c)
+	if err != nil {
+		return err
+	}
+	if mode, known := knownReplicas[replicaAddress]; !known || mode != types.RW {
```
yasker, line 193: Need to update this part as well. Also, use `mode :=` instead of `mode, _ :=` since we normally omit the existence check if it's not used for map.

yasker:
LGTM. Wait for the integration test.

joshimoo:
integration test longhorn/longhorn-tests#306

PR #485, opened 2020-05-02
Update longhorn/backupstore to v0.0.0-20200502041801-8d1f61550db9

joshimoo:
longhorn/longhorn#1266


yasker:
Need to add an integration test for engine as well. Or we can modify an existing test to cover this case.

joshimoo:
Added an integration test that checks that we are doing a full backup when we are missing the latest backup on S3

PR #487, opened 2020-05-07
The engine will mark the replica as ERROR if the related backend creation fails

shuo-wu:
longhorn/longhorn#995
longhorn/longhorn#1253

### DIFF
--- a/integration/core/test_cli.py
+++ b/integration/core/test_cli.py
@@ -6,6 +6,7 @@ import subprocess
 import json
 import datetime
 import pytest
+import tempfile
 
 from urllib.parse import urlparse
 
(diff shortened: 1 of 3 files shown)


yasker:
LGTM.

We also need engine integration test.

What would be the impact when manager saw all replicas are error when starting? Need to make sure auto reattach won't be triggered.

shuo-wu:
This PR is not enough for the continuous auto reattachment issue:
When all replicas are not available in the engine starting stage, the engine process will directly error out, and the engine monitor has no chance to get the replica status (mode ERROR). Then the volume will be reattached rather being Faulted status.

But if there are some of the replicas not available during the engine process launching, this PR works fine. The volume won't wrongly loop in detach and reattach.

yasker:
If there is no replica was healthy even once, `HealthyAt` won't be set. If no replica was healthy before auto reattached/salvage, we don't need to auto attach/salvage.

shuo-wu:
> If there is no replica was healthy even once, `HealthyAt` won't be set. If no replica was healthy before auto reattached/salvage, we don't need to auto attach/salvage.

This solution doesn't work. Since the volume had been attached before the snapshots corrupt, all replicas' `HealthyAt` are not empty then the solution cannot prevent auto attachment. We need a way to figure out if the engine process is crashed during the launching stage. If Yes, auto reattachment/salvage should be disabled.

yasker:
OK, if we need to reopen #995 to track fail to start the engine the second-time case, we can do that. Or file a new issue if that's more appropriate.

Just cover #1253 for now.

Still need engine integration test.

shuo-wu:
The summary of this PR:
1. If the engine finds the invalid replicas during starting, it will add the invalid replicas into the backend list but with backend client `nil` and the backend mode `ERR`, instead of directly erroring out. Then the engine can still work and the invalid replica instances will be shut down by the manager later.
2. An engine integration test is added for this case.

PR #488, opened 2020-05-07
sync: Fix deleting snapshot progress showing over 100%

boknowswiki:
https://github.com/longhorn/longhorn/issues/1261

### DIFF
--- a/pkg/sync/rpc/server.go
+++ b/pkg/sync/rpc/server.go
@@ -85,16 +85,14 @@ type PurgeStatus struct {
 	total     int
 }
 
-func NewPurgeStatus() *PurgeStatus {
-	return &PurgeStatus{
-		// avoid possible division by zero
-		total: 1,
-	}
-}
-
 func (ps *PurgeStatus) UpdateFoldFileProgress(progress int, done bool, err error) {
 	ps.Lock()
-	ps.Progress = int((float32(ps.processed)/float32(ps.total) + float32(progress)/(float32(ps.total)*100)) * 100)
+	// Avoid possible division by zero, also total 0 means nothing to be done
+	if ps.total == 0 {
+		ps.Progress = 100
+	} else {
+		ps.Progress = int((float32(ps.processed)/float32(ps.total) + float32(progress)/(float32(ps.total)*100)) * 100)
+	}
 	ps.Unlock()
 }
 
@@ -830,7 +828,7 @@ func (s *SyncAgentServer) postIncrementalRestoreOperations(restoreStatus *replic
 
 	logrus.Infof("Cleaning up incremental restore by Coalescing and removing the file")
 	// coalesce delta file to snapshot/disk file
-	ops := NewPurgeStatus()
+	ops := &PurgeStatus{}
 	if err := sparse.FoldFile(deltaFileName, restoreStatus.SnapshotDiskName, ops); err != nil {
 		logrus.Errorf("Failed to coalesce %s on %s: %v", deltaFileName, restoreStatus.SnapshotDiskName, err)
 		return err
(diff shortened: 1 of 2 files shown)


pkg/sync/rpc/server.go
```diff
@@ -87,14 +87,18 @@ type PurgeStatus struct {
 
 func NewPurgeStatus() *PurgeStatus {
 	return &PurgeStatus{
-		// avoid possible division by zero
-		total: 1,
+		RWMutex: sync.RWMutex{},
```
yasker, line 90: Seems we don't really need this function?

Also, `RWMutex: sync.RWMutex{},` shouldn't be necessary. It's not a pointer.

Same below for the other statuses.

pkg/sync/rpc/server.go
```diff
@@ -1000,8 +1002,12 @@ func (s *SyncAgentServer) purgeSnapshots() (err error) {
 		s.PurgeStatus.Progress = int(float32(removed) / float32(markedRemoved) * 100)
 		s.PurgeStatus.Unlock()
 	}
+
+	// Purge is done, clear total and processed
```
yasker, line 1006: Why do we need to do this?

pkg/sync/rpc/server.go
boknowswiki: I think we need this function, this function helps us to aggregate the usage of PurgeStatus. I will remove the RWMutex field.

pkg/sync/rpc/server.go
boknowswiki: do you mean the comment or the code to clear total and processed?
We need to clear the processed, currently the reason we have this issue is because after the first purge, the processed becomes to 1, and then we hit the over 100% issue.

Also I will add more explanation on the commit message.

pkg/sync/rpc/server.go
yasker: Currently, there is no other usage for this function, so we can just do `PurgeStatus: &PurgeStatus{}` below.

pkg/sync/rpc/server.go
```diff
@@ -146,9 +150,7 @@ func NewSyncAgentServer(startPort, endPort int, replicaAddress string) *SyncAgen
 		BackupList: &BackupList{
 			RWMutex: sync.RWMutex{},
```
yasker, line 151: As I mentioned before, those RWMutex initialization are not necessary as well.

pkg/sync/rpc/server.go
yasker: Oh, so it's prepared for the next purge? That makes sense, but it should also done by the purge initialization part as well.

pkg/sync/rpc/server.go
boknowswiki: Sorry, my bad. There is one place using the NewPurgeStatus:
https://github.com/longhorn/longhorn-engine/blob/4bf6f2b65fc8b64898ff76b788fe1dfb30d54003/pkg/sync/rpc/server.go#L833

yasker: This can be directly assigned too since we no longer do special initialization for PurgeStatus.

boknowswiki:
Passed local integration test:

================== 74 passed, 1 skipped in 1228.47s (0:20:28) ==================
___________________________________ summary ____________________________________
  flake8: commands succeeded
  py36: commands succeeded
  congratulations :)

Also manually tested the fix, don't see the over 100% progress any more.

pkg/sync/rpc/server.go
yasker: Hmm? Where is this part?

yasker: Can you add another commit to clean up those unnecessary RWMutex initializations?

boknowswiki:
Unnecessary RWMutex initializations removed.

PR #489, opened 2020-05-08
Add integration test for incremental backup logic

joshimoo:
We had some issues with incremental / full backups, this test covers
cases where we create a duplicate volume for a prior existing volume

longhorn/longhorn#1294
longhorn/backupstore#34

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/gorilla/handlers v1.4.2
 	github.com/gorilla/mux v1.7.3
 	github.com/gorilla/websocket v1.4.0 // indirect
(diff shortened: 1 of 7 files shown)


joshimoo:
will add the go mod update once longhorn/backupstore#34 is merged

yasker:
Add the backup store change before the integration test, so the integration test won't fail during `git bisect`

joshimoo:
> Add the backup store change before the integration test, so the integration test won't fail during `git bisect`

I already did, unless I screwed it up?

joshimoo:
Yep looks good on my end.

yasker:
Yeah it looks good when I checked it out locally. Seems it's that GitHub bug again:

Screen Shot 2020-05-07 at 8 53 45 PM

PR #490, opened 2020-05-08
Add integration test for corrupt s3 backup config

joshimoo:
longhorn/backupstore#35
longhorn/longhorn#1212

### DIFF
--- a/integration/core/test_cli.py
+++ b/integration/core/test_cli.py
@@ -798,11 +798,73 @@ def backup_core(bin, engine_manager_client,  # NOQA
     assert backup_list[backup2]["Size"] == backup2_info["Size"]
     assert backup_list[backup2]["Created"] == backup2_info["Created"]
 
-    # test backup volume list
-    # https://github.com/rancher/longhorn/issues/399
+    # test that corrupt backups are signaled during a list operation
+    # https://github.com/longhorn/longhorn/issues/1212
+    cmd = [bin, '--url', grpc_controller_client.address,
+           'backup', 'ls', backup_target]
+    data = subprocess.check_output(cmd, env=env)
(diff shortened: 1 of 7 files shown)


integration/core/test_cli.py
```diff
@@ -798,11 +798,63 @@ def backup_core(bin, engine_manager_client,  # NOQA
...
+    assert volume_info[VOLUME_NAME]["Backups"] is not None
+    backups = volume_info[VOLUME_NAME]["Backups"]
+    assert len(backups) == 1
+    assert backup1_info["URL"] in backups
+    assert backup2_info["URL"] not in backups
+
+    # will fail because of corrupt json data
+    cmd = [bin, '--url', grpc_controller_client.address,
+           'backup', 'inspect', backup2_info["URL"]]
+    with pytest.raises(subprocess.CalledProcessError):
```
yasker, line 848: Shoud `inspect` fail or return some error message in the json?

joshimoo: By doing with with pytest.raises, it will check that the command has failed.
Which happens on the engine side when an error is returned.

joshimoo: Since inspect is trying to load a backup.cfg that either doesn't exist or something went wrong during loading. Both cases lead to failure of inspect.

Both cases after the fix will be ignored, since we don't want the full list to fail just because a single backup.cfg couldn't be loaded.

joshimoo:
Need to add the updated go mod once longhorn/backupstore#35 is merged

yasker:
LGTM, wait for the backupstore update.

joshimoo:
rebased and added the updated backupstore

PR #493, opened 2020-05-15
Refactor the frontend start functions and add cleanup during the frontend init

shuo-wu:
longhorn/longhorn#1240


yasker:
LGTM. Wait for the vendor update.

shuo-wu:
The vender is updated.

PR #498, opened 2020-06-04
Update backupstore for blk fixes

joshimoo:
Integration tests for the backup delete changes.

We now have a test that verifies the correct gc operation for out of
order deletes and dependent blocks. We also now have a test that makes
sure that blocks are not deleted while there is an in progress backup.

longhorn/longhorn#1431 
longhorn/longhorn#1433 

With the orphaned block deletion in place we can now remove corrupted
backup.cfg files, since that will just orphan the blocks and they will
be cleaned up with all the other orphaned blocks.
With the removal of the volume.cfg existence check we can now also
cleanup a volume folder including all subfolders when requested via a
backup volume removal.

longhorn/longhorn#1213

depends on longhorn/backupstore#41

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/gorilla/handlers v1.4.2
 	github.com/gorilla/mux v1.7.3
 	github.com/gorilla/websocket v1.4.0 // indirect
-	github.com/longhorn/backupstore v0.0.0-20200530020211-e71b773c1805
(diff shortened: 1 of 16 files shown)


joshimoo:
will update the go mod to the released version once longhorn/backupstore#41 is merged.

yasker:
https://github.com/longhorn/backupstore/pull/41 merged.

joshimoo:
updated the backup store to the release version

PR #499, opened 2020-06-04
Add test for backup volume list

joshimoo:
The test makes sure that if there is failure when trying to retrieve the
backup names for that volume, that the list operation does not fail.
Since a failure would block the list operation of returning information
for all the other backup volumes. Instead of failing the whole operation
we add a Error message to the backup volume.

longhorn/longhorn#1410

### DIFF
--- a/integration/data/test_backup.py
+++ b/integration/data/test_backup.py
@@ -15,12 +15,15 @@ from data.snapshot_tree import (
 )
 from common.constants import (
     VOLUME_NAME, VOLUME_BACKING_NAME,
+    VOLUME2_NAME, ENGINE2_NAME, REPLICA_2_NAME,
     ENGINE_NAME, ENGINE_BACKING_NAME, BACKUP_DIR,
     BLOCK_SIZE, BLOCK_SIZE_STR,
 )
 from common.util import (
     finddir, findfile
 )
+
+MESSAGE_TYPE_ERROR = "error"
 
 
 def backup_test(dev, address,  # NOQA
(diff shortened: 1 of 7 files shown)


integration/data/test_backup.py
```diff
@@ -664,6 +667,50 @@ def test_backup_volume_deletion(grpc_replica1, grpc_replica2,  # NOQA
...
+        # place badly named backup.cfg file
+        # we want the list call to return correctly and
+        # include an error message otherwise a single volume error
+        # can stop all backup volumes from showing up
+        backup_dir = os.path.join(finddir(BACKUP_DIR, VOLUME_NAME), "backups")
+        cfg = open(os.path.join(backup_dir, "backup_1234@failure.cfg"), "w")
+        cfg.close()
+        info = cmd.backup_volume_list(address, VOLUME_NAME, backup_target,
+                                      include_backup_details=True)
+        pprint(info)
```
yasker, line 703: ?

integration/data/test_backup.py
```diff
@@ -664,6 +667,50 @@ def test_backup_volume_deletion(grpc_replica1, grpc_replica2,  # NOQA
...
+        assert "Invalid name" in info[VOLUME_NAME]["Messages"]["error"]
```
yasker, line 704: We still need to check if the other volumes are shown? There is only one volume we have here.

integration/data/test_backup.py
```diff
@@ -664,6 +667,50 @@ def test_backup_volume_deletion(grpc_replica1, grpc_replica2,  # NOQA
         cleanup_replica(grpc_replica1)
         cleanup_replica(grpc_replica2)
 
+def test_backup_volume_list(grpc_replica1, grpc_replica2,  # NOQA
+                            grpc_controller, backup_targets):  # NOQA
```
yasker, line 671: Can we add the test description as e2e cases? We are going to add the test description for the engine integration test as well.

integration/data/test_backup.py
joshimoo: If we are displaying one, we will display the others. Since we no longer error and go through the regular control flow.

integration/data/test_backup.py
joshimoo: Sure I will add the descriptions. I have to say though that I am a little concerned that there is a lot of redundancy in all of the test work for the backupstore area (2x test code + 2x test walkthrough which more or less end up doing the exact same thing)?

integration/data/test_backup.py
joshimoo: Leftover removed

integration/data/test_backup.py
joshimoo: Will look into using multiple volumes in the test.

integration/data/test_backup.py
yasker: It's like two lines of defense.

The first line is the core (engine). The test will be run on every build and the PR must pass the build in order to be merged. Not every change has impacts on the core, but whenever there is an impact, we need to be cautious and make sure the core logic works as expected.

The second line is end to end. It will involve the whole system and surrounding them around the core. Since other components may fail as well, so even the core works, there is no guarantee the end to end test will work.

But of course, if the core doesn't work, there is no point to test end to end.

In an ideal world, we don't need the test for core separately, since the end to end test should cover everything. But for two reasons we still want to improve the testing for the core:
1. There are corner cases that's hard to test in the e2e level, but easy to do in the core level.
2. We want to catch issues as soon as possible for the core since it's important to keep the foundation work solid.

So the engine integration test is essentially the price we paid to keep the core as stable as possible. There might be duplicates like this in case, but I would prefer to pay that price and factor it in as the part of the cost to keep the core stable.

Does that make sense?

joshimoo: I agree with having the engine core test, since it's handling all the backup business logic.
I agree with the layered defense argument, but I do think it comes with a cost, as long as we are willing to pay that cost we should continue doing so. But I do think we should consider the surface area when deciding whether a e2e (manager) test adds more value then it costs, when the core tests covers the whole business logic and the manager doesn't really have any business logic. Thanks for providing the context and explanation on this, it's not meant as a criticism :)

I will add the extra tests.

yasker: I am considering the end to end test to be the replacement for the manual test in general. If there is no end to end test, QA would need to validate those cases manually since we won't test the engine alone. In that sense, it saves time if we can automate the cases like this.

PR #501, opened 2020-06-10
Restore Rafactoring

shuo-wu:
https://github.com/longhorn/longhorn/issues/1188

Summary:
1. Refactor the implementation of restore:
    1. Check the parameters.
    2. Prepare to restore: Initiate the restore status and set the fields. Then all restore errors after the preparation will be recorded in the restore status.
    3. Start to restore: prepare the restore file and start to fetch inf/data from the backupstore.
    4. Finish restore: wait for the restore complete, reload/revert the snapshot, use a defer function to update the restore status and the fields.
2. Format the restore errors for replica syne agent servers and return 
3. Update the integration test: Test the error during the starting stage will be recorded in the restore status; check the error from sync agent servers are formated and printed.

### DIFF
--- a/app/cmd/backup.go
+++ b/app/cmd/backup.go
@@ -222,6 +222,12 @@ func BackupRestoreCmd() cli.Command {
 		},
 		Action: func(c *cli.Context) {
 			if err := restoreBackup(c); err != nil {
+				errInfo, jsonErr := json.MarshalIndent(err, "", "\t")
+				if jsonErr != nil {
+					logrus.Errorf("Cannot marshal err [%v] to json: %v", err, jsonErr)
+				} else {
+					fmt.Println(string(errInfo))
+				}
 				logrus.Fatalf("Error running restore backup command: %v", err)
 			}
 		},
@@ -336,7 +342,6 @@ func doRestoreBackupIncrementally(c *cli.Context) error {
 	lastRestored := c.String("last-restored")
 
 	if err := task.RestoreBackupIncrementally(backupURL, backupName, lastRestored, credential); err != nil {
-		logrus.Errorf("failed to perform incremental restore: %v", err)
 		return err
 	}
 	return nil
--- a/integration/data/test_restore_inc.py
+++ b/integration/data/test_restore_inc.py
@@ -110,15 +110,62 @@ def restore_inc_test(grpc_engine_manager,  # NOQA
     start_no_frontend_volume(grpc_engine_manager,
                              grpc_dr_controller,
                              grpc_dr_replica1, grpc_dr_replica2)
+    # mock restore crash/error:
+    # By adding attribute `immutable`, Longhorn cannot create a file
+    # for the restore. Then the following restore command will fail.
+    command = ["chattr", "+i", FIXED_REPLICA_PATH1]
+    subprocess.check_output(command).strip()
+    command = ["chattr", "+i", FIXED_REPLICA_PATH2]
+    subprocess.check_output(command).strip()
+    with pytest.raises(subprocess.CalledProcessError) as e:
+        cmd.backup_restore(dr_address, backup0)
+    assert "operation not permitted" in e.value.stdout
+    command = ["chattr", "-i", FIXED_REPLICA_PATH1]
+    subprocess.check_output(command).strip()
+    command = ["chattr", "-i", FIXED_REPLICA_PATH2]
+    subprocess.check_output(command).strip()
 
     cmd.backup_restore(dr_address, backup0)
     wait_for_restore_completion(dr_address, backup0)
     verify_no_frontend_data(grpc_engine_manager,
                             0, snap0_data, grpc_dr_controller)
 
-    # mock restore crash/error
     delta_file1 = "volume-delta-" + backup0_name + ".img"
+
+    # mock inc restore crash/error: cannot clean up the delta file path
+    command = ["mkdir", "-p", FIXED_REPLICA_PATH1 + delta_file1 + "/dir"]
+    subprocess.check_output(command).strip()
+    command = ["mkdir", "-p", FIXED_REPLICA_PATH2 + delta_file1 + "/dir"]
+    subprocess.check_output(command).strip()
+    with pytest.raises(subprocess.CalledProcessError):
+        cmd.restore_inc(dr_address, backup1, backup0_name)
+    command = ["rm", "-r", FIXED_REPLICA_PATH1 + delta_file1]
+    subprocess.check_output(command).strip()
+    command = ["rm", "-r", FIXED_REPLICA_PATH2 + delta_file1]
+    subprocess.check_output(command).strip()
+
     if "vfs" in backup_target:
+        # restore status should contain the error info
+        failed_restore, finished_restore = 0, 0
+        for i in range(RETRY_COUNTS):
+            failed_restore, finished_restore = 0, 0
+            rs = cmd.restore_status(dr_address)
+            for status in rs.values():
+                if status['backupURL'] != backup1:
+                    break
+                if 'error' in status.keys():
+                    if status['error'] != "":
+                        assert 'failed to clean up the existing file' in\
+                               status['error']
+                        failed_restore += 1
+                if not status["isRestoring"]:
+                    finished_restore += 1
+            if failed_restore == 2 and finished_restore == 2:
+                break
+            time.sleep(RETRY_INTERVAL)
+        assert failed_restore == 2 and finished_restore == 2
+
+        # mock inc restore error: invalid block
         command = ["find", VFS_DIR, "-type", "d", "-name", VOLUME_NAME]
         backup_volume_path = subprocess.check_output(command).strip()
         command = ["find", backup_volume_path, "-name", "*blk"]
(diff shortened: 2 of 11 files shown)


pkg/backup/main.go
```diff
@@ -128,51 +129,44 @@ func DoBackupCreate(volumeName string, snapshotName string, destURL string,
...
 func DoBackupRestore(backupURL string, toFile string, restoreObj *replica.RestoreStatus) error {
-	if backupURL == "" {
-		return RequiredMissingError("backup URL")
-	}
 	backupURL = util.UnescapeURL(backupURL)
+	log.Debugf("Start restoring from %v into snapshot %v", backupURL, toFile)
 
-	if toFile == "" {
-		return RequiredMissingError("snapshot")
+	// Check snap file
```
shuo-wu, line 135: Move this check to `backupstore.RestoreDeltaBlockBackup(config)`

pkg/sync/rpc/server.go
```diff
@@ -203,23 +203,38 @@ func (s *SyncAgentServer) PrepareRestore(backupURL, toFile, snapshotDiskName, la
...
-	return s.finishRestoreNoLock(currentRestored)
+	return s.finishRestoreNoLock(currentRestored, err)
 }
 
-func (s *SyncAgentServer) finishRestoreNoLock(currentRestored string) error {
+func (s *SyncAgentServer) finishRestoreNoLock(currentRestored string, restoreErr error) (err error) {
+	defer func() {
+		if s.RestoreInfo != nil {
+			if restoreErr != nil {
+				s.RestoreInfo.UpdateRestoreStatus(s.RestoreInfo.SnapshotName, 0, restoreErr)
```
shuo-wu, line 218: Check if there is an existing error in restoreInfo

pkg/sync/rpc/server.go
```diff
@@ -597,69 +597,47 @@ func (s *SyncAgentServer) BackupRestore(ctx context.Context, req *ptypes.BackupR
...
 func (s *SyncAgentServer) completeBackupRestore() (err error) {
+	currentRestoredBackupName := ""
 	defer func() {
-		if err != nil {
-			// Reset the isRestoring flag to false
-			if err := s.FinishRestore("", err); err != nil {
-				logrus.Errorf("failed to finish backup restore: %v", err)
-				return
-			}
+		if err := s.FinishRestore(currentRestoredBackupName, err); err != nil {
```
shuo-wu, line 613: `extraErr := s.FinishRestore(currentRestoredBackupName, err)`

pkg/sync/rpc/server.go
```diff
@@ -597,69 +597,47 @@ func (s *SyncAgentServer) BackupRestore(ctx context.Context, req *ptypes.BackupR
...
-		return err
-	}
-
-	s.Lock()
-	if s.RestoreInfo == nil {
-		s.Unlock()
-		logrus.Errorf("BUG: Restore completed but object not found")
-		return fmt.Errorf("cannot find restore status")
+		logrus.Errorf("failed to restore data: %v", err)
+		// There is no need to return the error here
```
shuo-wu, line 621: Consider `lastRestored` field

pkg/sync/rpc/server.go
```diff
@@ -597,69 +597,47 @@ func (s *SyncAgentServer) BackupRestore(ctx context.Context, req *ptypes.BackupR
...
+		// since it is already recorded in `s.restoreInfo`.
+		return nil
```
shuo-wu, line 623: Warp this error and return the error

pkg/sync/sync.go
```diff
@@ -15,12 +15,19 @@ import (
...
+	VolumeHeadName     = "volume-head"
+	TaskErrorKeyVolume = "volume"
+)
 
 type Task struct {
 	client *client.ControllerClient
 }
 
+type TaskError struct {
+	Errors map[string]error
```
shuo-wu, line 28: ReplicaErrors []errors

app/cmd/backup.go
```diff
@@ -222,7 +223,12 @@ func BackupRestoreCmd() cli.Command {
 		},
 		Action: func(c *cli.Context) {
 			if err := restoreBackup(c); err != nil {
-				logrus.Fatalf("Error running restore backup command: %v", err)
+				errInfo, jsonErr := json.MarshalIndent(err, "", "\t")
+				if jsonErr != nil {
+					logrus.Errorf("Cannot marshal err [%v] to json: %v", err, jsonErr)
+				}
+				fmt.Fprintln(os.Stderr, string(errInfo))
```
shuo-wu, line 230: Try `fmt.Println()`

integration/data/test_restore_inc.py
```diff
@@ -190,9 +214,8 @@ def restore_inc_test(grpc_engine_manager,  # NOQA
     compare_last_restored_with_backup(status, backup2_name)
 
     # mock race condition
-    with pytest.raises(subprocess.CalledProcessError) as e:
+    with pytest.raises(subprocess.CalledProcessError):
         restore_incrementally(dr_address, backup1, backup0_name)
-        assert "doesn't match lastRestored" in e
```
shuo-wu, line 195: Find a way to get the error info

integration/data/test_restore_inc.py
```diff
@@ -116,9 +116,40 @@ def restore_inc_test(grpc_engine_manager,  # NOQA
     verify_no_frontend_data(grpc_engine_manager,
                             0, snap0_data, grpc_dr_controller)
 
-    # mock restore crash/error
     delta_file1 = "volume-delta-" + backup0_name + ".img"
     if "vfs" in backup_target:
+        # mock restore crash/error: cannot clean up the delta file path
```
shuo-wu, line 121: Mock similar case for restore volume tests

shuo-wu: There is no test for the regular restore volume.

yasker: There is no test? https://github.com/longhorn/longhorn-engine/blob/master/integration/data/test_backup.py#L61

shuo-wu: Sorry for missing this... 
But we can test the regular restore command in this function, too.

shuo-wu: Sorry for missing this... 
But we can test the regular restore command in this function, too.

pkg/sync/sync.go
```diff
@@ -36,6 +47,42 @@ type ReplicaRebuildStatus struct {
 	FromReplicaAddress string `json:"fromReplicaAddress"`
 }
 
+func NewTaskError(res ...*ReplicaError) *TaskError {
```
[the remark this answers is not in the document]
shuo-wu: We can use value semantic for `ReplicaError` only.
There is a method `Append` for `TaskError` and this method will cause mutation. In order to keep consistency, it's better to use pointer semantic in this factory function.

app/cmd/backup.go
```diff
@@ -222,6 +222,12 @@ func BackupRestoreCmd() cli.Command {
 		},
 		Action: func(c *cli.Context) {
 			if err := restoreBackup(c); err != nil {
+				errInfo, jsonErr := json.MarshalIndent(err, "", "\t")
```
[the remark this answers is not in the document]
shuo-wu: I guess you are worrying that the printed `jsonErr` will intervene in the result.

Actually, all info printed by `logrus` is in stderr. And the error info we want to capture is printed by `fmt.Println` and in stdout.

pkg/replica/backup.go
```diff
@@ -59,15 +59,34 @@ func (rr *RestoreStatus) UpdateRestoreStatus(snapshot string, rp int, re error)
 	rr.SnapshotName = snapshot
 	rr.Progress = rp
 	if re != nil {
-		rr.Error = re.Error()
+		if rr.Error != "" {
+			rr.Error = fmt.Sprintf("%v: %v", re.Error(), rr.Error)
```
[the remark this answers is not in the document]
shuo-wu: Notice that `rr.Error` is type `string`. You mean `rr.Error = fmt.Errorf("%v: %v", re.Error(), rr.Error).Error()`? I don't think this can improve the readibility.

pkg/sync/sync.go
```diff
@@ -36,6 +47,42 @@ type ReplicaRebuildStatus struct {
...
+}
+
+func (e *TaskError) Error() string {
+	var errs []string
+	for _, re := range e.ReplicaErrors {
+		errs = append(errs, re.Error())
+	}
+
+	if errs == nil {
+		return "Unknown"
```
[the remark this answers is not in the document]
shuo-wu: If there is no error, the error object itself should be `nil` rather than containing an empty error message. Hence we shouldn't return/use this `TaskError` if there is no `ReplicaError`.

BTW, the definition of `TaskError` is similar to this `https://github.com/longhorn/longhorn-engine/blob/c10cd31baccd8008c25d5f1befc26196cb6c2622/pkg/controller/replicator.go#L34`

shuo-wu:
Summary:
1. Refactor the implementation of restore:
    1. Check the parameters.
    2. Prepare to restore: Initiate the restore status and set the fields. Then all restore errors after the preparation will be recorded in the restore status.
    3. Start to restore: prepare the restore file and start to fetch inf/data from the backupstore.
    4. Finish restore: wait for the restore complete, reload/revert the snapshot, use a defer function to update the restore status and the fields.
2. Format the restore errors for replica syne agent servers and return 
3. Update the integration test: Test the error during the starting stage will be recorded in the restore status; check the error from sync agent servers are formated and printed.

yasker:
LGTM. @joshimoo ?

PR #502, opened 2020-06-11
Backup list operation should ignore badly named files

joshimoo:
We should ignore badly named files instead of erroring, since the files don't match our search pattern we don't need to concern ourselves with them (we filter them out)

depends on longhorn/backupstore#42
longhorn/longhorn#1410


joshimoo:
will update to released backupstore version once longhorn/backupstore#42 is merged

joshimoo:
@yasker updated to the release version

PR #507, opened 2020-07-09
Update backupstore dependency

ttpcodes:
This PR updates the `vendor`, including updating longhorn/backupstore to de8e773, which introduces a check for NFSv3 in case the NFSv4 mount fails.

`github.com/pkg/errors` has also been updated as a result of running `update-longhorn-deps.sh`.


yasker:
Let's put it on hold for a bit, since Bo and Joshua's change should be in soon too.

PR #508, opened 2020-07-09
Update backupstore to v0.0.0-20200709195833-7392333d6d22

boknowswiki:
Update backupstore to v0.0.0-20200709195833-7392333d6d22 for fixing the
LastBackup fields in volume.cfg when deleting the latest backup

https://github.com/longhorn/longhorn/issues/1381

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/gorilla/handlers v1.4.2
 	github.com/gorilla/mux v1.7.3
 	github.com/gorilla/websocket v1.4.0 // indirect
-	github.com/longhorn/backupstore v0.0.0-20200709175939-de8e773a2a9d
+	github.com/longhorn/backupstore v0.0.0-20200709195833-7392333d6d22
 	github.com/longhorn/go-iscsi-helper v0.0.0-20200601183015-ca8b762c2f38
 	github.com/longhorn/sparse-tools v0.0.0-20191231185723-50ffab4d4329
 	github.com/mattn/go-colorable v0.1.4 // indirect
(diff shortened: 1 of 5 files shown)


yasker:
test_backup.py failed. Check the drone log for details.

yasker:
I've retriggered the test. The failed result is at https://drone-pr.rancher.io/longhorn/longhorn-engine/681

joshimoo:
@boknowswiki @yasker The test failure happens, because it's expecting a full backup creation since we removed the prior backup(2).
But now that we set the lastBackup variables after the delete operation to the next closest snapshot, we would be doing an incremental backup.

boknowswiki:
> @boknowswiki @yasker The test failure happens, because it's expecting a full backup creation since we removed the prior backup(2).
> But now that we set the lastBackup variables after the delete operation to the next closest snapshot, we would be doing an incremental backup.
> 
> It would be okay to change the assert to expect an incremental backup.
> Or to stick with the prior expectation, delete all prior backups (0,1,2)

@yasker and @joshimoo , based on my understand, the incremental will be set if the lastbackup is not nil. So in this case, it should be true. And I think the fix should be change `assert backup3['IsIncremental'] is False` to `assert backup3['IsIncremental'] is True`. Please let me know my understand is not correct.

Thanks,
Bo

joshimoo:
> Please let me know my understand is not correct.

Your understanding is correct and updating the test to be incremental is an appropriate solution.

yasker:
Do we still have test to cover the full backup case? Backup a same snapshot twice should trigger full backup as well.

boknowswiki:
> Do we still have test to cover the full backup case? Backup a same snapshot twice should trigger full backup as well.

I think so, there is `# final full backup after unavailable backup` on line 263

PR #509, opened 2020-07-10
pkg/sync/rpc, pkg/types, pkg/util: add HTTP_PROXY, HTTPS_PROXY, and N…

PhanLe1010:
…O_PROXY field to s3-secret. We use those infomation to setup http_proxy in Longhorn manager and engine

This fix is needed for users who are running a cluster behind a http_proxy and wish to use AWS S3 as the backupstore. First, we add 3 new fields in s3-secret (HTTP_PROXY, HTTPS_PROXY, and NO_PROXY) to get the information about the http_proxy from user. Then we use the information to setup environment variables in Longhorn manager and engine. This will allow the manager and engine to send the outbound traffic to AWS S3 backupstore through the Http Proxy.

longhorn/longhorn#1540


yasker:
I don't think the failure is caused by this commit. https://drone-pr.rancher.io/longhorn/longhorn-engine/689/1/3

Let's wait for the fix and retry later.

PhanLe1010:
Which fix is that?

PR #512, opened 2020-07-10
Backupstore file locks

joshimoo:
longhorn/longhorn#612

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/gorilla/handlers v1.4.2
 	github.com/gorilla/mux v1.7.3
 	github.com/gorilla/websocket v1.4.0 // indirect
-	github.com/longhorn/backupstore v0.0.0-20200717215103-739d80bd783e
+	github.com/longhorn/backupstore v0.0.0-20200819202924-268d5de4d5b2
 	github.com/longhorn/go-iscsi-helper v0.0.0-20200724233216-c35d2e7ee6b6
 	github.com/longhorn/sparse-tools v0.0.0-20191231185723-50ffab4d4329
 	github.com/mattn/go-colorable v0.1.4 // indirect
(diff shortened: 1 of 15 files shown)


joshimoo:
Punted for 1.0.2

joshimoo:
The build failure is caused by an error in the new restore logic?
https://github.com/longhorn/longhorn-engine/blob/4f6eefe90709a219e8d61eaccca263048a3fdd4c/pkg/sync/rpc/server.go#L199

Since we fail the restore operation on purpose to test the delete lock even though the actual restore operation wasn't even started yet, this will still be classified as errorState and not recovered. So after removing the Delete lock the following restore operation should succeed but because the previous restore operation failed (wasn't even started) the code no longer allows for restores?

@shuo-wu Shouldn't this just fallback to full restores?

shuo-wu:
> The build failure is caused by an error in the new restore logic?
> https://github.com/longhorn/longhorn-engine/blob/4f6eefe90709a219e8d61eaccca263048a3fdd4c/pkg/sync/rpc/server.go#L199
> 
> Since we fail the restore operation on purpose to test the delete lock even though the actual restore operation wasn't even started yet, this will still be classified as errorState and not recovered. So after removing the Delete lock the following restore operation should succeed but because the previous restore operation failed (wasn't even started) the code no longer allows for restores?
> 
> @shuo-wu Shouldn't this just fallback to full restores?

All errors after the restore initialization will be blindly recorded in `RestoreStatus`. Some errors won't affect the following restores and some are critical. But it would be complicated to enable Longhorn to distinguish the errors and start a full restore as the fallback. Instead, Longhorn can directly mark the replica as ERROR first then see if the existing data can be reused during the following rebuilding. (The data reusing is not implemented yet.)

As for the failure you encounter in the test, can you check if [function `sync_agent_server_reset `](https://github.com/longhorn/longhorn-engine/blob/4f6eefe90709a219e8d61eaccca263048a3fdd4c/integration/common/cmd.py#L167) works for you? :
https://github.com/longhorn/longhorn-engine/blob/4f6eefe90709a219e8d61eaccca263048a3fdd4c/pkg/sync/rpc/server.go#L257
Or you can create a separate test for each restore related case.

joshimoo:
longhorn/longhorn-engine#529 fixed the valid test failure
there is still a flaky test tracked at longhorn/longhorn#1628

joshimoo:
This is good to go, I kicked of the build again and we can look into the root cause of the flaky test during the bug fixing phase.

PR #513, opened 2020-07-11
Add support for arm64 (draft)

ivang:
Addresses partially https://github.com/longhorn/longhorn/issues/6 by adding support for arm64 (armhf not supported).

Possibly due to some race conditions integration tests were initially failing, which was resolved by introducing sleep at the problematic places. This is clearly a workaround and a cleaner solution would be needed.

With some small changes in longhorn-manager and longhorn-ui, I was able to deploy longhorn on k3s cluster of 4 Raspberry Pi 4s and, as far as I can tell, it seems to be functioning properly.

PRs for longhorn-manager and longhorn-ui are coming up.


ivang:
Closing in favor of https://github.com/longhorn/longhorn-engine/pull/514 where all commits are signed off.

PR #518, opened 2020-07-22
replica: Fix when replica crashed or removed triggers replica rebuild

boknowswiki:
This fix adds the replica to report self detected errors through ping
response. And this will trigger replica rebuild automatically.

https://github.com/longhorn/longhorn/issues/1615

### DIFF
--- a/pkg/replica/replica.go
+++ b/pkg/replica/replica.go
@@ -74,6 +74,7 @@ type Info struct {
 	Head            string
 	Dirty           bool
 	Rebuilding      bool
+	Error           string
 	Parent          string
 	SectorSize      int64
 	BackingFileName string
@@ -1118,5 +1119,13 @@ func (r *Replica) GetRemainSnapshotCounts() int {
 }
 
 func (r *Replica) getDiskSize(disk string) int64 {
-	return util.GetFileActualSize(r.diskPath(disk))
+	ret := util.GetFileActualSize(r.diskPath(disk))
+	if ret == -1 {
+		errMessage := fmt.Sprintf("Fail to getfile %v size", r.diskPath(disk))
+		if r.info.Error == "" {
+			r.info.Error = errMessage
+			logrus.Error(errMessage)
+		}
+	}
+	return ret
 }
--- a/integration/common/core.py
+++ b/integration/common/core.py
@@ -441,6 +441,25 @@ def verify_replica_state(grpc_c, addr, state):
     assert verified
 
 
+def verify_replica_mode(grpc_c, addr, mode):
+    if not addr.startswith("tcp://"):
+        addr = "tcp://" + addr
+
+    verified = False
+    for i in range(RETRY_COUNTS_SHORT):
+        replicas = grpc_c.replica_list()
+        snapList = cmd.snapshot_ls(grpc_c.address)
+        for r in replicas:
+            if r.address == addr and r.mode == mode:
+                verified = True
+                break
+        if verified:
+            break
+
+        time.sleep(RETRY_INTERVAL_SHORT)
+    assert verified
+
+
 def verify_read(dev, offset, data):
     for i in range(10):
         readed = read_dev(dev, offset, len(data))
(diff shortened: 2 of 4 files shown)


boknowswiki:
## Pre-merged Checklist
* [ ] Does the PR include the explanation for the fix or the feature?

* [ ] Is the backend code merged (Manager, Engine, Instance Manager, BackupStore etc)?
The PR is at

* [ ] Is the reproduce steps/test steps documented?

* [ ] ~~Which areas/issues this PR might have potential impacts on?
Area
Issues~~

* ~~[ ] **If the fix introduces the code for backward compatibility** Has an separate issue been filed with the label `release/obsolete-compatibility`?
The compatibility issue is filed at~~

* ~~[ ] **If labeled: area/ui** Has the UI issue filed or ready to be merged?
The UI issue/PR is at~~

* ~~[ ] **if labeled: require/doc** Has the necessary document PR submitted or merged?
The Doc issue/PR is at~~

* [ ] **If labeled: require/automation-e2e** Has the end-to-end test plan been merged? Have QAs agreed on the automation test case?
The automation skeleton PR is at
The automation test case PR is at

* ~~[ ] **if labeled: require/automation-engine** Has the engine integration test been merged?
The engine automation PR is at~~

* ~~[ ] **if labeled: require/manual-test-plan** Has the manual test plan been documented?
The updated manual test plan is at~~

boknowswiki:
> Can we add an integration test for this issue?

Let me take a look and think about it.

pkg/replica/replica.go
```diff
@@ -1118,5 +1119,13 @@ func (r *Replica) GetRemainSnapshotCounts() int {
 }
 
 func (r *Replica) getDiskSize(disk string) int64 {
-	return util.GetFileActualSize(r.diskPath(disk))
+	ret := util.GetFileActualSize(r.diskPath(disk))
+	if ret == -1 {
+		errMessage := fmt.Sprintf("Fail to getfile %v size", r.diskPath(disk))
```
yasker, line 1124: On second thought, why this function doesn't report an error? Should it be more straightforward if this reports the error?

boknowswiki: Since it's called from [ListDisks func](https://github.com/longhorn/longhorn-engine/blob/b33374258c994d9da0808f2d09161c10a65e3de9/pkg/replica/replica.go#L1087) and the return for that is `map[string]DiskInfo` type, I don't find a way to return error there.

boknowswiki:
One question about the test coverage. If we add the integration test, does it mean we don't need to automation-e2e test? Or we need both? Thanks.

pkg/replica/replica.go
yasker: We need to refactor ListDisks() to allow it to return error then.

yasker:
@boknowswiki for this one, I prefer to implement the test in both engine integration test and e2e test. We need the engine integration test to make sure the stability of the engine, but the engine integration won't trigger the rebuild process, which will be covered by the e2e test.

boknowswiki:
> @boknowswiki for this one, I prefer to implement the test in both engine integration test and e2e test. We need the engine integration test to make sure the stability of the engine, but the engine integration won't trigger the rebuild process, which will be covered by the e2e test.

Thanks for the input. I am looking for how to test the rebuild part on integration test. Let me put a test skeleton for both integration and automation e2e. Please help to take a look then.

pkg/replica/replica.go
yasker: In fact, we need to be more careful here.

Is ListDisks() call in the critical path or not? For example, does the failure of the call will result in a data path error or not? That will determine if we need to error out or not. Otherwise, we might have some unintended consequences.

boknowswiki: I think this is from `cmd.LsReplicaCmd`, and it's not in data path. It's ok to return whatever we found, and report the error through the monitor ping. Please let me know if you think we need to refactor ListDisks(), and this will impact the whole call path.

PR #519, opened 2020-07-23
Update instance manager and add a test for process creation failure case

shuo-wu:
longhorn/longhorn#1599

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -101,7 +101,7 @@ RUN cd integration && tox --notest
 RUN cd /go/src/github.com/longhorn && \
     git clone https://github.com/longhorn/longhorn-instance-manager.git && \
     cd longhorn-instance-manager && \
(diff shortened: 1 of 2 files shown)


yasker:
The dependency wasn't up to date so it's better to put it as a draft PR or mentioned `don't merge`.

https://github.com/longhorn/longhorn-instance-manager/pull/79 has been merged.

Dockerfile.dapper
```diff
@@ -99,9 +99,9 @@ RUN cd integration && tox --notest
 
 # Build longhorn-instance-manager
 RUN cd /go/src/github.com/longhorn && \
-    git clone https://github.com/longhorn/longhorn-instance-manager.git && \
+    git clone https://github.com/shuo-wu/longhorn-instance-manager.git && \
```
joshimoo, line 102: currently pointing to your fork

PR #521, opened 2020-07-28
Restore refactor

shuo-wu:
longhorn/longhorn#1279


shuo-wu:
The ci failure may be caused by https://github.com/longhorn/longhorn/issues/1628

yasker:
@shuo-wu Your guess is right. Can you check what was wrong?

shuo-wu:
> @shuo-wu Your guess is right. Drone now has the full log. Can you check what was wrong?

The cause of the test failure is already presented in the issue. In brief, the failure is related to a bug/race condition in the test implementation rather than a bug of the PR

PR #529, opened 2020-08-26
Avoid recording the pre-restore errors in syncagent server restore info

shuo-wu:
After the discussion with @joshimoo, we need to avoid recording the errors before performing the actual restore. Otherwise, the related replica will be marked as ERROR once there is a restore file locking timeout, which is bad for the user experience.

### DIFF
--- a/pkg/sync/rpc/server.go
+++ b/pkg/sync/rpc/server.go
@@ -177,7 +177,7 @@ func (s *SyncAgentServer) IsRestoring() bool {
 	return s.isRestoring
 }
 
-func (s *SyncAgentServer) PrepareRestore(backupURL, requestedBackupName, snapshotDiskName string) (err error) {
+func (s *SyncAgentServer) StartRestore(backupURL, requestedBackupName, snapshotDiskName string) (err error) {
 	s.Lock()
 	defer s.Unlock()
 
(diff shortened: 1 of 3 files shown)


pkg/sync/rpc/server.go
```diff
@@ -202,26 +202,53 @@ func (s *SyncAgentServer) PrepareRestore(backupURL, requestedBackupName, snapsho
...
+			validLastRestoredBackup = false
+		}
 
-	// Fall back to full restore.
-	if restoreStatus.LastRestored == "" {
-		logrus.Warnf("There is a restore record in the server but last restored backup is empty with restore state is %v, will do full restore instead", restoreStatus.State)
-		s.RestoreInfo.StartNewRestore(backupURL, requestedBackupName, replica.GenerateSnapTempFileName(snapshotDiskName), snapshotDiskName, false)
-		return nil
+		s.RestoreInfo.StartNewRestore(backupURL, requestedBackupName, toFileName, snapshotDiskName, validLastRestoredBackup)
 	}
```
joshimoo, line 226: How about adding a private util function for the condition checking? 
This makes the code easier to follow without losing any of the details.

example below:
```
isIncrementalRestore := canDoIncrementalRestore(x,y,z)
var toFilename
if isIncrementalRestore {
    toFileName = replica.GenerateDeltaFileName(restoreStatus.LastRestored)
 } else {
     toFileName = replica.GenerateSnapTempFileName(snapshotDiskName)
 }

s.RestoreInfo.StartNewRestore(backupURL, requestedBackupName, toFileName, snapshotDiskName, isIncrementalRestore)
```

joshimoo (changes requested):
Logic looks good to me, left some suggestions.

joshimoo (approved):
LGTM thanks for picking up the suggestions 👍

joshimoo:
This change was to differentiate between restore initiation failures and actual failures during the restore.
related to backup locks longhorn/longhorn#612

PR #531, opened 2020-08-31
Remove duplicated check for opened snapshot/volume

c3y1huang:
Since the similar logic already implemented in `rb.assertOpen` above, this seems safe to remove.


pkg/replica/backup.go
```diff
@@ -237,9 +237,6 @@ func (rb *BackupStatus) ReadSnapshot(snapID, volumeID string, start int64, data
 	if err := rb.assertOpen(id, volumeID); err != nil {
 		return err
 	}
-	if rb.SnapshotID != id && rb.volumeID != volumeID {
```
yasker, line 240: Good catch. And this is a bug too, since it should be `or` instead of `and`.

PR #538, opened 2020-09-04
Enhance error message

c3y1huang:
~Fix~ Enhance message.

### DIFF
--- a/pkg/sync/sync.go
+++ b/pkg/sync/sync.go
@@ -441,7 +441,7 @@ func (t *Task) checkAndExpandReplica(address string, size int64) error {
 		return err
 	}
 	if replicaSize > size {
-		return fmt.Errorf("cannot add a larger replica to the engine")
+		return fmt.Errorf("cannot add new replica larger than size %v", size)
 	} else if replicaSize < size {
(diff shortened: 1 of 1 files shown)


pkg/sync/sync.go
```diff
@@ -441,7 +441,7 @@ func (t *Task) checkAndExpandReplica(address string, size int64) error {
 		return err
 	}
 	if replicaSize > size {
-		return fmt.Errorf("cannot add a larger replica to the engine")
+		return fmt.Errorf("cannot create new replica less than size %v", replicaSize)
```
yasker, line 444: This doesn't look right. ReplicaSize is the new replica which going to be added. @shuo-wu can you confirm?

shuo-wu: You are right. `replicaSize ` is the size of the new/rebuilding replica. `size` is the size of the engine & existing replicas.

c3y1huang: Thanks for clarifying @yasker and @shuo-wu . I've updated the issue to be an enhancement of the error message instead.

yasker (approved):

PR #539, opened 2020-09-07
Fix CLI usage messages

c3y1huang:
Fix CLI usage messages.

### DIFF
--- a/app/cmd/replica.go
+++ b/app/cmd/replica.go
@@ -44,11 +44,11 @@ func ReplicaCmd() cli.Command {
 			},
 			cli.StringFlag{
 				Name:  "restore-from",
-				Usage: "specify backup to be restored, must be used with --restore-name",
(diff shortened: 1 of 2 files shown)


app/cmd/replica.go
```diff
@@ -44,11 +44,11 @@ func ReplicaCmd() cli.Command {
 			},
 			cli.StringFlag{
 				Name:  "restore-from",
-				Usage: "specify backup to be restored, must be used with --restore-name",
+				Usage: "specify backup to be restored, must use with --restore-name",
```
yasker, line 47: since there is no subject in the sentence here, it's better to keep `A must be used with B`.

yasker (changes requested):

yasker:
Hmm? @c3y1huang Why is this issue closed? We still need to update the `restore` usage part.

c3y1huang:
oops, too quick on hand. I will open another.

PR #542, opened 2020-09-15
508 replica revision counter engine

boknowswiki:
https://github.com/longhorn/longhorn/issues/508


pkg/controller/control.go
```diff
@@ -539,27 +542,82 @@ func (c *Controller) Start(addresses ...string) error {
...
+					// can be good replica.
+					if t.Add(time.Second * 5).After(lastTime) {
+						if replicaInfo.HeadFileSize >= largestSize {
+							bestCandidate = r
+						}
+					}
+				}
+
+				if bestCandidate == (types.Replica{}) {
+					logrus.Errorf("BUG: Should find one candidate for salvage")
```
shuo-wu, line 583: If there is no replica for the salvage, we may need to return an error here, right? Then longhorn manager shouldn't retry the auto salvage.

Similar below.

shuo-wu (changes requested):

app/cmd/replica.go
```diff
@@ -74,7 +80,12 @@ func startReplica(c *cli.Context) error {
 		return err
 	}
 
-	s := replica.NewServer(dir, backingFile, 512)
+	disableRevCounter, err := strconv.ParseBool(c.String("disableRevCounter"))
```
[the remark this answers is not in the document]
boknowswiki: I was using bool type, but hit some issue. maybe i wasn't use it correctly, let me try it again. Also with the controller command.

pkg/replica/revision_counter.go
```diff
@@ -91,6 +91,11 @@ func (r *Replica) initRevisionCounter() error {
 }
 
 func (r *Replica) GetRevisionCounter() int64 {
+	// Return 0 since revision counter has been diabled.
```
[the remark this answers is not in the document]
boknowswiki: I think this has the minimum changes. Could you explain more about your idea? Thanks.

pkg/controller/control.go
```diff
@@ -539,27 +542,82 @@ func (c *Controller) Start(addresses ...string) error {
...
+					return err
+				}
+
+				rep, err := repClient.GetReplica()
+				if err != nil {
+					return err
+				}
+				replicaCandidates[r] = rep
+				if lastModifyTime == 0 ||
+					rep.LastModifyTime < lastModifyTime {
```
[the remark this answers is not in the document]
boknowswiki: Yes, good catch. Will fix it. Thanks.

pkg/controller/control.go
boknowswiki: Sure, will fix them.

pkg/replica/revision_counter.go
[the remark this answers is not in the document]
shuo-wu: The purpose of thess functions is to CRUD the revision counter. We can consider it as a kind of util function. And it's the caller's responsibility (rather than this CRUD functions) to determine if it needs to use revision counter.
But in general, this is not restrictly required. It's more like a possible suggestion or a discussion. And it won't affect the functionality of this feature.

pkg/controller/control.go
```diff
@@ -539,27 +543,83 @@ func (c *Controller) Start(addresses ...string) error {
...
+				if lastModifyTime == 0 ||
+					rep.LastModifyTime > lastModifyTime {
+					lastModifyTime = rep.LastModifyTime
+				}
 			}
-			counter, err := c.backend.GetRevisionCounter(r.Address)
-			if err != nil {
-				return err
+
+			if len(replicaCandidates) != 0 {
```
shuo-wu, line 568: We can avoid redundant `else`:
```
if len(replicaCandidates) == 0 {
    return fmt.Errorf("BUG: Should find one candidate for salvage")
}

// len(replicaCandidates) != 0 
......
```

shuo-wu (changes requested):

app/cmd/replica.go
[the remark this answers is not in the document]
shuo-wu: It seems that you have changed to bool flags. 
If the new manager works with the volume using the old engine image, then that's fine.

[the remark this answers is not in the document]
boknowswiki: Yes, I am able to use the bool flags.

pkg/replica/revision_counter.go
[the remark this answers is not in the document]
boknowswiki: I understand your idea. For example, 'GetRevisionCounter' is exposed as interface, and the caller doesn't know if the revision counter is disabled or enabled. Also these CRUD are associated with the individual replicas. I think in this case, handle it in the CRUD function is better. And the caller doesn't need to worry about this either.

pkg/controller/control.go
boknowswiki: I see, sure. Will fix it. Thanks.

pkg/replica/revision_counter.go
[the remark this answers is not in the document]
boknowswiki: After discuss with @yasker, I will revert this part of change and change the caller side to not call CRUD function if the revision counter is disabled. Thanks.

shuo-wu:
In general LGTM except for the test failure.

boknowswiki:
@shuo-wu and @yasker , i hit a backup store integration test issue, first time it shows failed at `test_backup_cli` , but I have tried several time on my local and i couldn't reproduce it. then @meldafrawi helped me to restart the build, and it passed this time.

So far I don't see any my change related to this test.

boknowswiki:
create [a issue](https://github.com/longhorn/longhorn/issues/1816) to track the integration issue.
@shuo-wu  and @yasker , please help to review the PR and let me know any comments? Thanks.

pkg/backend/file/file.go
```diff
@@ -67,6 +67,29 @@ func (f *Wrapper) Size() (int64, error) {
 	return stat.Size(), nil
 }
 
+func (f *Wrapper) GetRevisionCounterDisabled() (bool, error) {
```
yasker, line 70: `IsRevisionCounterDisabled` or `IsRevisionCounterEnabled`.

pkg/types/types.go
```diff
@@ -93,6 +96,11 @@ type Replica struct {
 	Mode    Mode
 }
 
+type SalvageReplica struct {
```
yasker, line 99: How about `ReplicaSalvageInfo`?

yasker (changes requested):

pkg/backend/file/file.go
boknowswiki: Hi Sheng,
At beginning I was named it exactly the same as you suggested, but after I think about it may return error, not only a boolean, I changed it to get. Please let me know what you think. I just feel a little wired if we named it 'IsRevisionCounterDisabled', and it also may return an error. Thanks.

pkg/types/types.go
boknowswiki: Sure, I can update this, thanks.

pkg/backend/file/file.go
yasker: It's fine for `IsRevisionCounterDisabled` to return an error. `Is` indicates it's a boolean, which is better suited here.

shuo-wu (dismissed):

boknowswiki:
@shuo-wu , I have updated the engine part based on our discussion, please take a look. Thanks.

pkg/controller/control.go
```diff
@@ -539,27 +565,98 @@ func (c *Controller) Start(addresses ...string) error {
...
-		for _, r := range c.replicas {
-			// The related backend is nil if the mode is ERR
-			if r.Mode == types.ERR {
-				continue
-			}
-			counter, err := c.backend.GetRevisionCounter(r.Address)
-			if err != nil {
-				return err
+		err := c.checkReplicaRevCounterSettingMatch()
+		if err != nil {
```
shuo-wu, line 569: `if err := c.checkReplicaRevCounterSettingMatch(); err != nil {`

pkg/controller/control.go
```diff
@@ -539,27 +565,98 @@ func (c *Controller) Start(addresses ...string) error {
...
+			return err
+		}
+
+		if c.revisionCounterDisabled {
+			// Find best replica for salvge recovering, based on
+			// lastModifyTime and HeadFileSize.
+			if c.salvageRequested {
+				replicaCandidates := make(map[types.Replica]types.ReplicaSalvageInfo)
```
shuo-wu, line 577: Can we abstract this part as a function, e.g.,`salvageRevisionCounterDisabledReplicas()`
Similar below.

shuo-wu (changes requested):
Others LGTM

pkg/controller/control.go
boknowswiki: Sure, will update. Thanks.

pkg/controller/control.go
boknowswiki: Sure, will update. Thanks.

shuo-wu (approved):

yasker (approved):

PR #544, opened 2020-10-01
Add engine integration test for engine revision counter mismatch case

boknowswiki:
Test case for engine and replica revision counter setting mismatch.

Case 1: engine has revision counter enabled, one replica is revision
counter enabled, the other one is disabled.

Case 2: engine has revision counter disabled, one replica is revision
counter disabled, the other one is enabled.

https://github.com/longhorn/longhorn/issues/1840

### DIFF
--- a/integration/common/core.py
+++ b/integration/common/core.py
@@ -89,11 +89,14 @@ def wait_for_process_error(client, name):
 def create_replica_process(client, name, replica_dir="",
                            args=[], binary=LONGHORN_BINARY,
                            size=SIZE, port_count=15,
-                           port_args=["--listen,localhost:"]):
+                           port_args=["--listen,localhost:"],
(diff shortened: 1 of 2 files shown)


integration/core/test_controller.py
```diff
@@ -2,18 +2,28 @@
 import os
 import grpc
 import pytest
+import tempfile
```
yasker, line 5: Can we move this to the `integration/instance`? Core test doesn't mean to create/delete processes.

yasker (changes requested):
Need to move the test to `integration/instance`, otherwise LGTM.

integration/core/test_controller.py
boknowswiki: Sure, I just updated the PR.

yasker (approved):

PR #547, opened 2020-10-14
ARM64 support for longhorn engine

boknowswiki:
https://github.com/longhorn/longhorn/issues/6

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -8,8 +8,10 @@ platform:
 
 steps:
 - name: fetch
-  image: alpine/git
+  image: ubuntu:18.04
   commands:
+  - apt-get update
+  - apt-get install -y git
   - git fetch --tags
 
 - name: build
@@ -92,9 +94,10 @@ platform:
 
 steps:
 - name: fetch
-  image: arm64v8/alpine
+  image: ubuntu:18.04
   commands:
-  - apk add git
+  - apt-get update
+  - apt-get install -y git
   - git fetch --tags
 
 - name: build
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,6 +1,6 @@
 ---
 kind: pipeline
-name: default
+name: default-amd64
 
 platform:
   os: linux
@@ -14,7 +14,7 @@ steps:
 
 - name: build
   pull: default
-  image: rancher/dapper:1.10.3
+  image: rancher/dapper:v0.5.3
   commands:
   - dapper ci
   privileged: true
@@ -25,12 +25,14 @@ steps:
 - name: publish-master-image
   image: plugins/docker
   settings:
+    build_args:
+    - ARCH=amd64
     custom_dns: 1.1.1.1
     dockerfile: package/Dockerfile
     password:
       from_secret: docker_password
     repo: longhornio/longhorn-engine
-    tag: master
+    tag: "master-amd64"
     username:
       from_secret: docker_username
   when:
@@ -42,12 +44,14 @@ steps:
 - name: publish-tagged-image
   image: plugins/docker
   settings:
+    build_args:
+    - ARCH=amd64
     custom_dns: 1.1.1.1
     dockerfile: package/Dockerfile
     password:
       from_secret: docker_password
     repo: longhornio/longhorn-engine
-    tag: "${DRONE_TAG}"
+    tag: "${DRONE_TAG}-amd64"
     username:
       from_secret: docker_username
   when:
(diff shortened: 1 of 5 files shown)


.drone.yml
```diff
@@ -14,7 +14,7 @@ steps:
 
 - name: build
   pull: default
-  image: rancher/dapper:1.10.3
+  image: rancher/dapper:v0.5.3
```
joshimoo, line 17: Is this a downgrade or did the versioning scheme change?

boknowswiki: @meldafrawi , please help to confirm the version. thanks.

eoprede: Just a note  from somebody who is following (somewhat impatiently) the process of getting arm support for longhorn - according to https://hub.docker.com/r/rancher/dapper/tags?page=2 dapper 1.10.3 is  about 5 years old.

meldafrawi: @joshimoo 
It's not a downgrade, like @eoprede mentioned dapper 1.10 is old  and don't support arm.

yasker:
Seems the ARM64 test failed due to a MinIO issue?

.drone.yml
joshimoo: No problem, looks like the dapper versioning scheme was changed.
@boknowswiki , @meldafrawi We should make sure to update to the the same version for all of our components.

Dockerfile.dapper
```diff
@@ -26,20 +26,24 @@ RUN apt-get update && \
 RUN rm -f /bin/sh && ln -s /bin/bash /bin/sh
 
 # Install Go & tools
-ENV GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm=armv6l GOLANG_ARCH=GOLANG_ARCH_${ARCH} \
+ENV GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm64=arm64 GOLANG_ARCH=GOLANG_ARCH_${ARCH} \
     GOPATH=/go PATH=/go/bin:/usr/local/go/bin:${PATH} SHELL=/bin/bash
-RUN wget -O - https://storage.googleapis.com/golang/go1.13.3.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local && \
+RUN wget -O - https://storage.googleapis.com/golang/go1.14.4.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local && \
```
joshimoo, line 31: I don't mind updating to newer versions go, dapper, etc, but we should try to point out these changes in separate commits, same as if we update vendor dependencies. Since these are still implicit dependencies for our code.

Dockerfile.dapper
```diff
@@ -26,20 +26,24 @@ RUN apt-get update && \
...
+	DOCKER_URL_arm64=https://download.docker.com/linux/ubuntu/dists/bionic/pool/stable/arm64/docker-ce_18.06.3~ce~3-0~ubuntu_arm64.deb \
     DOCKER_URL=DOCKER_URL_${ARCH}
 
 RUN wget ${!DOCKER_URL} -O docker_ce_${ARCH} && dpkg -i docker_ce_${ARCH}
 
 # Minio
-RUN wget -O /usr/bin/minio https://dl.minio.io/server/minio/release/linux-amd64/archive/minio.RELEASE.2018-05-25T19-49-13Z && chmod +x /usr/bin/minio
+ENV MINIO_URL_amd64=https://dl.minio.io/server/minio/release/linux-amd64/archive/minio.RELEASE.2018-05-25T19-49-13Z \
+    MINIO_URL_arm64=https://dl.minio.io/server/minio/release/linux-arm64/archive/minio.RELEASE.2020-06-22T03-12-50Z \
+		MINIO_URL=MINIO_URL_${ARCH}
```
joshimoo, line 44: We should use the same minio version for all archictures.

Dockerfile.dapper
```diff
@@ -74,7 +86,11 @@ RUN cd /go/src/github.com/ && \
 RUN pip3 install grpcio==1.22.0 grpcio_tools==1.22.0
 
 # buf
-RUN wget -O /usr/bin/buf https://github.com/bufbuild/buf/releases/download/v0.16.0/buf-Linux-x86_64 && chmod +x /usr/bin/buf
+ENV GO111MODULE=on
+RUN go get \
+  github.com/bufbuild/buf/cmd/buf \
+  github.com/bufbuild/buf/cmd/protoc-gen-buf-check-breaking \
+  github.com/bufbuild/buf/cmd/protoc-gen-buf-check-lint
```
joshimoo, line 93: since these are external dependencies do we want to tag these to a specific version?

scripts/package
```diff
@@ -11,14 +11,26 @@ TAG=${TAG:-${VERSION}}
...
+		ARCH=amd64
+		;;
+	*)
+		echo "$(uname -a): unsupported architecture"
+		exit 1
+esac
+
 if [ ! -x ./bin/longhorn ]; then
     ./scripts/build
 fi
```
joshimoo, line 28: Just asking for clarification, it looks like we are always expecting to be building on the distribution os (amd build on an amd, arm build on an arm instance) ?

joshimoo (changes requested):
LGTM in general, non of my feedback are hard blockers but I do think we want to at least use the same version of dependencies/tools across different archs.

Also unify the formatting the new additions are tabs while the old ones are spaces.

Dockerfile.dapper
boknowswiki: Good catch, I will fix it. Thanks.

Dockerfile.dapper
boknowswiki: Yes, good point. Let me try if we can keep use 13.3 golang. Thanks.

scripts/package
boknowswiki: Yes, I think currently this is the way.

Dockerfile.dapper
boknowswiki: Good point. I will fix it. Thanks.

.drone.yml
```diff
@@ -77,3 +81,141 @@ volumes:
...
+kind: pipeline
+name: default-arm64
+
+platform:
+  os: linux
+  arch: arm64
+
+steps:
+- name: fetch
+  image: arm64v8/alpine
```
joshimoo, line 95: Can we use a different distro, `alpine` uses musl as it's standard libc by default.
Unless most arm oses use musl?

how abou? FROM arm=arm64/ubuntu:18.04

The reason I say mention this is that our go builds aren't completely static, the dns resolver relies on the stdlib at the moment. I will fix that when working on https://github.com/longhorn/longhorn/issues/1768

joshimoo (approved):
LGTM, left a comment about alpine.
Feel free to merge this after addressing that issue.

.drone.yml
boknowswiki: Sorry I don't quite understand your concern here, musl is a library and in arm64 case, it will be compiled for arm64 architecture. I don't think we have to use ubuntu because of this.

Also what do you mean our go builds are not completely static?

joshimoo (approved):
LGTM

PR #553, opened 2020-12-15
scripts: Disable cache for image build

shuo-wu:
longhorn/longhorn#2091


yasker:
Close in favor of https://github.com/longhorn/longhorn-engine/pull/555

PR #556, opened 2021-01-07
pkg/sync: return error when fail to start snapshot purge on one of the replicas

PhanLe1010:
There are 2 callers of the SnapshotPurge command:
1. From Longhorn manager API
2. After finish rebuilding a new replica and `Automatically Cleanup System Generated Snapshot` setting is on

This PR make sure that the error when start SnapshotPurge is reported back to Longhorn manager. It makes debugging easier.

longhorn/longhorn#1895

### DIFF
--- a/pkg/sync/sync.go
+++ b/pkg/sync/sync.go
@@ -124,21 +124,26 @@ func (t *Task) PurgeSnapshots(skip bool) error {
 		return err
 	}
 
+	taskErr := NewTaskError()
 	for _, r := range replicas {
 		if ok, err := t.isRebuilding(r); err != nil {
-			return err
+			taskErr.Append(NewReplicaError(r.Address, err))
 		} else if ok {
-			return fmt.Errorf("cannot purge snapshots because %s is rebuilding", r.Address)
+			taskErr.Append(NewReplicaError(r.Address, fmt.Errorf("cannot purge snapshots because %s is rebuilding", r.Address)))
 		}
 
 		if ok, err := t.isPurging(r); err != nil {
-			return err
+			taskErr.Append(NewReplicaError(r.Address, err))
 		} else if ok {
 			if skip {
 				return nil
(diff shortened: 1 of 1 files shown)


joshimoo (commented):
In principle LGTM, we would fail if we fail to create the replica client in line 152.
What are the failure cases for that and are we expected to mark the replica as failure as a result for an intermittent connection failure? cc @shuo-wu 

https://github.com/longhorn/longhorn-engine/pull/556/files#diff-2e7c810166140fd7a58d6bdb610fef7e6de8f5f9ad2002659392878ecafd6befR152

shuo-wu (changes requested):
Can we group all replica errors here rather than returning an error immediately when the 1st error is found? e.g., https://github.com/longhorn/longhorn-engine/blob/b9e072eb19cfcc27894e10a8e60794c66ca918f8/pkg/sync/backup.go#L164

shuo-wu:
> In principle LGTM, we would fail if we fail to create the replica client in line 152.
> What are the failure cases for that and are we expected to mark the replica as failure as a result for an intermittent connection failure? cc @shuo-wu
> 
> https://github.com/longhorn/longhorn-engine/pull/556/files#diff-2e7c810166140fd7a58d6bdb610fef7e6de8f5f9ad2002659392878ecafd6befR152

Yes. IIRC, the engine process will mark the replica as mode `ERROR` if the connection is lost for a while.

shuo-wu:
BTW, there is one more caller of the purge command: the restore function https://github.com/longhorn/longhorn-engine/blob/b9e072eb19cfcc27894e10a8e60794c66ca918f8/pkg/sync/backup.go#L208

PhanLe1010:
Thanks @joshimoo and @shuo-wu . I grouped the errors together.

shuo-wu (approved):

PR #557, opened 2021-01-15
Enable Backing image

shuo-wu:
longhorn/longhorn#2006


app/cmd/backup.go
```diff
@@ -42,6 +42,14 @@ func BackupCreateCmd() cli.Command {
 				Name:  "dest",
 				Usage: "destination of backup if driver supports, would be url like s3://bucket@region/path/ or vfs:///path/",
 			},
+			cli.StringFlag{
+				Name:  "backing-image-name",
```
yasker, line 46: Need to understand why we have both `name` and `URL`

pkg/replica/server.go
```diff
@@ -81,11 +73,10 @@ func (s *Server) Open() error {
 	}
 
 	_, info := s.Status()
-	size := s.getSize(info.Size)
 	sectorSize := s.getSectorSize()
 
-	logrus.Infof("Opening volume %s, size %d/%d", s.dir, size, sectorSize)
-	r, err := New(size, sectorSize, s.dir, s.backing, s.revisionCounterDisabled)
+	logrus.Infof("Opening volume %s, size %d/%d", s.dir, info.Size, sectorSize)
```
yasker, line 78: OK, what would happen if the actual volume size is bigger than the backing file size?

shuo-wu: In that case, it's similar to the volume expansion: the backing file is like an old snapshot with a small size.

app/cmd/backup.go
shuo-wu: `Name` will be used if the user doesn't specify the backing image when restoring this backup.
`URL` is designed to inform users which backing image the backup volume was using. Then they can check if they need to create a new backing image object before the restore.

shuo-wu:
I didn't find a better way to detect the file format/type. Currently executing `file` or `qemu-img info` is the simplest way.

shuo-wu:
The integration test failure is caused by this bug: https://github.com/longhorn/longhorn/issues/1628. Will re-run the test.

pkg/backing/backing.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+	default:
+		return nil, fmt.Errorf("format %v of the backing file %v is not supported", format, file)
+	}
+
+	size, err := f.Size()
+	if err != nil {
+		return nil, err
+	}
+	if size%defaultSectorSize != 0 {
```
joshimoo, line 99: Is the sparse-tools direct io issue/changes relevant, do we still need this check?
https://github.com/longhorn/longhorn/issues/1943
https://github.com/longhorn/sparse-tools/pull/69/files

joshimoo (approved):
LGTM, I left some questions.

pkg/qcow/libqcow.go
```diff
@@ -1,8 +1,6 @@
-// +build qcow
-
 package qcow
 
-// #cgo LDFLAGS: -lqcow -lz
+// #cgo LDFLAGS: -lqcow -lz -pthread
```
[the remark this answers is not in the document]
shuo-wu: It seems so. Without this parameter, `make` will fail and the error like `cgo undefined reference to pthread_rwlock_init` will be triggered

pkg/backing/backing.go
shuo-wu: Yes. This `size` is the backing image size, which can be different from the replica size.

pkg/replica/replica.go
```diff
@@ -167,6 +163,13 @@ func construct(readonly bool, size, sectorSize int64, dir, head string, backingF
 		return nil, err
 	}
 
+	// The backing file path can be changed, need to update it after loading
+	// the meta data.
+	r.info.BackingFile = backingFile
+	if backingFile != nil {
+		r.info.BackingFilePath = backingFile.Path
+	}
```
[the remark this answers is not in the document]
shuo-wu: `info.BackingFile` is part of the snapshot file chain and used to help the backing file read inside the replica server. Whereas `info.BackingFile.Path` can [be read in the sync agent server](https://github.com/longhorn/longhorn-engine/blob/1481ab6dd315557d10c7e51081b966b39d694ed0/pkg/backup/main.go#L90) during the backup creation, so that the sync agent server doesn't need to find a way to init the struct `*BackingFile`

PR #560, opened 2021-02-04
add command line argument directio (#1943)

cclhsu:
#### Proposed Changes ####

add command line argument directio to resolve ticket '[BUG] sparse-tools: ssync should explicitly asking for directIO or not, rather than determine is using filesize' (#1943)

1. Update to use `directIO` from the https://github.com/longhorn/sparse-tools/pull/69
1. Update to use `directIO` from the https://github.com/longhorn/sparse-tools/pull/70

#### Types of Changes ####

Bug fixing.

#### Verification ####

Updated/Added unit test cases, and integration test should pass

Build log should have log like following:

```
> grep -rnis directio ./logs_longhorn_longhorn-engine_1_3.log 
81:time="2021-03-22T07:56:54Z" level=info msg="source file size: 178, setting up directIO: false"
82:[test-0faf0f23-replica-1-2] time="2021-03-22T07:56:54Z" level=info msg="open: receiving fileSize: 178, setting up directIO: false"
87:time="2021-03-22T07:56:54Z" level=info msg="source file size: 4194304, setting up directIO: true"
92:time="2021-03-22T07:56:54Z" level=info msg="open: receiving fileSize: 4194304, setting up directIO: true"
96:time="2021-03-22T07:56:54Z" level=info msg="source file size: 126, setting up directIO: false"
101:time="2021-03-22T07:56:54Z" level=info msg="open: receiving fileSize: 126, setting up directIO: false"
793:time="2021-03-22T07:57:27Z" level=info msg="source file size: 178, setting up directIO: false"
794:[test-0faf0f23-replica-1-2] time="2021-03-22T07:57:27Z" level=info msg="open: receiving fileSize: 178, setting up directIO: false"
799:time="2021-03-22T07:57:27Z" level=info msg="source file size: 4194304, setting up directIO: true"
802:time="2021-03-22T07:57:27Z" level=info msg="open: receiving fileSize: 4194304, setting up directIO: true"
```

#### Linked Issues ####

Refs: # https://github.com/longhorn/longhorn/issues/1943

#### Further Comments ####

Signed-off-by: Clark Hsu <clark.hsu@suse.com>

### DIFF
--- a/pkg/sync/rpc/server.go
+++ b/pkg/sync/rpc/server.go
@@ -5,6 +5,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"sync"
@@ -329,8 +330,12 @@ func (*SyncAgentServer) FileRename(ctx context.Context, req *ptypes.FileRenameRe
 
 func (s *SyncAgentServer) FileSend(ctx context.Context, req *ptypes.FileSendRequest) (*empty.Empty, error) {
 	address := net.JoinHostPort(req.Host, strconv.Itoa(int(req.Port)))
+	directIO := true
+	if filepath.Ext(strings.TrimSpace(req.FromFileName)) == ".meta" {
+		directIO = false
+	}
 	logrus.Infof("Sending file %v to %v", req.FromFileName, address)
-	if err := sparse.SyncFile(req.FromFileName, address, FileSyncTimeout); err != nil {
+	if err := sparse.SyncFile(req.FromFileName, address, FileSyncTimeout, directIO); err != nil {
 		return nil, err
 	}
 	logrus.Infof("Done sending file %v to %v", req.FromFileName, address)
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/gorilla/websocket v1.4.0 // indirect
 	github.com/longhorn/backupstore v0.0.0-20210315131747-51dcc71901ed
 	github.com/longhorn/go-iscsi-helper v0.0.0-20210304120829-4c21e728023b
-	github.com/longhorn/sparse-tools v0.0.0-20201111045115-7a3bbbb6f408
+	github.com/longhorn/sparse-tools v0.0.0-20210322065822-14ef01cc10e9
 	github.com/mattn/go-colorable v0.1.4 // indirect
 	github.com/mattn/go-runewidth v0.0.5-0.20181218000649-703b5e6b11ae // indirect
 	github.com/moby/moby v1.11.1
--- a/go.sum
+++ b/go.sum
@@ -42,8 +42,8 @@ github.com/longhorn/go-iscsi-helper v0.0.0-20210304120829-4c21e728023b h1:RJukKq
 github.com/longhorn/go-iscsi-helper v0.0.0-20210304120829-4c21e728023b/go.mod h1:9z/y9glKmWEdV50tjlUPxFwi1goQfIrrsoZbnMyIZbY=
 github.com/longhorn/nsfilelock v0.0.0-20200723175406-fa7c83ad0003 h1:Jw9uANsGcHTxp6HcC++/vN17LfeuDmozHI2j6DoZf5E=
 github.com/longhorn/nsfilelock v0.0.0-20200723175406-fa7c83ad0003/go.mod h1:0CLeXlf59Lg6C0kjLSDf47ft73Dh37CwymYRKWwAn04=
-github.com/longhorn/sparse-tools v0.0.0-20201111045115-7a3bbbb6f408 h1:buF1O/6By6sR6lZODDPzl12svm/Yq63FI81hKw9piTk=
-github.com/longhorn/sparse-tools v0.0.0-20201111045115-7a3bbbb6f408/go.mod h1:1WSSJRDNret4VKADAzvD49oaVt1/idYVWBL8TiyevRk=
+github.com/longhorn/sparse-tools v0.0.0-20210322065822-14ef01cc10e9 h1:xz82ZNJhls2oZ8/Dwctgz2AQbAVMnG1LAYGymu1SX4A=
+github.com/longhorn/sparse-tools v0.0.0-20210322065822-14ef01cc10e9/go.mod h1:1WSSJRDNret4VKADAzvD49oaVt1/idYVWBL8TiyevRk=
 github.com/mattn/go-colorable v0.1.4 h1:snbPLB8fVfU9iwbbo30TPtbLRzwWu6aJS6Xh4eaaviA=
 github.com/mattn/go-colorable v0.1.4/go.mod h1:U0ppj6V5qS13XJ6of8GYAs25YV2eR4EVcfRqFIhoBtE=
 github.com/mattn/go-isatty v0.0.8 h1:HLtExJ+uU2HOZ+wI0Tt5DtUDrx8yhUqDcp7fYERX4CE=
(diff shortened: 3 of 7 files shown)


shuo-wu:
Remember to update the vendor/dependency.

cclhsu:
@shuo-wu This PR ready for review.  Can you take a look? Thanks!

pkg/sync/rpc/server.go
```diff
@@ -329,8 +329,9 @@ func (*SyncAgentServer) FileRename(ctx context.Context, req *ptypes.FileRenameRe
 
 func (s *SyncAgentServer) FileSend(ctx context.Context, req *ptypes.FileSendRequest) (*empty.Empty, error) {
 	address := net.JoinHostPort(req.Host, strconv.Itoa(int(req.Port)))
+	directIO := false
```
shuo-wu, line 332: By default, it should be true

shuo-wu (changes requested):

pkg/sync/rpc/server.go
cclhsu: When it set as true, it will cause 12 integration tests to fail.  All have failure message similar as below:

```
time="2021-02-24T08:33:54Z" level=fatal msg="Error running add replica command: failed to prepare rebuilding replica tcp://localhost:10015 for volume localhost:30001: rpc error: code = Unknown desc = failed to send file volume-head-002.img.meta to localhost:10018: rpc error: code = Unknown desc = syncDataInterval [       0:       1](  1) failed, err: failed to get checksums for client and server, server checksum error: <nil>, client checksum error: File to read interval:[       0:       1](  1), error: read volume-head-002.img.meta: invalid argument"
```

shuo-wu: That's unexpected... We may need to modify the related tests and make sure the replica size is the multiple of 512Byte

cclhsu: May I know the context of using 512byte? Seems like some place using 512byte and other using 4k.

```
Blocks int64 = 4 << 10 // 4k => github.com/longhorn/longhorn-engine/vendor/github.com/longhorn/sparse-tools/sparse/layout.go
const defaultSectorSize = 512 // => github.com/longhorn/longhorn-engine/pkg/backing/backing.go
defaultSectorSize = 4096 // => github.com/longhorn/longhorn-engine/pkg/replica/replica.go
```

yasker: 512 is the sector size of qcow2 files, which we might use for backing images. In general, we want to be 4k aligned for the size. We need the alignment for the directIO read/write.

In the error message above, it's about `.img.meta` which is a metadata file, so it's not 512 nor 4k aligned. So for those files, directIO should be disabled. Not sure we are having issues there.

cclhsu: It seems that DirectIO requires `fileSize > 4096` intead of `filesize % 4096`, so I have made modifications as following:

```
if `directIO == true` and `fileSize >= 4096`, then server/client use directIO.
if `directIO == true` and `fileSize < 4096`, then server/client do not use directIO.
if `directIO == false` and `fileSize >= 4096`, then server/client do not use directIO.
if `directIO == false` and `fileSize < 0`, then server/client do not use directIO.
```

And add unit test as in https://github.com/longhorn/sparse-tools/pull/70

shuo-wu: Is `fileSize >= 4096` enough for DirectIO? Is `fileSize % 512 == 0` unnecessary?

cclhsu: I have created a few [test cases](https://github.com/longhorn/sparse-tools/pull/70/files#diff-5f8226886adcb74f851e23e6bcdf39310d3e63cdc1feb8f6d85895d00ca52502).  After running the test with different configurations, It seems that error only occurred when file size is less than 4096 KB.  When I test what test case (4096 KB + 256 KB) for file size, `directIO` is able to function properly and `checkSparseFiles` is also passed.

shuo-wu: 4MB + 256KB? Or you mean 4MB + 256B? If it's 256KB, this mathes the 512B alignment...

shuo-wu: OK. It seems that the sector size depends on the filesystem type. For EXT4, the default sector size is 512B rather than 4KB
```
root@shuo-cluster-0-worker-1:~# fdisk -l
...

Disk /dev/sda: 79.5 GiB, 85362475008 bytes, 166723584 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes

...
```

I am not sure why we use 4K alignment in longhorn sparse file repo. @yasker Do you have any contexts?

Right now can you try 2 tests with directIO forcibly enabled first:
1. R/W a file that is 512B alignment but not 4KB alignment.
2. R/W a file that is neither 512B alignment nor 4KB alignment.

cclhsu: I think the reason we use 4k is to improve write speed.

cclhsu: I created files 
```
rm -rf /tmp/test*.tmp
dd iflag=fullblock if=/dev/random of=/tmp/test4096KB.tmp bs=4k count=25600 && sync
dd iflag=fullblock if=/dev/random of=/tmp/test512KB.tmp bs=512 count=204800 && sync
dd iflag=fullblock if=/dev/random of=/tmp/test256KB.tmp bs=256 count=409600 && sync
ls -alh /tmp/test*.tmp
```
start daemon
```
FILE=/tmp/file256KB.tmp # /tmp/file.qcow2 | /tmp/file4096KB.tmp | /tmp/file512KB.tmp | /tmp/file256KB.tmp
cd ~/src/github.com/cclhsu/sparse-tools/bin
rm -rf ${FILE}
./ssync -verbose -daemon ${FILE}
```
start client
```
~/src/github.com/cclhsu/sparse-tools/bin
FILE=/tmp/test256KB.tmp # ~/Documents/myImages/kvm/ubuntu/ubuntu-20.04-server-cloudimg-arm64.img | /tmp/test4096KB.tmp | /tmp/test512KB.tmp | /tmp/test256KB.tmp
./ssync -host 0.0.0.0 ${FILE}
```
It seems that diff and size are the same. For Qcow2 image, it is checked with command `qemu-img check` with not errors.

cclhsu: > I am not sure why we use 4K alignment in longhorn sparse file repo. @yasker Do you have any contexts?

@yasker do you know any information on this?

yasker: Yes, 4k was used before introducing the backing image feature, because then we can save 8x times of space for tracking a block/sector(each sector needs to be tracked inside memory for us to know where to read/write), also we might increase the performance in that way. However with qcow2 it was 512B aligned, so IIRC we basically switch to 512 later.

I am fine with either now. Stick with 4k for the size alignment is probably easier. But internally, at least with qcow2 backing image, we have to use 512B as sector size inside the replica/engine.

@shuo-wu For directIO, every read/write have to be aligned with the block size underneath. So it's not `fileSize > 4k`, but `filesize % 4k == 0` or `filesize % 512 == 0`.

@cclhsu I am not sure why we have files that are less than 4k but required for directIO. I would rather stick to 4k alignment for all directIO needed files if possible. The replica size should be always multiples of 4k IIRC.

yasker: The reason for the failure of

```
time="2021-02-24T08:33:54Z" level=fatal msg="Error running add replica command: failed to prepare rebuilding replica tcp://localhost:10015 for volume localhost:30001: rpc error: code = Unknown desc = failed to send file volume-head-002.img.meta to localhost:10018: rpc error: code = Unknown desc = syncDataInterval [       0:       1](  1) failed, err: failed to get checksums for client and server, server checksum error: <nil>, client checksum error: File to read interval:[       0:       1](  1), error: read volume-head-002.img.meta: invalid argument"
```

is this is a metadata file. Metadata file (ending with `.meta`) isn't designed to be transferred using directIO. The directIO is used to persist data to the disk for the images (files ending with `.img`), not metadata files.

pkg/sync/rpc/server.go
```diff
@@ -362,7 +369,8 @@ func (s *SyncAgentServer) launchReceiver(processName, toFileName string, ops spa
 		}()
 
 		logrus.Infof("Running ssync server for file %v at port %v", toFileName, port)
-		if err = sparserest.Server(strconv.Itoa(port), toFileName, ops); err != nil && err != http.ErrServerClosed {
+		directIO := true
+		if err = sparserest.Server(strconv.Itoa(port), toFileName, ops, directIO); err != nil && err != http.ErrServerClosed {
```
shuo-wu, line 373: Wait... Is it meaningful to add the `directIO` flag in the receiver side? Should the sender tell the receiver if it needs to use direct io to write the data into disk?

cclhsu: yes, with mod, it no longer make sense to have it on the server side. I have removed it and wait for the integration result.

innobead (approved):
LGTM

@shuo-wu do you have any concerns?

shuo-wu (changes requested):
you can use a single commit rather than 2 commits with the same title

shuo-wu (approved):

PR #565, opened 2021-03-03
Freeze a volumes filesystem if mounted before taking a snapshot

joshimoo:
We evaluate whether the fs is mounted by running `findmnt /dev/longhorn/<volume_name>` in the host mount namespace.

Since if the volume is used, the engine will be on the same node as the workload
that is using it and the volume will be mounted by kubelet in the host mount namespace.

In the case where the volume is not mounted we fall back to a system level sync.

I added an integration test, that mounts the device in the host mount namespace and writes some file onto it. Then does snapshots & restores.
I need to mount&umount the device after the restore, since the test will shutdown the frontend. This is equivalent to our killing of the workload pods, to get the volume to remount.

longhorn/longhorn#2187

depends on an update of the go-iscsi-helper dependency which is included
as a dev commit and will be replaced once this pr is merged.
https://github.com/longhorn/go-iscsi-helper/pull/38

### DIFF
--- a/integration/data/test_snapshot.py
+++ b/integration/data/test_snapshot.py
@@ -1,4 +1,6 @@
 import random
+import subprocess
+import pytest
 
 import common.cmd as cmd
 
@@ -9,6 +11,7 @@ from common.core import (  # NOQA
     Data, random_length, random_string,
     expand_volume_with_frontend,
     wait_and_check_volume_expansion,
+    checksum_dev
 )
 
 from common.constants import (
@@ -16,6 +19,8 @@ from common.constants import (
     VOLUME_NAME, VOLUME_BACKING_NAME,
     PAGE_SIZE, SIZE, EXPANDED_SIZE,
 )
+
+from common.util import checksum_data
 
 from data.snapshot_tree import snapshot_tree_build, snapshot_tree_verify_node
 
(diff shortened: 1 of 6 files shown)


pkg/controller/control.go
```diff
@@ -166,15 +169,28 @@ func (c *Controller) addReplica(address string, snapshotRequired bool, mode type
...
+		// we always try to unfreeze the volume after finishing the snapshot process
+		// if the user manually froze the filesystem this will lead to an unfreeze but it's better than
+		// potentially having a left over frozen volume.
+		defer unfreezeVolume(ne, c.Name)
+		isFrozen, err := freezeVolume(ne, c.Name)
+		if err != nil {
+			log.WithError(err).Error("WARNING: failed to freeze the volume filesystem falling back to sync")
+		}
+
+		if !isFrozen {
```
shuo-wu, line 188: We may need to be careful about the error from `unfreezeFilesystem`. See the comment in `unfreezeFilesystem` 

```
syncRequired := false
mountPoint, err := findVolumeMountPoint()
if err != nil || mountPoint == "" {
    log()
    syncRequired := true
}
if mountPoint != "" { 
    defer func() {
        // make sure there is no error here if the below freezing failed
        err = unfreeze()
    }()
    if err := freeze();err != nil {
        log()
        syncRequired := true
    }
}
if syncRequired {
    ......
}
```

pkg/controller/control.go
```diff
@@ -901,3 +917,117 @@ func (c *Controller) BackupReplicaMappingDelete(id string) error {
...
+	if err = unfreezeFilesystem(ne, mountPoint); err != nil {
+		logrus.WithField("volume", volume).WithError(err).Errorf("Failed to unfreeze the filesystem at %v", mountPoint)
+		return err
+	}
+
+	logrus.WithField("volume", volume).Infof("Unfroze filesystem of volume mounted at %v", mountPoint)
+	return nil
+}
+
+func findMountPoint(ne *iutil.NamespaceExecutor, devicePath string) (string, error) {
```
shuo-wu, line 961: Not sure if we need to consider the case that there are multiple mount points on a device.

pkg/controller/control.go
```diff
@@ -901,3 +917,117 @@ func (c *Controller) BackupReplicaMappingDelete(id string) error {
...
+	if err != nil && errors.As(err, &ee) && ee.ExitCode() == 1 {
+		// most likely the filesystem is already frozen or the mount point doesn't exist
+		logrus.WithError(err).Debugf("Freeze of filesystem mounted at %v failed", mountPoint)
+		return nil
+	}
+
+	return err
+}
+
+func unfreezeFilesystem(ne *iutil.NamespaceExecutor, mountPoint string) error {
```
shuo-wu, line 1011: if `unfreeze` failed due to some reasons other than invalid parameters/path, will it affect the volume io?

integration/data/test_snapshot.py
```diff
@@ -366,3 +371,96 @@ def test_expansion_with_backing_file(grpc_backing_controller,  # NOQA
...
+
+
+def test_snapshot_mounted_filesystem(grpc_controller,  # NOQA
+                         grpc_replica1, grpc_replica2):  # NOQA
+    address = grpc_controller.address
+    dev = get_dev(grpc_replica1, grpc_replica2, grpc_controller)
+    snapshot_mounted_filesystem_test(VOLUME_NAME, dev, address, ENGINE_NAME)
+
+
+def snapshot_mounted_filesystem_test(volume_name, dev, address, engine_name):  # NOQA
```
shuo-wu, line 383: Miss the test skeleton.

integration/data/test_snapshot.py
```diff
@@ -366,3 +371,96 @@ def test_expansion_with_backing_file(grpc_backing_controller,  # NOQA
...
+
+    print("\nsnapshot_revert_with_frontend snap3 begin")
+    snapshot_revert_with_frontend(address, engine_name, snap3)
+    print("snapshot_revert_with_frontend snap3 finish\n")
+    subprocess.check_call(mount_cmd)
+    assert checksum_test_file() == test3_checksum
+    subprocess.check_call(umount_cmd)
+
+    # remove the created mount folder
+    subprocess.check_call(nsenter_cmd + ["rmdir", mnt_path])
```
shuo-wu, line 466: The mount directory is not empty. I think this doesn't work

integration/data/test_snapshot.py
shuo-wu: And does this case test the mounted volume only?

shuo-wu (changes requested):

pkg/controller/control.go
```diff
@@ -52,6 +54,7 @@ type Controller struct {
 }
 
 const (
+	freezeTimeout         = 60 * time.Minute // qemu uses 60 minute timeouts for freezing
```
innobead, line 57: Could you provide the reference link about qemu case you menteiond here?

innobead (commented):
Overall, LGTM

Shuo also provided some good feedbacks waiting for you to resolve or answer.

integration/data/test_snapshot.py
joshimoo: The volume is unmounted as such the folder used for the mount point does not contain any data?

integration/data/test_snapshot.py
joshimoo: What do you mean? the test mounts the volume into the host namespace, so that it tests the fsfreeze implementation.
The sync fallback is tested by all the other snapshot tests.

pkg/controller/control.go
joshimoo: I only look for the `--first-only` since we are freezing the filesystem not the mount point.
Additional mount points are irrelevant.

pkg/controller/control.go
joshimoo: Sure, this is the change that got merged.
https://pve-devel.pve.proxmox.narkive.com/FCClXQF6/rfc-qemu-server-increase-timeout-from-guest-fsfreeze-freeze

Based on the discussion here:
https://forum.proxmox.com/threads/22192/

pkg/controller/control.go
joshimoo: What is your concern about looking for the mount point again?

shuo-wu: I mean, you will do mount point finding twice in `freezeVolume()` and `unfreezeVolume`, which is unnecessary.
`freeze` and `unfreeze` here means the operations to the filesystem, rather than invoking `freezeVolume()` and `unfreezeVolume`

integration/data/test_snapshot.py
shuo-wu: OK, got it. This function won't test the sync fallback.

integration/data/test_snapshot.py
shuo-wu: Oh, sorry.

pkg/controller/control.go
joshimoo: If a frozen filesystem does not get unfrozen then it remains frozen.
Deferred functions will be called even in the presence of a panic.

pkg/controller/control.go
joshimoo: I find it better to search for the mount point twice, since if for whatever reason the mountpoint has changed (old one removed, new one added) we would still unfreeze the filesystem. While in the case you are proposing, the freeze would succeed but the unfreeze would fail. The cost of the extra findmnt should be small since it's only done as part of the snapshot operation.

pkg/controller/control.go
shuo-wu: I mean, should Longhorn return an error and mark the volume as failed if the unfreezing failed? Then Longhorn or users would restart or clean up it.

joshimoo: There isn't really a good way to tell whether a filesystem is frozen. So it's naturally that unfreeze will fail in the case where the filesystem is not frozen. Unfreezing should not fail for a frozen filesystem. If the volume gets detached from the node and attached again the filesystem will be in an unfrozen state, even if previously frozen.

So in the case where the node crashed after the freeze but before the unfreeze, the volume will be detached and reattached on a new node which will mean the filesystem is unfrozen.

joshimoo:
@shuo-wu any outstanding concerns?

shuo-wu (approved):

PR #566, opened 2021-03-04
add logs for debugging snapshot (#2161)

cclhsu:
#### Proposed Changes ####

add logs for debugging snapshot to resolve ticket '[FEATURE]Add more debug logs for snapshotting' (#2161)

- add snapshot time
- add sendfile time

#### Types of Changes ####

- enhancement

#### Verification ####

Additional log like following

2021-03-04T04:10:53.970318517Z time="2021-03-04T04:10:53Z" level=info msg="Finish snapshot for tcp://10.42.1.12:10000: 075a33e7-343b-4cad-b0c6-ff9c19016ec5 and take 2.63582ms"

2021-03-04T04:10:53.970318517Z time="2021-03-04T04:10:53Z" level=info msg="Finish sending file /tmp/test256KB.tmp and take 1.644776033s"

#### Linked Issues ####

Refs: # https://github.com/longhorn/longhorn/issues/2161

#### Further Comments ####

Wait for https://github.com/longhorn/sparse-tools/pull/71 to update go.mod, go.sum, and module.txt

Signed-off-by: Clark Hsu <clark.hsu@suse.com>

### DIFF
--- a/pkg/controller/control.go
+++ b/pkg/controller/control.go
@@ -173,8 +173,9 @@ func (c *Controller) addReplica(address string, snapshotRequired bool, mode type
 // and will fallback to a system level sync in all other cases
 func (c *Controller) Snapshot(name string, labels map[string]string) (string, error) {
 	log := logrus.WithFields(logrus.Fields{"volume": c.Name, "snapshot": name})
+	log.Info("Starting snapshot")
 	if ne, err := iutil.NewNamespaceExecutor(util.GetInitiatorNS()); err != nil {
-		log.WithError(err).Error("WARNING: cannot get namespace executor for volume sync/freeze")
+		log.WithError(err).Errorf("WARNING: continue to snapshot for %v, but cannot sync due to cannot get the namespace executor", name)
 	} else {
 		// we always try to unfreeze the volume after finishing the snapshot process
 		// if the user manually froze the filesystem this will lead to an unfreeze but it's better than
@@ -182,14 +183,14 @@ func (c *Controller) Snapshot(name string, labels map[string]string) (string, er
 		defer unfreezeVolume(ne, c.Name)
 		isFrozen, err := freezeVolume(ne, c.Name)
 		if err != nil {
-			log.WithError(err).Error("WARNING: failed to freeze the volume filesystem falling back to sync")
+			log.WithError(err).Errorf("WARNING: failed to freeze the volume filesystem falling back to sync for %v", name)
 		}
 
 		if !isFrozen {
 			log.Info("Volume is not frozen, requesting system sync before snapshot")
 			if _, err := ne.ExecuteWithTimeout(syncTimeout, "sync", []string{}); err != nil {
 				// sync should never fail though, so it more like due to the nsenter
-				log.WithError(err).Errorf("WARNING: failed to sync continuing with snapshot")
+				log.WithError(err).Errorf("WARNING: failed to sync continuing with snapshot for %v", name)
 			}
 		}
 	}
@@ -211,6 +212,7 @@ func (c *Controller) Snapshot(name string, labels map[string]string) (string, er
 	if err := c.handleErrorNoLock(c.backend.Snapshot(name, true, created, labels)); err != nil {
 		return "", err
 	}
+	log.Info("Finished snapshot")
 	return name, nil
 }
 
--- a/pkg/backend/remote/remote.go
+++ b/pkg/backend/remote/remote.go
@@ -83,7 +83,7 @@ func (r *Remote) open() error {
 }
 
 func (r *Remote) Snapshot(name string, userCreated bool, created string, labels map[string]string) error {
-	logrus.Infof("Snapshot: %s %s UserCreated %v Created at %v, Labels %v",
+	logrus.Infof("Starting to snapshot: %s %s UserCreated %v Created at %v, Labels %v",
 		r.name, name, userCreated, created, labels)
 	conn, err := grpc.Dial(r.replicaServiceURL, grpc.WithInsecure())
 	if err != nil {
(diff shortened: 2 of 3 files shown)


shuo-wu (changes requested):
1. Need to add logs for [this function}(https://github.com/longhorn/longhorn-engine/blob/76cf784dda530f25e75bad2128a5ff0b15617c97/pkg/controller/control.go#L169) so that we can check if the engine receives/finishes the snapshot request.
2. add logs for [the replica](https://github.com/longhorn/longhorn-engine/blob/76cf784dda530f25e75bad2128a5ff0b15617c97/pkg/replica/replica.go#L1033) as well
3. Don't need to calculate the elapsed time in logs. The timestamp is enough.
4. Be consistent with [this PR](https://github.com/longhorn/longhorn-engine/pull/565/commits/9f0579135746559ae93e1b4df168149cdccbbf4a#diff-c1efcd0538e5be63994d3d895fcbef838dbcd577e1262fb88d680d1672163477R175): using `logger` rather than `logrus`

cclhsu:
> * Need to add logs for [this function}(https://github.com/longhorn/longhorn-engine/blob/76cf784dda530f25e75bad2128a5ff0b15617c97/pkg/controller/control.go#L169) so that we can check if the engine receives/finishes the snapshot request.
> * add logs for [the replica](https://github.com/longhorn/longhorn-engine/blob/76cf784dda530f25e75bad2128a5ff0b15617c97/pkg/replica/replica.go#L1033) as well

OK, I will add these.

cclhsu:
> 4\. Be consistent with [this PR](https://github.com/longhorn/longhorn-engine/pull/565/commits/9f0579135746559ae93e1b4df168149cdccbbf4a#diff-c1efcd0538e5be63994d3d895fcbef838dbcd577e1262fb88d680d1672163477R175): using `logger` rather than `logrus` for the engine snapshot function

I used logger which is previous used in then *.go file. Not all go file renamed to logger.

joshimoo (commented):
LGTM left some notes about the field usage for loggers.

c3y1huang (changes requested):
Except for the nit on the `.` to be consistent with the other messages, others LGTM.

And I'm not sure if we should keep the `WARNING` in the message since it's been placed as an Error?

pkg/sync/rpc/server.go
```diff
@@ -329,8 +330,12 @@ func (*SyncAgentServer) FileRename(ctx context.Context, req *ptypes.FileRenameRe
 
 func (s *SyncAgentServer) FileSend(ctx context.Context, req *ptypes.FileSendRequest) (*empty.Empty, error) {
 	address := net.JoinHostPort(req.Host, strconv.Itoa(int(req.Port)))
+	directIO := true
+	if filepath.Ext(strings.TrimSpace(req.FromFileName)) == ".meta" {
+		directIO = false
+	}
 	logrus.Infof("Sending file %v to %v", req.FromFileName, address)
-	if err := sparse.SyncFile(req.FromFileName, address, FileSyncTimeout); err != nil {
+	if err := sparse.SyncFile(req.FromFileName, address, FileSyncTimeout, directIO); err != nil {
```
joshimoo, line 338: This seems like a change in functionality, not just adding logs.
Also log whether direct io is being used as part of the log call.

pkg/controller/control.go
```diff
@@ -211,6 +212,7 @@ func (c *Controller) Snapshot(name string, labels map[string]string) (string, er
 	if err := c.handleErrorNoLock(c.backend.Snapshot(name, true, created, labels)); err != nil {
 		return "", err
 	}
+	log.Infof("Finished snapshot")
```
joshimoo, line 215: You can just use `Info` when you don't pass a format string arg.
Same for the other places.

joshimoo (commented):
Your missing the signoff on your commits, if you click the link below it will show you how to signoff your commits.
https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for
https://github.com/longhorn/longhorn-engine/pull/566/checks?check_run_id=2085671526

pkg/sync/rpc/server.go
cclhsu: seems like conflict with other issue https://github.com/longhorn/longhorn-engine/pull/560

joshimoo: I probably misunderstood, but since your log changes are unrelated to the direct io changes in (#560) don't include them as part of this pr. Then we can merge this and sparse-files will be updated as part of the PR that introduces the direct io change?

pkg/controller/replicator.go
```diff
@@ -192,12 +192,15 @@ func (r *replicator) Snapshot(name string, userCreated bool, created string, lab
 		if backend.mode != types.ERR {
 			wg.Add(1)
 			go func(address string, backend types.Backend) {
+				log := logrus.WithFields(logrus.Fields{"snapshot": name, "address": address})
+				log.Info("Starting to snapshot")
 				if err := backend.Snapshot(name, userCreated, created, labels); err != nil {
```
shuo-wu, line 197: There is already a start log in `backend.Snapshot(name, userCreated, created, labels)`. We can improve the logs in the implementation of this interface instead:
1. Add the address in the log
2. Add finish log

The implementation is here:
https://github.com/longhorn/longhorn-engine/blob/0c9b45e92f4c4b5dfd7dae5067d017846dbdc724/pkg/backend/remote/remote.go#L85

shuo-wu (changes requested):
Add a log for the replica server if the snapshot succeeds: https://github.com/longhorn/longhorn-engine/blob/0c9b45e92f4c4b5dfd7dae5067d017846dbdc724/pkg/replica/server.go#L174

pkg/controller/replicator.go
cclhsu: done.

shuo-wu (approved):
LGTM

PR #568, opened 2021-03-10
Support AWS IAM Role

jenting:
#### Proposed Changes
- Allowed AWS credential to be empty (it will check at longhorn-manager) in order to support AWS IAM Role
- Bump longhorn/backupstore version

#### Linked Issues
https://github.com/longhorn/longhorn/issues/1526


innobead (approved):
LGTM

PR #571, opened 2021-04-13
vendor: Update longhorn/sparse-tools

shuo-wu:
longhorn/longhorn#2416


shuo-wu:
Regression test: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/

innobead (approved):
LGTM

PR #617, opened 2021-05-18
add drone configuration for engine live-upgrade and upgrade-test images

meldafrawi:
https://github.com/longhorn/longhorn/issues/1961

Signed-off-by: Mohamed Eldafrawi <mohamed.eldafrawi@rancher.com>

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -222,3 +222,333 @@ steps:
 depends_on:
 - default-amd64
 - default-arm64
+---
+
+kind: pipeline
+name: engine-upgrade-test-images-amd64
+
+platform:
+  os: linux
+  arch: amd64
+
+steps:
+- name: generate new commit
+  image: ubuntu:18.04
+  commands:
+  - apt-get update
+  - apt-get install -y git
+  - git fetch --tags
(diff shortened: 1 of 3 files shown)


meldafrawi:
for test purposes, images contains `staging` word in tag is pushed to docker hub from staging branch

https://hub.docker.com/r/longhornio/longhorn-engine/tags?page=1&ordering=last_updated&name=staging

shuo-wu (commented):
Does this contain incompatible test image automation? https://github.com/longhorn/longhorn-tests/blob/3c8c1e841cb44cef9cf5cb512ccb37a41d1e8404/manager/test_containers/build_test_images.sh#L6-L10

In dockerhub, there is no this kind of images with word `staging`

innobead:
> Does this contain incompatible test image automation? https://github.com/longhorn/longhorn-tests/blob/3c8c1e841cb44cef9cf5cb512ccb37a41d1e8404/manager/test_containers/build_test_images.sh#L6-L10
> 
> In dockerhub, there is no this kind of images with word `staging`

@meldafrawi any comment for @shuo-wu question?

PR #621, opened 2021-05-25
Return a non-zero value of one of the WriteAts is successful.

keithalucas:
This makes MultiWriteAt.WriteAt return len(p) if at least one of the WriteAt's to the replicas was successful.  Previously, if at least one of the WriteAt's failed, it would return 0.  This change has no impact to the behavior of longhorn-engine.

### DIFF
--- a/pkg/controller/multi_writer_at.go
+++ b/pkg/controller/multi_writer_at.go
@@ -33,9 +33,8 @@ func (m *MultiWriterError) Error() string {
 	}
 }
 
-func (m *MultiWriterAt) WriteAt(p []byte, off int64) (n int, err error) {
+func (m *MultiWriterAt) WriteAt(p []byte, off int64) (int, error) {
 	errs := make([]error, len(m.writers))
-	errored := false
 	wg := sync.WaitGroup{}
 
 	for i, w := range m.writers {
@@ -43,20 +42,28 @@ func (m *MultiWriterAt) WriteAt(p []byte, off int64) (n int, err error) {
 		go func(index int, w io.WriterAt) {
 			_, err := w.WriteAt(p, off)
 			if err != nil {
-				errored = true
 				errs[index] = err
 			}
+
 			wg.Done()
 		}(i, w)
 	}
 
 	wg.Wait()
-	if errored {
-		return 0, &MultiWriterError{
-			Writers: m.writers,
-			Errors:  errs,
+
+	n := 0
+	var err error = nil
+
+	for i := range errs {
+		if errs[i] != nil {
+			err = &MultiWriterError{
+				Writers: m.writers,
+				Errors:  errs,
+			}
+		} else {
+			n = len(p)
 		}
 	}
 
-	return len(p), nil
+	return n, err
 }


pkg/controller/multi_writer_at.go
```diff
@@ -33,9 +33,11 @@ func (m *MultiWriterError) Error() string {
 	}
 }
 
-func (m *MultiWriterAt) WriteAt(p []byte, off int64) (n int, err error) {
+func (m *MultiWriterAt) WriteAt(p []byte, off int64) (int, error) {
 	errs := make([]error, len(m.writers))
 	errored := false
+	succeeded := false
+	n := 0
 	wg := sync.WaitGroup{}
```
joshimoo, line 41:
```suggestion
	errs := make([]error, len(m.writers))
	written := make([]int, len(m.writers)
	wg := sync.WaitGroup{}
	wg.Add(len(m.writers))
```

pkg/controller/multi_writer_at.go
```diff
@@ -45,18 +47,26 @@ func (m *MultiWriterAt) WriteAt(p []byte, off int64) (n int, err error) {
 			if err != nil {
```
joshimoo, line 47:
```go
deffer wg.Done()
n, err := w.WriteAt(p, off)
errors[index] = err
written[index] = n
```

Now you don't need to keep the state variables and this pattern also doesn't require any syncronisation.

pkg/controller/multi_writer_at.go
```diff
@@ -45,18 +47,26 @@ func (m *MultiWriterAt) WriteAt(p []byte, off int64) (n int, err error) {
...
 		}(i, w)
 	}
 
 	wg.Wait()
+
+	// Return length of p if at least one WriteAt succeeded
+	if succeeded {
+		n = len(p)
+	}
+
```
joshimoo, line 63:
```go
n: = max(written)
var err *MultiWriterError
total := sum(written)
if total < (len(p) * len(wri) {
    err = &MultiWriterError{ Writers: m.writers, Errors:  errs }
}
return n, err
```

This makes the code a little nicer to read, you can either use the total approach or check the errors array explicitly for non null.

joshimoo (commented):
nit: Left you some code suggestions.

Can you comment/explain/show why this change does not lead to any issues in the caller, why this does not change the behavior or the engine?

shuo-wu (changes requested):
1. Do not set the commit title to `longhorn/longhorn#1917`. Instead, put this issue number to the commit comment.
2. Does this commit fix the issue? Or will you have one more commit for the issue later? The caller should do something when one writer succeeds in writing all content.

keithalucas:
> nit: Left you some code suggestions.
> 
> Can you comment/explain/show why this change does not lead to any issues in the caller, why this does not change the behavior or the engine?

The return value doesn't seem to be used.  The only place I could see that it was used is in https://github.com/longhorn/longhorn-engine/blob/master/pkg/dataconn/server.go#L95.  I couldn't make the condition where err == io.EOF occur so that the count was used.  I tested this by opening the device file and lseeking to a position near the end and writing a block a data that would go past the end.  I also made this WriteAt always return 0 and I could use the block device without any problems.

pkg/controller/multi_writer_at.go
keithalucas: I had a discussion with Joshua and we came up with a solution that doesn't have unsafe variable access from goroutines and doesn't compute the max and sum.

keithalucas:
> 1. Do not set the commit title to `longhorn/longhorn#1917`. Instead, put this issue number to the commit comment.
> 2. Does this commit fix the issue? Or will you have one more commit for the issue later? The caller should do something when one writer succeeds in writing all content.

This does not fix the issue.

pkg/controller/multi_writer_at.go
```diff
@@ -22,7 +22,6 @@ func (m *MultiWriterError) Error() string {
 			errors = append(errors, err.Error())
 		}
 	}
-
```
yasker, line 25: Unrelated change.

yasker (changes requested):
Can you squash the commits into one? The commits should be separated by the function in general. There are times that you can create a commit just for the convenience of review, in case the change was big and complex, but it should be separated properly by function before the merge. In this case, it's better to always squash into the one commit since it only does one thing and the change is small.

Also, as Shuo mentioned, the issue number should be in the commit message instead of title. Otherwise, it will be confusing if we're looking at `git log`.

And in the commit message, we should call it out explicitly this doesn't affect the upper layer function and explain briefly why. So when someone look at the history of the commits or use `git blame`, they can understand why this commit was made for what reason.

And we need to call out it doesn't fix the issue, due to we are unable to reproduce the issue.

pkg/controller/multi_writer_at.go
keithalucas: I removed it.

keithalucas:
> Can you squash the commits into one? The commits should be separated by the function in general. There are times that you can create a commit just for the convenience of review, in case the change was big and complex, but it should be separated properly by function before the merge. In this case, it's better to always squash into the one commit since it only does one thing and the change is small.
> 
> Also, as Shuo mentioned, the issue number should be in the commit message instead of title. Otherwise, it will be confusing if we're looking at `git log`.
> 
> And in the commit message, we should call it out explicitly this doesn't affect the upper layer function and explain briefly why. So when someone look at the history of the commits or use `git blame`, they can understand why this commit was made for what reason.
> 
> And we need to call out it doesn't fix the issue, due to we are unable to reproduce the issue.
> 
> The original issue should be kept open. For this PR, we won't able to test it explicitly. Any side effects should be shown from our nightly testing result.

I squashed the multiple commits, explained why this doesn't impact longhorn-engine, and mentioned that this was done as part of the analysis of #1917 but does not fix it.

yasker (changes requested):
LGTM now. A small issue regarding the commit message: it's a bit too much now. :) 

We can read most of the flow through the code or upcoming article you are working on.

I would suggest keeping the last part `pkg/dataconn/sever.go` to explain why the return result doesn't matter.

Though regarding the `io.EOF`. Is it the right behavior for us to not returning it? And should we treat `io.EOF` as an error in the `MultiWriteError`?

yasker:
Btw, we have a race condition bug in the engine integration test now, which is why the test was failing last time. https://github.com/longhorn/longhorn/issues/1628 . The failed test log is from https://drone-pr.longhorn.io/longhorn/longhorn-engine/330/1/3 , which seems slightly different though.

pkg/controller/multi_writer_at.go
```diff
@@ -33,30 +33,37 @@ func (m *MultiWriterError) Error() string {
...
+
+	n := 0
+	var err error = nil
+
+	for i := range errs {
+		if errs[i] != nil {
+			err = &MultiWriterError{
+				Writers: m.writers,
+				Errors:  errs,
+			}
```
shuo-wu, line 62: This function can `break` from this loop if there is an error. 
BTW, any reason for removing the boolean `errored` and using this loop instead?

shuo-wu (changes requested):

pkg/controller/multi_writer_at.go
shuo-wu: Make sense. Thank you for your explanation.

shuo-wu (approved):

yasker:
Ping @keithalucas . I haven't seen an update since my comment on https://github.com/longhorn/longhorn-engine/pull/621#pullrequestreview-670425457 ?

keithalucas:
> LGTM now. A small issue regarding the commit message: it's a bit too much now. :)
> 
> We can read most of the flow through the code or upcoming article you are working on.
> 
> I would suggest keeping the last part `pkg/dataconn/sever.go` to explain why the return result doesn't matter.
> 
> Though regarding the `io.EOF`. Is it the right behavior for us to not returning it? And should we treat `io.EOF` as an error in the `MultiWriteError`?

Sorry, I took the LGTM as no further action needed. I updated the commit message.  I am apprehensive about changing the behavior in regarding io.EOF because I can't make io.EOF occur.

yasker (approved):

PR #623, opened 2021-05-30
Support new backup list behavior and backup inspect-volume command

jenting:
#### Propose Changes

- Bump longhorn/backupstore to support new `backup list` behavior and `backup inspect-volume` command.
- Updates integration test for new `backup list` behavior and `backup inspect-volume` command.
- Create a new backup supports passing backup name.

#### Issue

https://github.com/longhorn/longhorn/issues/1761


app/cmd/backup.go
```diff
@@ -5,12 +5,11 @@ import (
...
 	"strings"
 
+	"github.com/longhorn/backupstore/cmd"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 
-	"github.com/longhorn/backupstore/cmd"
-
 	"github.com/longhorn/longhorn-engine/pkg/sync"
```
shuo-wu, line 13:
```
"github.com/pkg/errors"
"github.com/sirupsen/logrus"
"github.com/urfave/cli"

"github.com/longhorn/backupstore/cmd"

"github.com/longhorn/longhorn-engine/pkg/sync"
......
```

shuo-wu (changes requested):
You can test the updated backup creation command in the integration test as well.

jenting:
Depends on https://github.com/longhorn/longhorn-engine/pull/631

jenting:
> Depends on #631

This is ready to review, the failed integration test is depends on #631.

go.sum
```diff
@@ -55,6 +55,7 @@ github.com/moby/moby v1.11.1/go.mod h1:fDXVQ6+S340veQPv35CzDahGBmHsiclFwfEygB/TW
 github.com/pkg/errors v0.8.1/go.mod h1:bwawxfHBFNV+L2hUp1rHADufV3IMtnDRdf1r5NINEl0=
 github.com/pkg/errors v0.9.1 h1:FEBLx1zS214owpjy7qsBeixbURkuhQAwrK5UwLGTwt4=
 github.com/pkg/errors v0.9.1/go.mod h1:bwawxfHBFNV+L2hUp1rHADufV3IMtnDRdf1r5NINEl0=
+github.com/pkg/sftp v1.10.1/go.mod h1:lYOWFsE0bwd1+KfKJaKeuokY15vzFx25BLbzYYoAxZI=
```
joshimoo, line 58: There seems to be a bunch of new packages, also there are older ones like windows-terminal.
Which make no sense, since we don't support windows at all.

integration/core/test_cli.py
```diff
@@ -751,21 +754,22 @@ def backup_core(bin, engine_manager_client,  # NOQA
...
-    assert volume_info[VOLUME_NAME]["Messages"] is not None
-    assert MESSAGE_TYPE_ERROR not in volume_info[VOLUME_NAME]["Messages"]
+    url = get_backup_volume_url(backup_target, VOLUME_NAME)
+    volume_info = cmd.backup_inspect_volume(
+        grpc_controller_client.address, url)
+    assert volume_info["Messages"] is not None
+    assert MESSAGE_TYPE_ERROR not in volume_info["Messages"]
 
     # backup doesn't exists so it should error
+    url = backup_target + "?backup=backup-unk" + "&volume=" + VOLUME_NAME
```
joshimoo, line 771: Any reason this was moved from the inner scope?

go.sum
joshimoo: Did you by chance update your go version?

joshimoo (commented):
I only skimmed the integration test changes, I like the ability of passing a backup name.
But even further in the long run I would like to have a 1 to 1 mapping between snapshots <-> backups. So there is never multiple backups for a single snapshot, this could be accomplished by setting backupName == snapshotName.

joshimoo:
CC @innobead @meldafrawi @khushboo-rancher I would like QA to also have a look at the engine integration tests that way they get knowledge and exposure to this component as well.

go.sum
jenting: I use `go1.16.5`. I'll check the package relationship.

integration/core/test_cli.py
jenting: can't remember, revert it.

jenting:
> I only skimmed the integration test changes, I like the ability of passing a backup name.
> But even further in the long run I would like to have a 1 to 1 mapping between snapshots <-> backups. So there is never multiple backups for a single snapshot, this could be accomplished by setting backupName == snapshotName.

okay, I'll make it on the longhorn-manager to have 1 to 1 mapping between snapshots <-> backups.

go.sum
jenting: The windows-terminal comes from package `github.com/longhorn/backupstore` -> `github.com/sirupsen/logrus` -> `github.com/konsorten/go-windows-terminal-sequences`.

shuo-wu:
> > I only skimmed the integration test changes, I like the ability of passing a backup name.
> > But even further in the long run I would like to have a 1 to 1 mapping between snapshots <-> backups. So there is never multiple backups for a single snapshot, this could be accomplished by setting backupName == snapshotName.
> 
> okay, I'll make it on the longhorn-manager to have 1 to 1 mapping between snapshots <-> backups.

Thank you for the notification. Updated the PR.

yasker:
@joshimoo it's hard to get it to be one to one mapping. 
1. Snapshot or backup can be deleted, so the other one might not exist.
2. Backup is suppose to be separated concept from snapshot. It's from a snapshot, but doesn't mean it's bound to a snapshot. 
I don't feel we need to enforce the constraints and say snapshot to backup must be 1:1 mapping.

Any reason we prefer having 1:1 mapping from snapshot to backup?

joshimoo:
@jenting discussed with @yasker currently we don't want to use the 1 to 1 mapping from snapshot -> backup.
Since it would require additional validation logic, for example with snapshot name conflicts.
Also we have no migration path at this point in time.
We could consider this down the line in the future, but it's really not a requirement and we are getting so much already from just having the backup crds, so let's not worry about this small detail.

jenting:
> @jenting discussed with @yasker currently we don't want to use the 1 to 1 mapping from snapshot -> backup.
> Since it would require additional validation logic, for example with snapshot name conflicts.
> Also we have no migration path at this point in time.
> We could consider this down the line in the future, but it's really not a requirement and we are getting so much already from just having the backup crds, so let's not worry about this small detail.

Then, should the Longhorn CSI plugin still need the 1 to 1 mapping from snapshot -> backup? https://github.com/longhorn/longhorn-manager/blob/cec2e82/csi/controller_server.go#L457-L464

joshimoo:
> Then, should the Longhorn CSI plugin still need the 1 to 1 mapping from snapshot -> backup? https://github.com/longhorn/longhorn-manager/blob/cec2e82/csi/controller_server.go#L457-L464

Yes that is still required but once we have a persistent backup crd, we can change that logic to rely on the backup crd instead of having to traverse and find a valid snapshot.

jenting:
> > Then, should the Longhorn CSI plugin still need the 1 to 1 mapping from snapshot -> backup? https://github.com/longhorn/longhorn-manager/blob/cec2e82/csi/controller_server.go#L457-L464
> 
> Yes that is still required but once we have a persistent backup crd, we should be able to rely on the backup crd, instead of doing this eval & traversal.

Okay, then I have to make a workaround on the Longhorn API on getting BackupVolume https://github.com/longhorn/longhorn-manager/blob/cec2e82/csi/controller_server.go#L448.
After my backup refactoring enhancement, if the BackupVolume CR does not exist, get the non-existing BackupVolume returns error. However, to fulfills the Longhorn CSI plugin code, I have to make it return success although personally I don't like this behavior.

jenting:
This is ready to review.

app/cmd/backup.go
```diff
@@ -59,8 +59,8 @@ func BackupCreateCmd() cli.Command {
 				Usage: "specify labels for backup, in the format of `--label key1=value1 --label key2=value2`",
 			},
 			cli.StringFlag{
-				Name:  "backing-image-url",
-				Usage: "deprecated. specify backing image URL of the volume for backup",
+				Name:  "backup-name",
+				Usage: "specify the backup name",
```
shuo-wu, line 63: `If it is not set, a random name will be generated automatically.`

shuo-wu (approved):
In general LGTM

innobead (approved):
LGTM

After resolving @shuo-wu comment, we can merge this.

innobead (approved):

PR #625, opened 2021-06-08
Backport replica rebuild enhancement

joshimoo:
Backport sparse-tools update for iterative sync/fold

This also has a typo fix for the go-iscsi-helper dependency this is already
in master, since it doesn't have a semantic change I included it as well.

longhorn/longhorn#1948
longhorn/longhorn#2507
longhorn/longhorn#2640


joshimoo:
Since these are backports of dependency updates, I decided to merge them without review.
Since the code had not changed between the original one.

PR #626, opened 2021-06-16
Speed up backup creation: Avoid using channel in CompareSnapshot

shuo-wu:
longhorn/longhorn#2218

### DIFF
--- a/pkg/controller/client/controller_client.go
+++ b/pkg/controller/client/controller_client.go
@@ -20,7 +20,7 @@ type ControllerClient struct {
 }
 
 const (
-	GRPCServiceTimeout = 1 * time.Minute
+	GRPCServiceTimeout = 3 * time.Minute
 )
 
 func NewControllerClient(address string) *ControllerClient {
--- a/pkg/replica/backup.go
+++ b/pkg/replica/backup.go
@@ -356,13 +356,8 @@ func preload(d *diffDisk) error {
 			}
 		}
 
-		generator := newGenerator(d, f)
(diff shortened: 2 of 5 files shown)


jenting (approved):
Not sure had you test the max file size could be finished within 3 mins or not.
For example 16TB.

shuo-wu:
> Not sure had you test the max file size could be finished within 3 mins or not.
> For example 16TB.

This is not the ultimate solution for the timeout issue. But removing channel would speed up the process.

joshimoo (approved):
Discussed with @shuo-wu this is sort of a hack to get it into v1.1.2
For v1.2 we can parallesize the creation of the sectors -> file map.

map: Either by utilising a bitsets for each file
reduce: Then merging them into the final sector->file map.

map: Or by having each go routine create & write on a buffered output channel of the form:
Location{
sector uint64
fileID byte
}

reduce: Then merging the channels unified output channel which can then be processed by the main routine for writing to the sector -> file map since a higher fileID overwrites a lower fileID for a sector, this process will lead to the correct sector -> fileID map.

joshimoo:
Leaving a note about the memory consumption for the bit set approach.
```
sectorsPerGB = (1024×1024×1024)÷(4×1024) = 262144
bytesPerGB = sectorsPerGB / 8 = 32768 bytes = 32KB

So for each GB of filesize we only need to store 32KB in the bitset.
100 GB = 32 * 100 = 3.125MB
1TB = 31.25MB

```

So the memory consumption is quite small for the bitset approach.

PR #630, opened 2021-06-21
Timeout replica when no I/O responses are received

keithalucas:
This changes the timeout logic in dataconn.Conn.  Previously the timeout occurred when a response operation was not received for eight seconds (with one retry).   When multiple operations were being performed simultaneously on a slow disk, this timeout was likely to occur.  However, when multiple operations were being performed simultaneously, the replica is regularly sending responses to those operations.  So, this change makes the timeout occur when no response to any of the pending read or write operation has been received for eight seconds. 

This will allow slow disks to not timeout because the replica will be completing and sending the responses to operation regularly.   This addresses slow disks failing as part of #2206 but doesn't fixed the performance issue will spinning disks in general.

https://github.com/longhorn/longhorn/issues/2206

### DIFF
--- a/pkg/backend/remote/remote.go
+++ b/pkg/backend/remote/remote.go
@@ -1,6 +1,7 @@
 package remote
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	"strconv"
@@ -20,9 +21,12 @@ import (
 )
 
 var (
-	pingInveral   = 2 * time.Second
-	timeout       = 30 * time.Second
-	requestBuffer = 1024
+	ErrPingTimeout = errors.New("ping timeout")
+)
+
+const (
+	PingInterval = 2 * time.Second
+	PingTimeout  = 8 * time.Second
 )
 
 func New() types.BackendFactory {
@@ -284,7 +288,10 @@ func (rf *Factory) Create(address string) (types.Backend, error) {
 }
 
 func (r *Remote) monitorPing(client *dataconn.Client) {
-	ticker := time.NewTicker(pingInveral)
+	pingStatus := make(chan error, 1)
+	pingDeadline := time.Time{}
+	inProgress := false
+	ticker := time.NewTicker(PingInterval)
 	defer ticker.Stop()
 
 	for {
@@ -293,13 +300,35 @@ func (r *Remote) monitorPing(client *dataconn.Client) {
 			r.monitorChan <- nil
 			return
 		case <-ticker.C:
-			if err := client.Ping(); err != nil {
+			if !inProgress {
+				inProgress = true
+				pingDeadline = time.Now().Add(PingTimeout)
+				r.ping(client, pingStatus)
+				continue
+			}
+
+			if time.Now().After(pingDeadline) {
+				client.SetError(ErrPingTimeout)
+				r.monitorChan <- ErrPingTimeout
+				return
+			}
+
+		case err := <-pingStatus:
+			inProgress = false
+
+			if err != nil {
 				client.SetError(err)
 				r.monitorChan <- err
 				return
 			}
 		}
 	}
+}
+
+func (r *Remote) ping(client *dataconn.Client, status chan<- error) {
+	go func() {
+		status <- client.Ping()
+	}()
 }
 
 func (r *Remote) GetMonitorChannel() types.MonitorChannel {
--- a/pkg/dataconn/client.go
+++ b/pkg/dataconn/client.go
@@ -13,13 +13,11 @@ import (
 
 var (
 	//ErrRWTimeout r/w operation timeout
-	ErrRWTimeout   = errors.New("r/w timeout")
-	ErrPingTimeout = errors.New("Ping timeout")
+	ErrRWTimeout = errors.New("r/w timeout")
+)
 
-	opRetries      = 2
-	opReadTimeout  = 4 * time.Second // client read
-	opWriteTimeout = 4 * time.Second // client write
-	opPingTimeout  = 8 * time.Second
+const (
+	OPTimeout = 8 * time.Second
 )
 
 //Client replica client
@@ -32,7 +30,6 @@ type Client struct {
 	messages  map[uint32]*Message
 	wire      *Wire
 	peerAddr  string
-	err       error
 }
 
 //NewClient replica client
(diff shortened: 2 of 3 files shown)


pkg/dataconn/client.go
```diff
@@ -33,6 +30,8 @@ type Client struct {
 	wire      *Wire
 	peerAddr  string
 	err       error
+	timer     *time.Timer
```
yasker, line 33: I feel we have a race condition here without a lock?

Also call this one `timeoutTimer` instead of `timer`.

pkg/backend/remote/remote.go
```diff
@@ -20,8 +21,10 @@ import (
 )
 
 var (
+	ErrPingTimeout = errors.New("Ping timeout")
+
 	pingInveral   = 2 * time.Second
-	timeout       = 30 * time.Second
+	pingTimeout   = 2 * time.Second
```
yasker, line 27: 2 seconds is quite a small ping timeout. I am not sure if we should set a timeout so small in the data connection layer.

pkg/backend/remote/remote.go
```diff
@@ -302,6 +305,24 @@ func (r *Remote) monitorPing(client *dataconn.Client) {
 	}
 }
 
+func (r *Remote) pingWithTimeout(client *dataconn.Client) error {
+	timeout := time.After(pingTimeout)
+	err := ErrPingTimeout
+
+	status := make(chan error, 1)
```
yasker, line 312: Allocating a channel every 2 seconds? It is hard work for GC.

pkg/dataconn/client.go
```diff
@@ -217,10 +217,26 @@ func (c *Client) handleResponse(resp *Message) {
...
+			// We received a response to an I/O operation
+			// Stop the timer.
+			c.timer.Stop()
+			c.iops--
+
+			// If there are other reads and writes in
+			// progress, wait opTimeout seconds for any
+			// response to an I/O operation.
+			if c.iops > 0 {
+				c.timer = time.NewTimer(opTimeout)
```
yasker, line 231: Can we reset the timer instead of recreating one? Saving GC efforts.

pkg/dataconn/client.go
```diff
@@ -152,10 +115,17 @@ func (c *Client) Close() {
 }
 
 func (c *Client) loop() {
+	c.timer = time.NewTimer(opTimeout)
+	if !c.timer.Stop() {
+		<-c.timer.C
```
yasker, line 120: why are we waiting for the opTimeout here?

I guess it's because of the following `c.timer.C`? But I feel there should be a better way.

Also, waiting for an expired timer seems bad.

pkg/dataconn/client.go
```diff
@@ -152,10 +115,17 @@ func (c *Client) Close() {
...
+	}
+
 	defer close(c.send)
 
 	for {
 		select {
+		case <-c.timer.C:
```
yasker, line 127: In fact I am not sure when `select` will refresh the conditions if `c.timer` has changed.

pkg/dataconn/client.go
```diff
@@ -217,10 +217,26 @@ func (c *Client) handleResponse(resp *Message) {
...
+			if c.iops > 0 {
+				c.timer = time.NewTimer(opTimeout)
+			}
+		}
+
 		req.Type = resp.Type
 		req.Size = resp.Size
 		req.Data = resp.Data
 		req.Complete <- struct{}{}
+
```
yasker, line 239: Unnecessary change.

yasker (changes requested):

pkg/dataconn/client.go
keithalucas: c.timer is only changed by one of the function called via the select.   select waits for one of the channels then it executes the function that can change c.timer.  c.timer isn't changed while the select is executing and waiting for the channels.   The channels that select is waiting for won't channel during the select.  The next iteration of select may be using different channels because c.timer was updated.

pkg/dataconn/client.go
keithalucas: The Timer API doesn't allow the creation of a timer without a timeout value set.  I made the Timer and immediately stopped it so the channel that the timer uses would exist.   Another approach could be to create the timer with the maximum Duration.

pkg/dataconn/client.go
keithalucas: I changed it to reset.

pkg/backend/remote/remote.go
keithalucas: I'll change it to eight seconds which is what it previously was in dataconn.Client.

pkg/dataconn/client.go
keithalucas: I renamed it.  The timer is only used by the loop goroutine.

pkg/backend/remote/remote.go
keithalucas: The time.After was making a channel too.  The only thing I could think of it reusing the ticker for pings for determining a ping timeout which doesn't create new channels every iteration.

pkg/backend/remote/remote.go
```diff
@@ -314,15 +314,14 @@ func (r *Remote) monitorPing(client *dataconn.Client) {
 			}
 
 		case err := <-pingStatus:
+			inProgress = false
```
yasker, line 317: Should be in the previous commit.

yasker (changes requested):
LGTM other than a minor comment.

@joshimoo can you take a look as well?

pkg/backend/remote/remote.go
keithalucas: I moved it to the correct commit.

yasker:
LGTM. I'd like to have @joshimoo take a look as well since this is the critical data path.

joshimoo (commented):
As discussed on Slack, I am not a big fan of the timer and would prefer a simpler deadline approach. I implemented the deadline approach on top of your changes below:
PR: https://github.com/longhorn/longhorn-engine/pull/633
https://github.com/longhorn/longhorn-engine/pull/633/commits/848a6526cc26b2b0a8d460eaed893eb9cc8dab28

I also moved the client error handling / timeout mechanism into the loop function.
so that request/response only deals with valid messages and io tracking is done as part of the loop function.

The early client errors made it hard to make sure that the io timeout is correctly implemented, so I felt it would be better to have the error handling together with the timeout handling. And to simplify the message processing.

joshimoo:
@keithalucas can you have a look at my timeout handling changes here https://github.com/longhorn/longhorn-engine/pull/633
https://github.com/longhorn/longhorn-engine/pull/633/commits/848a6526cc26b2b0a8d460eaed893eb9cc8dab28
Any issues with that implementation?
if you prefer the timer approach, let me know I find it's easier to reason about the correctness of the implementation by reducing the global state, since now all the relevant code is contained right next to each other.

joshimoo (dismissed):
LGTM

can you squash all the commits down to 3:
- update liblonghorn
- change io timeout handling
- change ping timeout handling

If it's to much of a hassle, you can also squash down to 2
- update liblonghorn
- change timeout handling

keithalucas:
> LGTM
> 
> can you squash all the commits down to 3:
> 
> * update liblonghorn
> * change io timeout handling
> * change ping timeout handling
> 
> If it's to much of a hassle, you can also squash down to 2
> 
> * update liblonghorn
> * change timeout handling

I squashed it down to two.

yasker (approved):
LGTM.

yasker:
On a side note, trying to keep commits separate and clearly self-contained is a good practice. It makes it easier to review and revisit even after the commits are merged. I would definitely prefer 3 commits in this case, but since it's a separate file for ping, 2 works.

yasker:
I've also added the related Longhorn issue to the PR's comment.

The issue needs to be added to the PR's comment to link the PR to the issue. But don't add the link directly in the commits' comment, since it can cause repeated messages to show up in the issue.

joshimoo (approved):
LGTM

joshimoo:
Leaving a reference to the related liblonghorn changes here:
https://github.com/rancher/liblonghorn/pull/10

PR #631, opened 2021-06-28
Backing image enhancement

shuo-wu:
longhorn/longhorn#2530


joshimoo (commented):
Replacing the Backing Image URL with a checksum seems fine.

shuo-wu:
> Replacing the Backing Image URL with a checksum seems fine.

Then we may need to bump up the min API version, which is unexpected.

joshimoo:
@shuo-wu I might have missed it but it looked like this was the only change that was done here? i.e. replace backing image url with backing image checksum?

shuo-wu:
`--backing-image-name` indicates the volume that creates the backup is using a backing image. Later if users want to restore a volume from this backup volume, this backing image will be set for the restore volume by default.
`--backing-image-checksum` can be used to verify if the backing image matching before backup creation or restoring.
`--backing-image-url` is for recording only. It won't be used to create a backing image during restoring even if the backing image doesn't exist.

jenting (commented):
CI failed at `pkg/sync/rpc/server.go:652:30: undefined: backupstore.GetBackupFromBackupURL`

shuo-wu:
> GetBackupFromBackupURL

Thank you for the notification. Updated the PR.

jenting:
`app/cmd/backup.go:31:4: undefined: cmd.BackupInspectCmd`

Seems like it highly depends on #623 :(

jenting:
Integration test still checks `VolumeBackingImageURL`

shuo-wu:
Fixed the integration test failures.

pkg/meta/version.go
```diff
@@ -2,7 +2,7 @@ package meta
 
 const (
 	// CLIAPIVersion used to communicate with user e.g. longhorn-manager
-	CLIAPIVersion    = 4
+	CLIAPIVersion    = 5
```
jenting, line 5: Could you please elaborate more on when to bump this version?

integration/common/core.py
```diff
@@ -431,16 +431,14 @@ def reset_volume(grpc_c, *grpc_r_list):
...
+                  backing_image_name="", backing_image_checksum=""):
     backup = cmd.backup_create(url, snap, backup_target,
-                               [], backing_image_name, backing_image_url)
+                               [], backing_image_name, backing_image_checksum)
     backup_info = cmd.backup_inspect(url, backup)
     assert backup_info["URL"] == backup
     assert backup_info["VolumeSize"] == volume_size
     if backing_image_name != "":
         assert backup_info["VolumeBackingImageName"] == backing_image_name
-    if backing_image_url != "":
```
jenting, line 442: Do we need to check the checksum if it's not empty?

jenting (commented):
LGTM in general.

pkg/meta/version.go
shuo-wu: if the engine binary command definition used by longhorn-manager changes, we need to bump up CLIAPIVersion at least. If we don't allow the longhorn manager to use the old binary to talk with the old engines, we need to bump up CLIMinAPIVersion as well. Here, `don't allow` means there is no compatible code in the longhorn manager part: e,g, https://github.com/shuo-wu/longhorn-manager/blob/4368a4ac6521b267f756276a41e23e75b47f24a6/engineapi/backups.go#L260-L264

Ideally, CLIMinAPIVersion should be increased as long as the new longhorn manager cannot use the old binary without the compatible code. But here I don't want to block the live upgrade. The above ugly compatible code is introduced to avoid CLIMinAPIVersion change.

integration/common/core.py
shuo-wu: Now the checksum it's store in BackupVolume info only. The backup info doesn't know anything about the backing image checksum.

pkg/meta/version.go
jenting: Thanks for the info.

jenting (dismissed):

jenting:
There're some conflicting files.

shuo-wu:
Conflicts resolved

innobead (approved):
LGTM

@shuo-wu Some conflicting code from master due to the recently merged change. Need to resolve before merge.

innobead (approved):

jenting (approved):

PR #632, opened 2021-06-28
Stop monitoring goroutine when removing replica

keithalucas:
The monitoring goroutine was continuing to run after the replica was
removed.  This caused a problem when a replica is recreated for the same
IP address and port after one is deleted.  When a replica is removed,
the channel that the monitoring goroutine is waiting for will eventually
get an error message.  In the normal case, the SetReplicaMode won't find
a replica with the address to mark as ERR.  However, when another
replica is with the same address is added after the original was
removed, SetReplicaMode for the old monitorChan will mark the new
replica as ERR with SetReplicaMode.

This change makes removing the replica also stop the monitoring
goroutine that was launched for that replica.  It puts a nil error on
the replica's monitor channel so monitoring goroutine  will exit as part
of removing the replica.

longhorn/longhorn#1628

### DIFF
--- a/pkg/controller/replicator.go
+++ b/pkg/controller/replicator.go
@@ -79,6 +79,9 @@ func (r *replicator) RemoveBackend(address string) {
 	// We cannot wait for it's return because peer may not exists anymore
 	// The backend may be nil if the mode is ERR
 	if backend.backend != nil {
+		// Stop the monitoring goroutine in the Controller
+		backend.backend.StopMonitoring()
+
 		go backend.backend.Close()
 	}
 	delete(r.backends, address)


pkg/controller/replicator.go
```diff
@@ -76,6 +76,10 @@ func (r *replicator) RemoveBackend(address string) {
 
 	logrus.Infof("Removing backend: %s", address)
 
+	// Stop the monitoring goroutine in the Controller
+	monitorChan := backend.backend.GetMonitorChannel()
```
joshimoo, line 80: The comment below mentions that backend can be nil, then this would throw a nil pointer.
Did you investigate whether it's safe to rely on this not being nil?

joshimoo (changes requested):
the below comment mentions that backend can be nil, is this a concern?

pkg/controller/replicator.go
```diff
@@ -76,6 +76,10 @@ func (r *replicator) RemoveBackend(address string) {
...
+	monitorChan <- nil
```
joshimoo, line 81: Instead of directly writing against the monitoring channel we should instead consider calling.
backend.backend.StopMonitoring()

Since the above, will lead to a deadlock, for file based backends which return a nil channel.

joshimoo (changes requested):

pkg/controller/replicator.go
keithalucas: I moved it into the if.

pkg/controller/replicator.go
keithalucas: I changed it to StopMonitoring.

joshimoo (approved):
LGTM

I think stopping monitoring is a good change I am not sure if we are covering all cases. Since I am not sure when backend.backend would be nil and if that case also handles monitoring cancellation correctly.

But this is already an improvement, so we could merge it already.
As quickly discussed during the call one refactoring that we could do is replace the closeChannel with a context.WithCancel then just call the cancel function during stop monitoring/close.

joshimoo:
@keithalucas I approved it, if you feel comfortable with the implementation you can use the `rebase and merge` button or let me know and I will do it instead :) I saw there was some issue with reproduction during the team meeting, just double checking if this impacts this pr?

joshimoo:
Only maintainers can merge even after approval, so I will merge it. Thanks for reminding me @keithalucas

PR #634, opened 2021-07-02
Add test for removing a replica and recreating it

keithalucas:

### DIFF
--- a/integration/core/test_cli.py
+++ b/integration/core/test_cli.py
@@ -6,6 +6,7 @@ import json
 import datetime
 import pytest
 import tempfile
+import grpc
 
 from urllib.parse import urlparse
 
@@ -28,6 +29,7 @@ from common.core import (  # NOQA
     reset_volume,
     get_replica_client_with_delay,
     get_controller_version_detail,
+    verify_replica_mode,
 )
 
 import common.cmd as cmd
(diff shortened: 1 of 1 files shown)


integration/core/test_cli.py
```diff
@@ -1397,3 +1400,50 @@ def test_engine_restart_after_sigkill(bin):  # NOQA
...
+    r2_url = grpc_replica_client2.url
+    v = grpc_controller_client.volume_start(replicas=[
+        r1_url, r2_url,
+    ])
+    assert v.replicaCount == 2
+
+    # Remove the first replica and recreate it immediately.
+    grpc_controller_client.replica_delete(r1_url)
+
+    for i in range(RETRY_COUNTS):
```
shuo-wu, line 1437: Add wait interval for this retry loop. Then the test doesn't need to retry for 100 times.

shuo-wu (changes requested):
It's better to add the test skeleton before implementation.

integration/core/test_cli.py
keithalucas: I added the sleep and changed it to use RETRY_COUNTS_SHORT.

keithalucas:
> It's better to add the test skeleton before implementation.

Added the sketelon.

shuo-wu (approved):

joshimoo (approved):
LGTM

PR #635, opened 2021-07-02
Add timeout test.

keithalucas:
https://github.com/longhorn/longhorn/issues/2206


yasker:
We will close this PR since the timeout test in the integration test doesn't make much sense by itself. We will depend on the manual test for it.

PR #637, opened 2021-07-09
Make closing all the replicas stop all the replica monitors

keithalucas:
The replica monitor channel goroutine is still running when a
controller's volume is shutdown via the GRPC VolumeShutdown method.
This change makes stops the Controller's replica monitor goroutine when
the volume is shutdown.

### DIFF
--- a/pkg/controller/replicator.go
+++ b/pkg/controller/replicator.go
@@ -292,6 +292,9 @@ func (r *replicator) Close() error {
 		if backend.mode == types.ERR {
 			continue
 		}
+
+		backend.backend.StopMonitoring()
+
 		if err := backend.backend.Close(); err != nil {
 			lastErr = err
 		}


yasker:
Hmm, this got me thinking, can we detect the "leaked" goroutine in a systematic way in the future? No fd leak, but goroutine leak.

pkg/controller/replicator.go
```diff
@@ -292,6 +292,10 @@ func (r *replicator) Close() error {
 		if backend.mode == types.ERR {
 			continue
 		}
+
+		// Stop the monitoring goroutine in the Controller
```
yasker, line 296: The code is pretty obvious by itself, so don't need the comment.

yasker (changes requested):

pkg/controller/replicator.go
keithalucas: Removed.

yasker (approved):

yasker:
Unrelated to this fix: I found there is another place removing monitoring thread for backend, but for the `close` call, it's done in a goroutine instead of blocking, with a comment
```
// We cannot wait for it's return because peer may not exists anymore
// The backend may be nil if the mode is ERR
```

https://github.com/longhorn/longhorn-engine/blob/7f42612eb488b2b829a18711d605323f7a77c50a/pkg/controller/replicator.go#L83

I am thinking can the `close` in this PR being blocked? Though it seems we haven't seen the real-world example of it yet.

keithalucas:
> Unrelated to this fix: I found there is another place removing the monitoring thread for backend, but for the `close()` call, it's done in a goroutine instead of blocking, with a comment
> 
> ```
> // We cannot wait for it's return because peer may not exists anymore
> // The backend may be nil if the mode is ERR
> ```
> 
> https://github.com/longhorn/longhorn-engine/blob/7f42612eb488b2b829a18711d605323f7a77c50a/pkg/controller/replicator.go#L85
> 
> I am thinking can the `close` in this PR being blocked? Though it seems we haven't seen the real-world example of it yet.

The remote backend's Close method tries to communicate with the replica process over GRPC.  https://github.com/longhorn/longhorn-engine/blob/fd7c58ca7bc5c2e52ddb248d7cd8551a304ec4eb/pkg/backend/remote/remote.go#L51.  If somehow there was a network disconnection, I thought it could take some time for the GRPC call to timeout.

keithalucas:
> Hmm, outside of the topic: this got me thinking, can we detect the "leaked" goroutine in a systematic way in the future? No fd leak, but goroutine leak.

The problem is this case was mostly due to the Controller object only having a list of addresses of the replica not the object that holds the replica and the Controller object might.  I originally want to make whoever "owns" (the Controller in this case) the goroutine to have a channel or context to stop it when it is no longer interested in it (when it deleted the correspond address string from the list of replica addresses), but instead elected to using the existing channel to stop it.

PR #640, opened 2021-07-21
Implementing VolumeExport() gRPC in replica SyncAgentServer

PhanLe1010:
When called, VolumeExport() gRPC export volume data at the input snapshot to the receiver on the remote host

longhorn/longhorn#2403
longhorn/longhorn#2534
longhorn/longhorn#1815


pkg/replica/backup.go
```diff
@@ -2,11 +2,14 @@ package replica
 
 import (
 	"fmt"
+
```
shuo-wu, line 5: unnecessary empty line

pkg/replica/replica.go
```diff
@@ -1156,3 +1177,72 @@ func (r *Replica) GetReplicaStat() (int64, int64) {
...
+			defer close(fileIntervalChannel)
+			defer close(errChannel)
+
+			// The replica consists of data and and hole region.
+			// Inside this function scope, let represent 0 as hole sector and 1 as data sector.
+			// Then a replica could look like 010001111...
+
+			getSectorType := func(diskIndex byte) int {
+				if diskIndex >= byte(baseDiskIndex) {
+					return 1
```
shuo-wu, line 1204: Use const to represent data or holes. e.g.:
```
const (
    sectorTypeData = 1  // The value is consist with sparse.SparseData
    sectorTypeHole = 2
)
```

Or you may directly use `sparse.SparseHole` or `sparse.SparseData`

pkg/replica/replica.go
```diff
@@ -1156,3 +1177,72 @@ func (r *Replica) GetReplicaStat() (int64, int64) {
...
+		return nil, err
+	}
+	// baseDiskIndex is the smallest index in r.volume.files that is needed to export
+	// Note that index 0 in r.volume.files is nil. Index 1 is backing file if the volume has backing file.
+	// Otherwise, index 1 is regular disk file.
+	baseDiskIndex := 1
+
+	getFileInterval := func(ctx context.Context) (<-chan sparse.FileInterval, <-chan error, error) {
+		fileIntervalChannel := make(chan sparse.FileInterval)
+		errChannel := make(chan error)
```
shuo-wu, line 1192: Can we use buffered channels?

pkg/replica/replica.go
```diff
@@ -1156,3 +1177,72 @@ func (r *Replica) GetReplicaStat() (int64, int64) {
 
 	return lastModifyTime, headFileSize
 }
+
+func (r *Replica) GetFileInterval(includeBackingFileLayer bool) (sparse.GetFileInterval, error) {
+	if err := preload(&r.volume, includeBackingFileLayer); err != nil {
+		return nil, err
+	}
```
shuo-wu, line 1184: I am thinking if we can avoid using `preload` here. Because the extent retrieval for all snapshots files in `preload` is duplicate with the following function `getFileInterval`.

Maybe you can:
    1. Move the main part of this function into `diff_disk.go`
    2. Directly iterate all snapshot files to get the replica fiemap. This may be similar to `diffDisk.lookup`

pkg/replica/replica.go
```diff
@@ -1156,3 +1177,72 @@ func (r *Replica) GetReplicaStat() (int64, int64) {
...
+				}
+				return sparse.SparseData
+			}
+
+			lastSectorType := 0
+			lastOffset := int64(0)
+			for sectorIndex, diskIndex := range r.volume.location {
+				offset := int64(sectorIndex) * r.volume.sectorSize
+				currentSectorType := getSectorType(diskIndex)
+				if currentSectorType != lastSectorType {
```
shuo-wu, line 1220: This seems incorrect. It will miss the last interval since the update of `lastOffset` is in `if currentSectorType != lastSectorType`

shuo-wu (changes requested):

pkg/replica/replica.go
PhanLe1010: The last interval will be handle outside of this loop in the `if lastOffset < r.volume.size {` statement below. The if statement also take care of conner case where the whole replica is a hole or a data region.

pkg/replica/replica.go
PhanLe1010: I think preload is needed for the function `getFileInterval` to work correctly. Let's have a sync call about this.

pkg/replica/replica.go
PhanLe1010: Why do we need to use buffered channels?
Is it because we have 4 worker gorotines that are taking the work from this channels?

pkg/replica/replica.go
PhanLe1010: Thanks! I added the const declaration as suggested. But I still prefer 0 as hole and 1 as data. This shouldn't affect anything else because the convention is only within the scope of this function.

pkg/replica/backup.go
PhanLe1010: Thank you. Fixed

pkg/replica/replica.go
shuo-wu: Sure. The value is not required to be the consist with `sparse.SparseData` or `sparse.SparseHole`

pkg/replica/replica.go
shuo-wu: With a buffer, the goroutine can load the subsequential intervals when the caller is sending one interval. It makes the producing and consuming parallel.

PhanLe1010: Thank you. Fixed

pkg/replica/replica.go
PhanLe1010: Discussed with Shuo

pkg/replica/replica.go
PhanLe1010: Discussed with Shuo

Dockerfile.dapper
```diff
@@ -52,6 +52,7 @@ RUN cd /usr/src/libqcow-20181117 && \
 RUN cd /usr/src/libqcow-20181117 && \
     make -j$(nproc) && \
     make install
+RUN ldconfig
```
joshimoo, line 55: Not sure why this is now necessary?
Since it worked previously, but I saw the thread on slack

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/longhorn/longhorn-engine
 
 go 1.13
 
+replace github.com/longhorn/sparse-tools => /Users/peterle/go/src/github.com/longhorn/sparse-tools
```
joshimoo, line 5: I merge the sparse tools PR https://github.com/longhorn/sparse-tools/pull/77 thanks

pkg/replica/client/client.go
```diff
@@ -407,6 +407,23 @@ func (c *ReplicaClient) SendFile(from, host string, port int32) error {
 	return nil
 }
 
+func (c *ReplicaClient) ExportVolume(snapshotName, host string, port int32, exportBackingImageIfExist bool) error {
+	syncAgentServiceClient := c.getSyncServiceClient()
```
joshimoo, line 411: Just leaving a note here:
I will add an error return to getSyncServiceClient() since I need to lazy init (optimization) the client due to the way the longhorn-manager uses the engine-binary for monitoring at the moment.

joshimoo (dismissed):
LGTM had a review session with @PhanLe1010 we discussed that the volume export would end up exporting all the zero bytes, this is something we can optimize in the future to turn the zero sectors into holes.

But since we are also giving the user the ability to convert the volume into a QCOW file for use as a backing image, we decided this is not a high priority for the moment.

Dockerfile.dapper
PhanLe1010: Previously we didn't load the libqcow library in the unit test but now we do import the `pkg/backingfile` in the replica unit test [here](https://github.com/longhorn/longhorn-engine/blob/fd02e4955e965a1d8e1c0d1155b25816e5d3ce18/pkg/replica/backup_test.go#L8). This in turn import the `libqcow` package which inturn includes the `#include <libqcow.h>`. So, I guess we need to run this command so that unit test can find the header `libqcow.h`

go.mod
PhanLe1010: Thanks

PhanLe1010:
>LGTM had a review session with @PhanLe1010 we discussed that the volume export would end up exporting all the zero bytes, this is something we can optimize in the future to turn the zero sectors into holes.

Note that this only happens if the user wants to export both data layer and backing file layer (e.g., creating/exporting a new backing image from a volume that is using a backing image). If the user doesn't want to export the backing file layer (e.g., volume clone) or the backing file layer doesn't exist, we don't export zero bytes

pkg/replica/client/client.go
joshimoo: Once this PR https://github.com/longhorn/longhorn-engine/pull/646 is merged, getSyncServiceClient() will return (client, err) tuple.

pkg/replica/backup.go
```diff
@@ -301,7 +303,8 @@ func (rb *BackupStatus) CompareSnapshot(snapID, compareSnapID, volumeID string)
 		Offset: -1,
 	}
 
-	if err := preload(&rb.replica.volume); err != nil {
+	rb.replica.volume.initializeSectorLocation(nilFileIndex)
+	if err := rb.replica.volume.preload(); err != nil {
```
shuo-wu, line 307: `if err := rb.replica.Preload(); err != nil {`

app/cmd/volume_export.go
```diff
@@ -0,0 +1,106 @@
+package cmd
+
+import (
```
shuo-wu, line 3: Please rearrange imports

shuo-wu (changes requested):
In general LGTM

innobead (approved):
In general LGTM.

pkg/replica/backup.go
PhanLe1010: Thank you. Fixed

app/cmd/volume_export.go
PhanLe1010: Thank you. Fixex

innobead (approved):

PR #644, opened 2021-07-23
2818-bulk-backup-status-retrieval

joshimoo:
This enables GRPC connection reuse for the Controller/Replica clients

We also initialize the clients as part of the Constructors to deal with incorrect addresses for the example
I retained the prior behavior of the background lazy connection, so connection establishment errors would still
be reported as part of the first grpc call.

With this code structure we have the ability to add retry loops for transient errors, i.e. subconnection in failures.
I will create an EPIC for grpc communication improvemets so we can track this further.

I only implement the Task change for  the BackupStatus retrieval, the others can be just as easily implemented after this PR.
We should consider the Task struct as deprecated and over time turn the instance methods into free functions.
Then the caller can pass the appropriatly required clients.

longhorn/longhorn#2818

### DIFF
--- a/app/cmd/add_replica.go
+++ b/app/cmd/add_replica.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -36,7 +37,12 @@ func addReplica(c *cli.Context) error {
 	replica := c.Args()[0]
 
 	url := c.GlobalString("url")
-	task := sync.NewTask(url)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	task, err := sync.NewTask(ctx, url)
+	if err != nil {
+		return err
+	}
 
 	if c.Bool("restore") {
 		return task.AddRestoreReplica(replica)
@@ -63,7 +69,13 @@ func startWithReplicas(c *cli.Context) error {
 	replicas := c.Args()
 
 	url := c.GlobalString("url")
-	task := sync.NewTask(url)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	task, err := sync.NewTask(ctx, url)
+	if err != nil {
+		return err
+	}
+
 	return task.StartWithReplicas(replicas)
 }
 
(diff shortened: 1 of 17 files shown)


joshimoo:
Can be reviewed already but /hold merge till I had a chance to test on cluster :)
E2E for engine passed a couple of times, will do some E2E with manager.
And do some testing on cluster tomorrow.

shuo-wu (commented):
There are some questions that need to be clarified:
1. Will each engine monitor goroutine in the manager reuse a task client when this PR is in? Otherwise, there will be new connections between controllers and replicas for each backup status retrieval call. The last commit of this PR is just avoiding duplicate connections establishing in one backup status call.
2. Can we refactor [this part](https://github.com/longhorn/longhorn-engine/blob/23ee52cf830277a4e13a3c00e4167952cd9cb7c5/pkg/backend/remote/remote.go#L43) as well? If YES, then there won't be duplicate new connections in [controller::backend](https://github.com/longhorn/longhorn-engine/blob/23ee52cf830277a4e13a3c00e4167952cd9cb7c5/pkg/controller/control.go#L29)::[backends](https://github.com/longhorn/longhorn-engine/blob/23ee52cf830277a4e13a3c00e4167952cd9cb7c5/pkg/controller/replicator.go#L22).

joshimoo:
> There are some questions that need to be clarified:
> 
> 1. Will each engine monitor goroutine in the manager reuse a task client when this PR is in?

This PR lays the ground work and acts as an example for better grpc connection usage.
This PR also turns the per volume backup status retrieval from O(n) connections into O(1) connections, which is a huge improvement. 
Example: 50 backups on a volume: 50+ connections -> 1+3 connections ideally.


> 2. Can we refactor [this part](https://github.com/longhorn/longhorn-engine/blob/23ee52cf830277a4e13a3c00e4167952cd9cb7c5/pkg/backend/remote/remote.go#L43) as well?

Sure can be done in the future,

Further improvements on the cross component communication handling are/should be tracked in this EPIC #https://github.com/longhorn/longhorn/issues/2821

shuo-wu (approved):

PR #645, opened 2021-07-29
1628 backport closing monitoring channel

keithalucas:

### DIFF
--- a/pkg/controller/replicator.go
+++ b/pkg/controller/replicator.go
@@ -289,6 +289,9 @@ func (r *replicator) Close() error {
 		if backend.mode == types.ERR {
 			continue
 		}
+
+		backend.backend.StopMonitoring()
+
 		if err := backend.backend.Close(); err != nil {
 			lastErr = err
 		}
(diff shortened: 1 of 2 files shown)


pkg/controller/replicator.go
```diff
@@ -289,6 +292,9 @@ func (r *replicator) Close() error {
 		if backend.mode == types.ERR {
 			continue
 		}
+
+		backend.backend.StopMonitoring()
+
 		if err := backend.backend.Close(); err != nil {
 			lastErr = err
 		}
```
joshimoo, line 300: Do we need a nil check (backend.backend != nil) in this area?
Or is it guaranteed that the internal backend is always non nil?
Since we do a nil check in the RemoveBackend case.

keithalucas: It might just be overly defensive programming.  There are several other places where we loop over the backends map and don't check if each backend in the map is not nil, (`Snapshot`, `RemainSnapshots`, `buildReadWriters`, `Expand`).   I don't think it's necessarily a problem.  If it is, we could make a separately issue/pr to check that it's not nil everywhere.

joshimoo: Agree I don't think we have seen a code panic in this area before, but if we do we can open a new issue and investigate how to make this backend.backend business a little nicer :)

joshimoo (approved):
LGTM

PR #646, opened 2021-07-30
2818 further socket usage optimizations

joshimoo:
Lazily initialize the replica clients for replica/sync service

This is to optimize the connection count for the longhorn-manager engine
monitor loop, in a single loop the monitor executes the engine binary 8
times this leads to a total connection count of `f() = 8e + 6 * 2r = 20`
for a volume with 1 replica.

Since all the calls the monitor makes only require either the replica or
the sync client, we can reduce this further with this optimization to
`g() = 8e + 6r = 14` per monitor loop.

The monitor loop executes `60/5 = 12` times per minute, so our total 
connections per minute are reduced from **240** to **168**. For volume with 3 
replicas we end up with **312** connection per minute.

No further optimization on this end is possible, the next required task
is removal of the direct engine binary invocations by the longhorn-manager.
This will reduce the connection counts to `h() = 1e + 2*r` which as one
can see would lead to the desired behavior for each volume of 1 engine connection 
and 2 connections per replica (replica / sync).

longhorn/longhorn#2818

Signed-off-by: Joshua Moody <joshua.moody@suse.com>

### DIFF
--- a/app/cmd/backup.go
+++ b/app/cmd/backup.go
@@ -77,23 +77,13 @@ func BackupStatusCmd() cli.Command {
 	}
 }
 
-func getReplicaModeMap(c *cli.Context) (map[string]types.Mode, error) {
-	controllerClient, err := getControllerClient(c)
-	if err != nil {
-		return nil, err
-	}
-
-	replicas, err := controllerClient.ReplicaList()
-	if err != nil {
-		return nil, fmt.Errorf("failed to get replica list: %v", err)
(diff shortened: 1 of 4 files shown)


joshimoo:
Good to review, let me merge afterwards.

shuo-wu (approved):
LGTM

innobead (approved):
LGTM.

This is good for better resource usage.

PR #648, opened 2021-08-03
Implement SnapshotCloneCmd() and SnapshotCloneStatusCmd() CLIs

PhanLe1010:
Longhorn manager will call `SnapshotCloneCmd()` Cli to clone a snapshot from source volume to a replica on target volume. 
The `SnapshotCloneStatusCmd()` Cli is used to fetch the cloning status.

longhorn/longhorn#1815


PhanLe1010:
Sorry. This PR need more testing. I should have opened a draft instead.

pkg/sync/sync.go
```diff
@@ -767,3 +776,106 @@ func (t *Task) RebuildStatus() (map[string]*ReplicaRebuildStatus, error) {
...
+	for _, r := range replicas {
+		go func(r *types.ControllerReplicaInfo) {
+			defer wg.Done()
+			repClient, err := replicaClient.NewReplicaClient(r.Address)
+			if err != nil {
+				syncErrorMap.Store(r.Address, err)
+				return
+			}
+			defer repClient.Close()
+			if err := repClient.CloneSnapshot(sourceReplica.Address, snapshotFileName, exportBackingImageIfExist); err != nil {
```
innobead, line 821: This means when multiple replicas of the dest volume will only clone from the same source replica of the source volume. 

My 2 cents, in the future, probably think about how to share load to better leverage available replicas to clone instead of only one source replica.

innobead (dismissed):
LGTM.

pkg/sync/sync.go
PhanLe1010: @innobead  Thanks for the advice! 

Currently we only allow the target volume to have exactly 1 replica while cloning is happening. After that, we scale up the replica. So, the multiple cloning is not an issue here. 

However, in the future if we want to speed up process, then yeah we can follow your advice.

app/cmd/snapshot.go
```diff
@@ -326,3 +365,52 @@ func infoSnapshot(c *cli.Context) error {
...
+	}
+	exportBackingImageIfExist := c.Bool("export-backing-image-if-exist")
+
+	controllerClient, err := getControllerClient(c)
+	if err != nil {
+		return err
+	}
+	defer controllerClient.Close()
+
+	fromControllerClient, err := client.NewControllerClient(fromControllerAddress)
```
joshimoo, line 386: looks like this client is leaked?

You should be able to close it in a defer here, since the following code seems to be blocking `defer fromControllerClient.Close()`

joshimoo (commented):
Generally LGTM from a quick review

pkg/sync/sync.go
innobead: Thank for the explanation :+1:

innobead:
@PhanLe1010 let's merge it after resolving the @joshimoo 's comment.

app/cmd/snapshot.go
PhanLe1010: Thanks Joshua for the catch! Yeah, I missed this one. Fixed

innobead (approved):

PR #650, opened 2021-08-09
Bump up the CLI API version to 6

PhanLe1010:
Bump up the CLI API version because of adding new `SnapshotCloneCmd()` and `SnapshotCloneStatusCmd()` CLIs

longhorn/longhorn#1815

### DIFF
--- a/pkg/meta/version.go
+++ b/pkg/meta/version.go
@@ -2,7 +2,7 @@ package meta
 
 const (
 	// CLIAPIVersion used to communicate with user e.g. longhorn-manager
-	CLIAPIVersion    = 5
+	CLIAPIVersion    = 6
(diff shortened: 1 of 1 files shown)


PhanLe1010:
@shuo-wu 

Can I have a question. The last time we bump the cli api version to 5. It is only on master-head and doesn't get into v1.1.2 yet, right? If so, I think we don't need this PR.

jenting:
I remember yes.

innobead:
ref: https://github.com/longhorn/longhorn-engine/commit/0b0c8202075f29d93d94c1252ab77833c9fb873e#diff-41f3352cafc153a83763338fd3a03b513a9d04eb1ad077395c66ce96ca057e20

PR #651, opened 2021-08-11
Revert "Support new backup list behavior and backup inspect-volume co…

innobead:
This reverts commit 192e2a6922bfec7ca54ce8de67dc193cbec07c92.

Need to wait for the final update of https://github.com/longhorn/longhorn-manager/pull/952 before introducing this change. Revert this commit first.


shuo-wu (approved):

innobead:
The regression of https://github.com/longhorn/longhorn-manager/pull/952 passed, so we don't need this revert PR anymore.

PR #652, opened 2021-08-16
Remove ping timeout

keithalucas:
The ping timeout was occurring in scenarios with limited bandwidth and
high network usage.  The limited bandwidth and high network traffic was
delaying the transmission of the ping message and the response, causing
the timeout to occur when the replica was in fact operational.

Without the ping timeout, the code can still take advantage of the
default TCP keep alive timeout that golang's networking code sets to 15
seconds.  The TCP keep alive timeout is a socket option whose
functionality is implemented by the Linux kernel.

https://github.com/longhorn/longhorn/issues/2882

### DIFF
--- a/pkg/backend/remote/remote.go
+++ b/pkg/backend/remote/remote.go
@@ -1,7 +1,6 @@
 package remote
 
 import (
-	"errors"
 	"fmt"
 	"net"
 	"strconv"
@@ -20,13 +19,8 @@ import (
 	"github.com/longhorn/longhorn-engine/proto/ptypes"
 )
 
-var (
-	ErrPingTimeout = errors.New("ping timeout")
-)
-
 const (
 	PingInterval = 2 * time.Second
-	PingTimeout  = 8 * time.Second
 )
 
 func New() types.BackendFactory {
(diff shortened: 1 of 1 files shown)


yasker (changes requested):
1. We need an issue associate with the PR.
2. Do we need to remove `ping` from the protocol as well? 

Also, somehow CI failed.

keithalucas:
> 1. We need an issue associate with the PR.
> 2. Do we need to remove `ping` from the protocol as well?
> 
> Also, somehow CI failed.

I associated the PR with github.com/longhorn/longhorn/issue/2882.  I felt like I should have removed ping from the protocol  because it would be confusing if it's there and not used so I removed it.  https://github.com/longhorn/sparse-tools/blob/a0fb4226a960dfa0a77a843b052a44824a9d729a/stats/stats.go#L87 sparse-tools still references ping which could do in another pull request.  I believe those python tests indirectly rely on a ping timeout occurring.  I need to investigate and fix them.

yasker:
Yes, let's address the sparse file in another PR. That one seems hardcoded in its own repo so I guess there is no dependency.

yasker (dismissed):
LGTM. Pending the CI result.

keithalucas:
Upon investigation, in the tests that fail, the ping is necessary because the controller uses it to get the status of the replica when the replica is closed or failed.  I added the ping back but the ping doesn't timeout.  This allows the controller to see that a replica has been closed and set the replica's mode to ERR.  In the future, it might make more sense to have the replica disconnect when it is closed or send a message indicating a change in status.

yasker:
CI was still failing on ARM part, I just restarted it.

innobead:
Failed again restarted to see the result

innobead (approved):

PR #659, opened 2021-09-08
Bump longhorn/backupstore version

jenting:
#### Proposal Change

Set NFS mount option actimeo to 1 second

#### Issue

https://github.com/longhorn/longhorn/issues/2926


innobead (approved):
LGTM

@jenting need to backport to 1.2.x.

PR #662, opened 2021-10-07
Replace fsfreeze with prior behavior of system sync due to race condition

joshimoo:
there is a race condition around the fsfreeze between mount point evaluation of a device and the invocation of the fsfreeze if the device is unmounted from that mount point, the folder will now reference the original device that hosts the mount point folder.

As a hotfix I am removing the fsfreeze invocation and replacing it with our previous behavior of calling sync.

As a result of this change, the volume head is no longer dirty after
a snapshot operation unless there was additional io, so we are again
skipping recurring backups if no new io is present.
REF: longhorn/longhorn#2555

ISSUE: longhorn/longhorn#3125

### DIFF
--- a/pkg/controller/control.go
+++ b/pkg/controller/control.go
@@ -177,21 +177,10 @@ func (c *Controller) Snapshot(name string, labels map[string]string) (string, er
 	if ne, err := iutil.NewNamespaceExecutor(util.GetInitiatorNS()); err != nil {
 		log.WithError(err).Errorf("WARNING: continue to snapshot for %v, but cannot sync due to cannot get the namespace executor", name)
 	} else {
-		// we always try to unfreeze the volume after finishing the snapshot process
-		// if the user manually froze the filesystem this will lead to an unfreeze but it's better than
-		// potentially having a left over frozen volume.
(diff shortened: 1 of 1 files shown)


joshimoo:
@yasker the integration test was added to validate the correct functionality of the volume + snapshot operations for a mounted filesystem, previously if I remember correctly we only ever tested the block dev non mounted case. 

Due to the timing of the fsfreeze/snapshot operation O(1) and the complexities involved to query a fs to see if it's frozen (fsfreeze has no mechanism to do so besides calling freeze/unfreeze, which would negatively impact the test) there is no automatic verification done that the fs is frozen during the snapshot operation outside from the log entries.

Instead it's only verified that the volume + snapshot operation produce the correct expected result.

Some additional research is required in this regard for a new fsfreeze solution but from a quick look it seems like there a syscalls that can be used for bdev and fs evaluation. https://linux-fsdevel.vger.kernel.narkive.com/Dfcmf0Aq/patch-1-2-fsfreeze-add-fiisfrozen-ioctl

yasker:
LGTM.

PhanLe1010:
LGTM

PR #665, opened 2021-10-13
Format error messages correctly

slayer321:
No error . Related to `BugBash`.

Fix `ST1005`

https://lift.sonatype.com/result/longhorn/longhorn-engine/01FHW2ASRM2CB3NTFBG95HJKEQ?t=CustomTool%20%22Staticcheck%22%7CST1005

### DIFF
--- a/integration/core/test_cli.py
+++ b/integration/core/test_cli.py
@@ -1456,7 +1456,7 @@ def test_replica_removal_and_recreation(bin):  # NOQA
         try:
             grpc_controller_client.replica_create(r1_url, True, "RW")
         except grpc.RpcError as grpc_error:
-            assert "Replica must be closed" in grpc_error.details()
+            assert "replica must be closed" in grpc_error.details()
             continue
         break
         time.sleep(RETRY_INTERVAL_SHORT)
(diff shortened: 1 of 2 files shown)


pkg/backend/remote/remote.go
```diff
@@ -261,7 +261,7 @@ func (rf *Factory) Create(address string) (types.Backend, error) {
 	}
 
 	if replica.State != "closed" {
-		return nil, fmt.Errorf("Replica must be closed, Can not add in state: %s", replica.State)
+		return nil, fmt.Errorf("replica must be closed, Can not add in state: %s", replica.State)
```
shuo-wu, line 264: Need to update the assertion that is using this error message in the integration test. Similar below.
https://github.com/longhorn/longhorn-engine/blob/192e2a6922bfec7ca54ce8de67dc193cbec07c92/integration/core/test_cli.py#L1459

pkg/controller/control.go
```diff
@@ -138,10 +138,10 @@ func (c *Controller) canAdd(address string) (bool, error) {
 		return false, nil
 	}
 	if c.hasWOReplica() {
-		return false, fmt.Errorf("Can only have one WO replica at a time")
+		return false, fmt.Errorf("can only have one WO replica at a time")
```
shuo-wu, line 141: https://github.com/longhorn/longhorn-engine/blob/f00c127cd88594521457989ab7b76d696b92b935/integration/core/test_controller.py#L42

shuo-wu (changes requested):

PR #668, opened 2021-10-25
Remove the in-memory backup-replica mapping

jenting:
#### Proposal Changes
Add new flag `--replica` to pass replica address to fetch current backup status.
If the caller does not pass the flag `--replica`, try to find the corresponding replica which has backup creation in progress.

#### Issue
https://github.com/longhorn/longhorn/issues/3035


innobead (approved):
LGTM

PR #671, opened 2021-10-29
nfs: only return coalesce error when last NFSv4 version fails to mount

derekbit:
This PR is related to backupstore.

(1) Only return coalesce error when last NFSv4 version fails to mount.
(2) Remove NFSv3 try mount and umount from defer function.


jenting (approved):

innobead (approved):

innobead:
@derekbit Please backport to 1.2.x.

PR #685, opened 2021-12-30
change base image to SUSE BCI

meldafrawi:
Ref: https://github.com/longhorn/longhorn/issues/2939

Signed-off-by: Mohamed Eldafrawi <mohamed.eldafrawi@suse.com>

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04
+FROM registry.suse.com/bci/bci-base:15.3
 
 ARG DAPPER_HOST_ARCH=amd64
 ARG http_proxy
@@ -14,15 +14,18 @@ ENV DAPPER_RUN_ARGS --privileged --tmpfs /go/src/github.com/longhorn/longhorn-en
 ENV DAPPER_SOURCE /go/src/github.com/longhorn/longhorn-engine
 WORKDIR ${DAPPER_SOURCE}
 
-ENV DEBIAN_FRONTEND=noninteractive
+RUN zypper -n addrepo --refresh https://download.opensuse.org/repositories/system:/snappy/SLE_15/system:snappy.repo && \
+    zypper -n addrepo --refresh https://download.opensuse.org/repositories/devel:tools:scm/SLE_15_SP3/devel:tools:scm.repo && \
+    zypper --gpg-auto-import-keys ref
 
 # Install packages
-RUN apt-get update && \
-    apt-get dist-upgrade -y && \
-    apt-get install -y cmake wget curl git less file \
-        libglib2.0-dev libkmod-dev libnl-genl-3-dev linux-libc-dev pkg-config psmisc tox qemu-utils fuse python3-dev \
-        devscripts debhelper bash-completion librdmacm-dev libibverbs-dev xsltproc docbook-xsl \
-        libconfig-general-perl libaio-dev libc6-dev sg3-utils iptables libltdl7 python3-pip uuid-runtime libdevmapper-dev iproute2 jq
+RUN zypper -n install cmake wget curl git less file gcc \
+    libkmod-devel libnl3-devel linux-glibc-devel pkg-config psmisc tox qemu-tools fuse python3-devel \
+    bash-completion librdmacm1 librdmacm-utils libibverbs xsltproc docbook-xsl-stylesheets \
+    perl-Config-General libaio-devel glibc-devel-static glibc-devel sg3_utils iptables libltdl7 \
+    python3-pip uuid-runtime libdevmapper1_03 iproute2 jq unzip zlib-devel zlib-devel-static \
+    rpm-build rdma-core-devel gcc-c++ docker && \
+    rm -rf rm -rf /var/cache/zypp/*
 
 # needed for ${!var} substitution
 RUN rm -f /bin/sh && ln -s /bin/bash /bin/sh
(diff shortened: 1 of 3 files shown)


Dockerfile.dapper
```diff
@@ -100,16 +96,16 @@ RUN cd /usr/src && \
...
 
 # Build TGT
 RUN cd /usr/src && \
     git clone https://github.com/rancher/tgt.git && \
     cd tgt && \
     git checkout e042fdd3616ca90619637b5826695a3de9b5dd8e && \
-    ./scripts/build-pkg.sh deb && \
-    dpkg -i ./pkg/tgt_*.deb
+    make && \
+    make install
```
innobead, line 108: The same as instance manager, do we still need these (tgt and liblonghorn build in the dapper image)? because you made the change to build it during the build time (package/Dockerfile).

innobead (changes requested):

Dockerfile.dapper
meldafrawi: Yes, for the ci running in Drone.

Dockerfile.dapper
```diff
@@ -14,35 +14,31 @@ ENV DAPPER_RUN_ARGS --privileged --tmpfs /go/src/github.com/longhorn/longhorn-en
...
+    rm -rf rm -rf /var/cache/zypp/*
 
 # needed for ${!var} substitution
 RUN rm -f /bin/sh && ln -s /bin/bash /bin/sh
 
 # Install Go & tools
 ENV GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm64=arm64 GOLANG_ARCH=GOLANG_ARCH_${ARCH} \
     GOPATH=/go PATH=/go/bin:/usr/local/go/bin:${PATH} SHELL=/bin/bash
-RUN wget -O - https://storage.googleapis.com/golang/go1.14.4.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local && \
+RUN wget -O - https://storage.googleapis.com/golang/go1.14.15.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local && \
```
meldafrawi, line 36: Changed the version to v1.13.15 to match the version defined in go.mod

Dockerfile.dapper
```diff
@@ -14,35 +14,31 @@ ENV DAPPER_RUN_ARGS --privileged --tmpfs /go/src/github.com/longhorn/longhorn-en
...
+    rm -rf rm -rf /var/cache/zypp/*
 
 # needed for ${!var} substitution
 RUN rm -f /bin/sh && ln -s /bin/bash /bin/sh
 
 # Install Go & tools
 ENV GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm64=arm64 GOLANG_ARCH=GOLANG_ARCH_${ARCH} \
     GOPATH=/go PATH=/go/bin:/usr/local/go/bin:${PATH} SHELL=/bin/bash
-RUN wget -O - https://storage.googleapis.com/golang/go1.14.4.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local && \
+RUN wget -O - https://storage.googleapis.com/golang/go1.13.15.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local && \
```
meldafrawi, line 36: Changed the version to v1.13.15 to match the version defined in go.mod

Dockerfile.dapper
```diff
@@ -100,16 +96,16 @@ RUN cd /usr/src && \
     git clone https://github.com/rancher/liblonghorn.git && \
     cd liblonghorn && \
     git checkout 09956ed9329d51f41a85d6e06113948c947976a8 && \
-    make deb && \
-    dpkg -i ./pkg/liblonghorn_*.deb
+    make && \
+    make install
```
keithalucas, line 100: Should we had a .spec file to liblonghorn and build an RPM since we use the debian package on Ubuntu.

Dockerfile.dapper
```diff
@@ -100,16 +96,16 @@ RUN cd /usr/src && \
...
 
 # Build TGT
 RUN cd /usr/src && \
     git clone https://github.com/rancher/tgt.git && \
     cd tgt && \
     git checkout e042fdd3616ca90619637b5826695a3de9b5dd8e && \
-    ./scripts/build-pkg.sh deb && \
-    dpkg -i ./pkg/tgt_*.deb
+    make && \
+    make install
```
keithalucas, line 108: There's an RPM spec file in https://github.com/rancher/tgt/blob/longhorn_oct2019/scripts/tgtd.spec.  We could build the RPM like how we build the debian package.

Dockerfile.dapper
innobead: I think it should be fine for now, because we actually don't deliver the artifact but rather just build then install directly.

Dockerfile.dapper
innobead: Exactly, it seems the code is already ready and we can just adopt that.

For s390, I will merge this, and we can change it to rpm way later. (but the same, this is fine as well because the reason is the same as https://github.com/longhorn/longhorn-engine/pull/685#discussion_r851090253)

innobead: cc @yangchiu @meldafrawi

package/Dockerfile
```diff
@@ -1,34 +1,49 @@
...
+    make && \
+    make install && \
+    rm -rf /usr/src/tgt
+
 COPY bin/longhorn bin/longhorn-instance-manager /usr/local/bin/
 
 COPY package/launch-simple-longhorn package/engine-manager package/launch-simple-file /usr/local/bin/
 
-COPY bin/tgt_*.deb /opt/
-
```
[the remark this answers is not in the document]
innobead: Agree, let's improve this later as I mentioned above.

innobead (approved):

PR #691, opened 2022-03-16
Snapshot prune feature & snapshot revert improvement

shuo-wu:
https://github.com/longhorn/longhorn/issues/3613
https://github.com/longhorn/longhorn/issues/3748

### DIFF
--- a/pkg/controller/revert.go
+++ b/pkg/controller/revert.go
@@ -12,23 +12,48 @@ import (
 )
 
 func (c *Controller) Revert(name string) error {
-	doable := false
 	rw := false
 	wo := false
 	for _, rep := range c.replicas {
-		//BUG: Need to deal with replica in rebuilding(WO) mode.
+		// BUG: Need to deal with replica in rebuilding(WO) mode.
 		if rep.Mode == types.RW {
 			rw = true
 		} else if rep.Mode == types.WO {
 			wo = true
 		}
 	}
-	if rw && !wo {
-		doable = true
-	}
-
+	doable := rw && !wo
 	if !doable {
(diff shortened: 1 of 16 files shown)


innobead:
@shuo-wu Please update the dep because https://github.com/longhorn/sparse-tools/pull/80 has been merged.

pkg/controller/revert.go
```diff
@@ -12,25 +12,56 @@ import (
...
+			}
+			if disk.Removed {
+				return fmt.Errorf("cannot revert to removed snapshot %v", name)
+			}
+			// for c := range disk.Children {
+			// 	if c == head {
+			// 		logrus.Infof("there is no need to do revert since the parent of the current volume head is already snapshot %v", name)
+			// 		return nil
+			// 	}
+			// }
```
shuo-wu, line 53: Should we check that the revert target is the snapshot that is directly behind the volume head?

innobead:
@shuo-wu LGTM, please help resolve the conflicts before merging.

pkg/controller/revert.go
innobead: If the current volume head has data then the volume is reverted to the previous snapshot, we should still create a new empty volume head pointing to the snapshot, right? If yes, we don't need this check.

shuo-wu: YES. I just tested this part. The volume will discard the old volume head regardless of the reverting snapshot being the parent.
I have removed this code.

shuo-wu:
2 fixes for the integration failure:
1. Now the replica will reload the snapshot from the backup **full** restore. This guarantees that the snapshot is in the disk chain before the reverting start.
2. Return error if the target snapshot does not exist or is not in the replica disk chain

PR #692, opened 2022-04-12
Check that array isn't empty before accessing

keithalucas:
Prevent longhorn-engine from crashing because it attempts to access the first element in an array whose length is zero.

https://github.com/longhorn/longhorn/issues/3822


shuo-wu (approved):

innobead (approved):

innobead:
@keithalucas let's backport this to 1.2.x.

PR #696, opened 2022-04-22
Fix bug forget to remove the snapshot from r.diskChildrenMap when removing snapshot

PhanLe1010:
Since we forgot to remove the snapshot from the r.diskChildrenMap when removing the snapshot, when the user creates snapshot with the same name again, the children map of the snapshot is messed up

longhorn/longhorn#3883

### DIFF
--- a/pkg/replica/replica.go
+++ b/pkg/replica/replica.go
@@ -492,6 +492,7 @@ func (r *Replica) removeDiskNode(name string, force bool) error {
 		return err
 	}
 	delete(r.diskData, name)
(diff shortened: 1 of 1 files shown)


shuo-wu (approved):
LGTM.
Can we have an engine integration test for this?

innobead (approved):
LGTM, we need to backport this to 1.2.x and 1.1.x.

innobead:
@PhanLe1010 you can have another following unit test later.

PR #702, opened 2022-05-06
Fix bug volume cloning is broken because of failing to revert snapshot

PhanLe1010:
longhorn/longhorn#3886


innobead (approved):

shuo-wu (approved):

innobead:
@PhanLe1010 this probably should be backported to 1.2.x.

PR #704, opened 2022-05-26
Fix the protobuf-python version

derekbit:
[Longhorn 4032](https://github.com/longhorn/longhorn/issues/4032)

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -4,6 +4,7 @@ ARG DAPPER_HOST_ARCH=amd64
 ARG http_proxy
 ARG https_proxy
 ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH}
+ENV PROTOBUF_VER=3.18.0
 
 # Setup environment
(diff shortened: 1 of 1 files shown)


Dockerfile.dapper
```diff
@@ -4,6 +4,8 @@ ARG DAPPER_HOST_ARCH=amd64
 ARG http_proxy
 ARG https_proxy
 ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH}
+ENV PROTOBUF_VER=3.7.1
+ENV PROTOBUF_S390_VER=3.18.0
```
innobead, line 8: Why s390 use a different version?

derekbit: It is because protobuf v3.7.1 does not support s390.
Should we use the same version (3.18.0) for all protobuf utilities?

innobead: As discussed, let's make the version consistent. Also need to check other repos as well.

innobead (approved):

innobead (approved):

PR #706, opened 2022-06-07
Test to find if the longhorn volume is getting automounted

keithalucas:


innobead:
The related issue has been verified in https://github.com/longhorn/longhorn-engine/pull/723.

PR #708, opened 2022-06-16
Reduce write and read calls while reading and writing requests

derekbit:
Writing or reading a header of a request calls multiple writes and reads, which increase the latency of the data io.
The task is to reduce the number of the calls, memory allocations and memory copies.

[Longhorn 4122](https://github.com/longhorn/longhorn/issues/4122)

### DIFF
--- a/pkg/dataconn/wire.go
+++ b/pkg/dataconn/wire.go
@@ -6,39 +6,48 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"unsafe"
 )
 
 type Wire struct {
-	conn   net.Conn
-	writer *bufio.Writer
-	reader io.Reader
+	conn        net.Conn
+	writer      *bufio.Writer
+	reader      io.Reader
+	writeHeader []byte
+	readHeader  []byte
 }
 
 func NewWire(conn net.Conn) *Wire {
 	return &Wire{
-		conn:   conn,
-		writer: bufio.NewWriterSize(conn, writeBufferSize),
-		reader: bufio.NewReaderSize(conn, readBufferSize),
+		conn:        conn,
+		writer:      bufio.NewWriterSize(conn, writeBufferSize),
+		reader:      bufio.NewReaderSize(conn, readBufferSize),
+		writeHeader: make([]byte, getRequestHeaderSize()),
+		readHeader:  make([]byte, getRequestHeaderSize()),
 	}
 }
 
 func (w *Wire) Write(msg *Message) error {
-	if err := binary.Write(w.writer, binary.LittleEndian, msg.MagicVersion); err != nil {
-		return err
-	}
-	if err := binary.Write(w.writer, binary.LittleEndian, msg.Seq); err != nil {
-		return err
-	}
-	if err := binary.Write(w.writer, binary.LittleEndian, msg.Type); err != nil {
-		return err
-	}
-	if err := binary.Write(w.writer, binary.LittleEndian, msg.Offset); err != nil {
-		return err
-	}
-	if err := binary.Write(w.writer, binary.LittleEndian, msg.Size); err != nil {
-		return err
-	}
-	if err := binary.Write(w.writer, binary.LittleEndian, uint32(len(msg.Data))); err != nil {
+	offset := 0
+
+	binary.LittleEndian.PutUint16(w.writeHeader[offset:], msg.MagicVersion)
+	offset += int(unsafe.Sizeof(msg.MagicVersion))
+
+	binary.LittleEndian.PutUint32(w.writeHeader[offset:], msg.Seq)
+	offset += int(unsafe.Sizeof(msg.Seq))
+
+	binary.LittleEndian.PutUint32(w.writeHeader[offset:], msg.Type)
+	offset += int(unsafe.Sizeof(msg.Type))
+
+	binary.LittleEndian.PutUint64(w.writeHeader[offset:], uint64(msg.Offset))
+	offset += int(unsafe.Sizeof(msg.Offset))
+
+	binary.LittleEndian.PutUint32(w.writeHeader[offset:], msg.Size)
+	offset += int(unsafe.Sizeof(msg.Size))
+
+	binary.LittleEndian.PutUint32(w.writeHeader[offset:], uint32(len(msg.Data)))
+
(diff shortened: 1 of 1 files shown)


pkg/dataconn/wire.go
```diff
@@ -6,6 +6,11 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"unsafe"
+)
+
+const (
+	requestHeaderSize = 26
```
shuo-wu, line 13: I prefer to make this as a variable in `Wire` and it would be calculated during the init of the struct.

shuo-wu (commented):
Nice Catch!

pkg/dataconn/wire.go
```diff
@@ -23,22 +28,27 @@ func NewWire(conn net.Conn) *Wire {
...
+
+	binary.LittleEndian.PutUint64(header[offset:], uint64(msg.Offset))
+	offset += int(unsafe.Sizeof(msg.Offset))
+
+	binary.LittleEndian.PutUint32(header[offset:], msg.Size)
+	offset += int(unsafe.Sizeof(msg.Size))
+
+	binary.LittleEndian.PutUint32(header[offset:], uint32(len(msg.Data)))
+
+	if _, err := w.writer.Write(header); err != nil {
```
keithalucas, line 51: In general, it seems like we are trying to "buffer" the I/O to reduce system calls.  The `writer` is a `bufio.Writer` which seems buffer the I/O for us already.   Is the improvement because of less calls to `bufio.Writer.Write`?  I can't see how this could result in less system calls when the I/O is already using a buffer to reduce the amount of system calls?  Is the buffering not working?

pkg/dataconn/wire.go
keithalucas: Could you make this `unsafe.Sizeof(msg.MagicVersion) + unsafe.Sizeof(msg.Seq) ...`?

derekbit: Sure.

pkg/dataconn/wire.go
```diff
@@ -55,29 +65,32 @@ func (w *Wire) Read() (*Message, error) {
 		length uint32
 	)
 
-	if err := binary.Read(w.reader, binary.LittleEndian, &msg.MagicVersion); err != nil {
+	header := make([]byte, requestHeaderSize)
+	offset := 0
+
+	if _, err := io.ReadFull(w.reader, header); err != nil {
```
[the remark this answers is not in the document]
derekbit: Yes, you are right. After checking the `bufio`, the amount of the syscall is not minimized. The improvement here is the reduction of the amount of memory allocations.

pkg/dataconn/wire.go
derekbit: https://github.com/longhorn/longhorn-engine/pull/708#discussion_r899271299

pkg/dataconn/wire.go
```diff
@@ -23,22 +28,27 @@ func NewWire(conn net.Conn) *Wire {
...
+	binary.LittleEndian.PutUint64(header[offset:], uint64(msg.Offset))
+	offset += int(unsafe.Sizeof(msg.Offset))
+
+	binary.LittleEndian.PutUint32(header[offset:], msg.Size)
+	offset += int(unsafe.Sizeof(msg.Size))
+
+	binary.LittleEndian.PutUint32(header[offset:], uint32(len(msg.Data)))
+
+	if _, err := w.writer.Write(header); err != nil {
 		return err
```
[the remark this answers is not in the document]
derekbit: A little confused.
Do we really need to flush the header before writing data? The whole write `Write` operation already ends up with `Flush`.

pkg/dataconn/wire.go
```diff
@@ -23,22 +24,27 @@ func NewWire(conn net.Conn) *Wire {
 }
```
shuo-wu, line 24: Put the header size calculation & the header buffer init here

pkg/dataconn/wire.go
```diff
@@ -23,22 +28,27 @@ func NewWire(conn net.Conn) *Wire {
...
-		return err
-	}
-	if err := binary.Write(w.writer, binary.LittleEndian, msg.Offset); err != nil {
-		return err
-	}
-	if err := binary.Write(w.writer, binary.LittleEndian, msg.Size); err != nil {
-		return err
-	}
-	if err := binary.Write(w.writer, binary.LittleEndian, uint32(len(msg.Data))); err != nil {
+	header := make([]byte, requestHeaderSize)
```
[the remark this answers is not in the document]
derekbit: I've moved the header buffer to `wire` struc, so we don't need a goroutine for this.

pkg/dataconn/wire.go
derekbit: Updated. Thank you. The performance (latency) is improved more.

pkg/dataconn/wire.go
derekbit: Updated.

pkg/dataconn/wire.go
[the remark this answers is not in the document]
keithalucas: Sorry I missed the flush expanding the code from the PR.

keithalucas (approved):

shuo-wu (approved):

pkg/dataconn/wire.go
```diff
@@ -91,3 +102,14 @@ func (w *Wire) Read() (*Message, error) {
...
+
+func getRequestHeaderSize() int {
+	var msg Message
+
+	return int(unsafe.Sizeof(msg.MagicVersion)) +
+		int(unsafe.Sizeof(msg.Seq)) +
+		int(unsafe.Sizeof(msg.Type)) +
+		int(unsafe.Sizeof(msg.Offset)) +
+		int(unsafe.Sizeof(msg.Size)) +
+		4 // int(len(msg.Data))
```
innobead, line 114: LGTM.

After discussing with @derekbit it's better to explain why we can assume the size of data is 4 bytes here.

derekbit: Updated. Thank you.

innobead (approved):

innobead:
@derekbit Need to backport to 1.3.x and 1.2.x if needed.

PR #710, opened 2022-06-23
Reduce write and read calls while processing requests

derekbit:
Writing or reading a header of a request calls multiple writes and reads
which increase the latency of the data io because the memory allocations
and data copy.

https://github.com/longhorn/longhorn/issues/4122
https://github.com/longhorn/longhorn/issues/4170


innobead (approved):

innobead:
Merge this after the PR check fixed and passed.

PR #716, opened 2022-07-09
Add a code to judge whether the address is duplicate

DaleWYZ:
In the 3 nodes k8s environment, if longhorn use the setting of storage-network, and the second network is contronllerd by ovs-cni,  there is 3 instance-manager of replica, and the instance-manager of replica may have the same storage ip address. And now we create a volume whit 2 replicas, the instance-manager of engine may connect to the 2 instance-manager of replica with the same  ip address, this will make some mistake, and we can see the log below(In the log the error is caused by network, but  ip address duplication is still a problem ). So i add some code to judge whether the address of instance-manager-r is duplicated.
 
`[longhorn-instance-manager] time="2022-07-09T07:12:39Z" level=info msg="Process Manager: created process vol111-e-52d90c97"
[vol111-e-52d90c97] time="2022-07-09T07:12:39Z" level=info msg="**Starting with replicas [\"tcp://192.168.188.2:10000\" \"tcp://192.168.188.2:10000\"]**"
time="2022-07-09T07:12:39Z" level=info msg="Connecting to remote: 192.168.188.2:10000"
[vol111-e-52d90c97] time="2022-07-09T07:12:39Z" level=warning msg="failed to create backend with address tcp://192.168.188.2:10000: failed to get replica 192.168.188.2:10000 info from remote: rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = \"transport: Error while dialing dial tcp 192.168.188.2:10000: connect: connection refused\""
time="2022-07-09T07:12:39Z" level=info msg="Connecting to remote: 192.168.188.2:10000"
time="2022-07-09T07:12:39Z" level=warning msg="failed to create backend with address tcp://192.168.188.2:10000: failed to get replica 192.168.188.2:10000 info from remote: rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = \"transport: Error while dialing dial tcp 192.168.188.2:10000: connect: connection refused\""
time="2022-07-09T07:12:39Z" level=info msg="Adding backend: tcp://192.168.188.2:10000"
2022/07/09 07:12:39 **cannot create an available backend for the engine from the addresses [tcp://192.168.188.2:10000 tcp://192.168.188.2:10000]**
`

### DIFF
--- a/pkg/controller/control.go
+++ b/pkg/controller/control.go
@@ -584,6 +584,14 @@ func (c *Controller) Start(addresses ...string) error {
 		return nil
 	}
 
+	addressReplicaMap := map[string]string{}
+	for _, address := range addresses {
+		if addressReplicaMap[address] != "" {
+			return fmt.Errorf("invalid ReplicaAddress: duplicate replica addresses")
+		}
+		addressReplicaMap[address] = address
+	}
+
 	if len(c.replicas) > 0 {
 		return nil
 	}
--- a/pkg/controller/control.go
+++ b/pkg/controller/control.go
@@ -584,12 +584,12 @@ func (c *Controller) Start(addresses ...string) error {
 		return nil
 	}
 
-	addressReplicaMap := map[string]string{}
+	checkDuplicate := map[string]struct{}{}
 	for _, address := range addresses {
-		if addressReplicaMap[address] != "" {
+		if _, exist := checkDuplicate[address]; exist {
 			return fmt.Errorf("invalid ReplicaAddress: duplicate replica addresses")
 		}
-		addressReplicaMap[address] = address
+		checkDuplicate[address] = struct{}{}
 	}
 
 	if len(c.replicas) > 0 {
--- a/pkg/controller/control.go
+++ b/pkg/controller/control.go
@@ -587,7 +587,7 @@ func (c *Controller) Start(addresses ...string) error {
 	checkDuplicate := map[string]struct{}{}
 	for _, address := range addresses {
 		if _, exist := checkDuplicate[address]; exist {
-			return fmt.Errorf("invalid ReplicaAddress: duplicate replica addresses")
+			return fmt.Errorf("invalid ReplicaAddress: duplicate replica addresses %s", address)
 		}
 		checkDuplicate[address] = struct{}{}
 	}


pkg/controller/control.go
```diff
@@ -584,6 +584,14 @@ func (c *Controller) Start(addresses ...string) error {
 		return nil
 	}
 
+	addressReplicaMap := map[string]string{}
+	for _, address := range addresses {
+		if addressReplicaMap[address] != "" {
+			return fmt.Errorf("invalid ReplicaAddress: duplicate replica addresses")
+		}
+		addressReplicaMap[address] = address
+	}
```
c3y1huang, line 593: nit
```suggestion
	checkDuplicate := map[string]struct{}{}
	for _, address := range addresses {
		if _, exist := checkDuplicate[address]; exist {
			return fmt.Errorf("invalid ReplicaAddress: duplicate replica addresses")
		}
		checkDuplicate[address] = struct{}{}
	}
```

c3y1huang (approved):
Just a nit.
Need to backport to master.

innobead:
> In the 3 nodes k8s environment, if longhorn use the setting of storage-network, and the second network is contronllerd by ovs-cni, there is 3 instance-manager of replica, and the instance-manager of replica may have the same storage ip address. And now we create a volume whit 2 replicas, the instance-manager of engine may connect to the 2 instance-manager of replica with the same ip address, this will make some mistake, and we can see the log below(In the log the error is caused by network, but ip address duplication is still a problem ). So i add some code to judge whether the address of instance-manager-r is duplicated.
> 
> `[longhorn-instance-manager] time="2022-07-09T07:12:39Z" level=info msg="Process Manager: created process vol111-e-52d90c97" [vol111-e-52d90c97] time="2022-07-09T07:12:39Z" level=info msg="**Starting with replicas [\"tcp://192.168.188.2:10000\" \"tcp://192.168.188.2:10000\"]**" time="2022-07-09T07:12:39Z" level=info 

@DaleWYZ could you elaborate more details about `ovs-cni` for the storage network? Why will it cause instance manager replica pods to have the same IP when enabling the storage network? In Longhorn design, each instance manage pod will have a different IP whatever it's in the default/storage network mode.

 We need to clarify the behavior first before adding this validation even if it's OK to do that.

DaleWYZ:
@innobead 
Because when  creating the ovs-cni NetworkAttachmentDefinition ,i use the 'host-local' as the config of ipam's type. 'host-local'  can only ensure that the IP address on a single node is not repeated, but in the whole cluster may appear duplicate IP address. Then i use the 'whereabouts' as the config of ipam's type, a CNI IPAM plugin that assigns IP addresses cluster-wide, to aviod this problem.

innobead:
> @innobead Because when creating the ovs-cni NetworkAttachmentDefinition ,i use the 'host-local' as the config of ipam's type. 'host-local' can only ensure that the IP address on a single node is not repeated, but in the whole cluster may appear duplicate IP address. Then i use the 'whereabouts' as the config of ipam's type, a CNI IPAM plugin that assigns IP addresses cluster-wide, to aviod this problem.

Thanks @DaleWYZ . Just confirm that after using the whereabouts IPAM plugin, right now the storage IP of each instance manage pod should have unique IP, right?

DaleWYZ:
> > @innobead Because when creating the ovs-cni NetworkAttachmentDefinition ,i use the 'host-local' as the config of ipam's type. 'host-local' can only ensure that the IP address on a single node is not repeated, but in the whole cluster may appear duplicate IP address. Then i use the 'whereabouts' as the config of ipam's type, a CNI IPAM plugin that assigns IP addresses cluster-wide, to aviod this problem.
> 
> Thanks @DaleWYZ . Just confirm that after using the whereabouts IPAM plugin, right now the storage IP of each instance manage pod should have unique IP, right?

@innobead Yes, and here some introduction about whereabouts below:
_CNI (Container Network Interface) plugins typically have a configuration element named ipam. CNI IPAM plugins can assign IP addresses, and Whereabouts assigns IP addresses within a range -- without having to use a DHCP server.

Whereabouts takes an address range, like 192.168.2.0/24 in CIDR notation, and will assign IP addresses within that range. In this case, it will assign IP addresses from 192.168.2.1 to 192.168.2.255. When an IP address is assigned to a pod, Whereabouts tracks that IP address in a data store for the lifetime of that pod. When the pod is removed, Whereabouts then frees the address and makes it available to assign on subsequent requests. Whereabouts always assigns the lowest value address that's available in the range._

shuo-wu (dismissed):

pkg/controller/control.go
```diff
@@ -584,6 +584,14 @@ func (c *Controller) Start(addresses ...string) error {
 		return nil
 	}
 
+	checkDuplicate := map[string]struct{}{}
+	for _, address := range addresses {
+		if _, exist := checkDuplicate[address]; exist {
+			return fmt.Errorf("invalid ReplicaAddress: duplicate replica addresses")
```
innobead, line 590: Could you have a format string w/ the duplicated address inside? `("invalid ReplicaAddress: duplicate replica addresses %s", address)`

innobead (changes requested):

innobead:
> @innobead Yes, and here some introduction about whereabouts below: _CNI (Container Network Interface) plugins typically have a configuration element named ipam. CNI IPAM plugins can assign IP addresses, and Whereabouts assigns IP addresses within a range -- without having to use a DHCP server.
> 
> Whereabouts takes an address range, like 192.168.2.0/24 in CIDR notation, and will assign IP addresses within that range. In this case, it will assign IP addresses from 192.168.2.1 to 192.168.2.255. When an IP address is assigned to a pod, Whereabouts tracks that IP address in a data store for the lifetime of that pod. When the pod is removed, Whereabouts then frees the address and makes it available to assign on subsequent requests. Whereabouts always assigns the lowest value address that's available in the range._

@DaleWYZ thanks for the update. It's good to know it works at the end after applying whereabouts IPAM 👍  (cc @c3y1huang )

innobead (approved):

innobead:
@mantissahz Please help this contribution backport to the master and 1.2.x. Thank you.

joshimoo:
@innobead @shuo-wu I am a little confused and feel like this needs additional investigation.
How can 2 processes bind to the same IP:PORT pair on the same node?
Each replica process would still need to have a unique ip:port (socket) for communication

shuo-wu:
Here the user said that
> when creating the ovs-cni NetworkAttachmentDefinition ,i use the 'host-local' as the config of ipam's type. 'host-local' can only ensure that the IP address on a single node is not repeated, but in the whole cluster may appear duplicate IP address.

I am not sure if the pods with the same IP are still reachable, but the check added by this PR can be a guard.

innobead:
> @innobead @shuo-wu I am a little confused and feel like this needs additional investigation. How can 2 processes bind to the same IP:PORT pair on the same node? Each replica process would still need to have a unique ip:port (socket) for communication

This is more like the incorrect operation from the users when they apply the storage network with the incorrect NetworkAttachmentDefinition w/ host-local IPAM which will potentially cause duplicated IP for the storage interface in the pod (ref: https://github.com/longhorn/longhorn-engine/pull/716#issuecomment-1186794262).

So, adding this validation just ensures the right error is raised when the volume engine runs under this situation, so it's helpful for users to understand the situation.

PR #717, opened 2022-07-11
Retry couple times if restoring a backup failed

mantissahz:
simple retry to give another change for restoring to get the lock.

longhorn/longhorn#3055

Signed-off-by: James Lu <james.lu@suse.com>


go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/gorilla/handlers v1.4.2
 	github.com/gorilla/mux v1.7.3
 	github.com/gorilla/websocket v1.4.2 // indirect
+	github.com/jpillora/backoff v1.0.0
```
shuo-wu, line 15: Oh, it's a new backoff pkg. We are using `"k8s.io/client-go/util/flowcontrol"` in longhorn-manager repo. Not sure which one is better for this case

pkg/backup/main.go
```diff
@@ -136,14 +143,33 @@ func DoBackupRestore(backupURL string, toFile string, restoreObj *replica.Restor
...
+
 	config := &backupstore.DeltaRestoreConfig{
 		BackupURL: backupURL,
 		DeltaOps:  restoreObj,
 		Filename:  toFile,
 	}
 
-	if err := backupstore.RestoreDeltaBlockBackup(config); err != nil {
-		return err
+	for {
```
shuo-wu, line 160: I don't think we can add a wait & retry loop here. This will block the restore call then lead to timeout errors.

shuo-wu (changes requested):
Right now the restore handling in the sync agent server consists of 2 parts:
1. Sync part: Init the restore progress then return immediately. This cannot be blocked.
2. Async part: wait for the restore complete then do some wind-up operations

go.mod
mantissahz: I think `"k8s.io/client-go/util/flowcontrol"` is good enough, I would turn to use this pkg.

pkg/backup/main.go
mantissahz: Yes, so It has just 5 times to retry and timeout in one minute.
I should turn to `Async part` to handle the lock acquisition failed, right?

mantissahz: or Could I check the getting lock failed here [sync() and refresh()](https://github.com/longhorn/longhorn-manager/blob/master/controller/engine_controller.go#L727) and retry?

mantissahz:
Close this PR, and retry the restore in upper layer.

PR #719, opened 2022-07-18
Enable test_snapshot_mounted_filesystem

derekbit:
[Longhorn 4089](https://github.com/longhorn/longhorn/issues/4089)


derekbit:
> I believe this is only for testing purpose, so we will not need this anymore after resolving multipath running on our infra nodes.

Yes, if the tests pass, I will enable the test function `test_snapshot_mounted_filesystem` in this PR.

innobead (approved):

innobead:
@derekbit Need to backport to 1.3.x and 1.2.x if the case was disabled.

PR #722, opened 2022-07-20
Reject snapshot creation request if the snapshot is already existing

derekbit:
Creating a snapshot with a same ID (name) will lead to the volume
corruption. The fix is rejecting the snapshot creation requst if
the snapshot is already existing.

[Longhorn 4250](https://github.com/longhorn/longhorn/issues/4250)

### DIFF
--- a/pkg/controller/control.go
+++ b/pkg/controller/control.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"fmt"
 	"net"
+	"regexp"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -713,19 +714,37 @@ func (c *Controller) ReadAt(b []byte, off int64) (int, error) {
 	return n, err
 }
 
+func isSnapshotDiskExist(err error) bool {
+	match, _ := regexp.MatchString("snapshot (.*) is already existing", err.Error())
+	return match
+}
+
 func (c *Controller) handleErrorNoLock(err error) error {
 	if bErr, ok := err.(*BackendError); ok {
+		snapshotExistList := make(map[string]struct{})
+
 		if len(bErr.Errors) > 0 {
 			for address, replicaErr := range bErr.Errors {
-				logrus.Errorf("Setting replica %s to ERR due to: %v", address, replicaErr)
-				c.setReplicaModeNoLock(address, types.ERR)
+				if isSnapshotDiskExist(replicaErr) {
+					// The snapshot request using a existing snapshot's name might be caused by
+					// users and callers unexpectedly.
+					// We reject the request, so do not set the replica to ERR if the snapshot is already existing.
+					snapshotExistList[address] = struct{}{}
+				} else {
+					logrus.Errorf("Setting replica %s to ERR due to: %v", address, replicaErr)
+					c.setReplicaModeNoLock(address, types.ERR)
+				}
 			}
-			// if we still have a good replica, do not return error
-			for _, r := range c.replicas {
-				if r.Mode == types.RW {
-					logrus.Errorf("Ignoring error because %s is mode RW: %v", r.Address, err)
-					err = nil
-					break
+
+			// Always return error if the snapshot is already existing.
+			if len(snapshotExistList) == 0 {
+				// if we still have a good replica, do not return error
+				for _, r := range c.replicas {
+					if r.Mode == types.RW {
+						logrus.Errorf("Ignoring error because %s is mode RW: %v", r.Address, err)
+						err = nil
+						break
+					}
 				}
 			}
 		}
--- a/pkg/replica/replica.go
+++ b/pkg/replica/replica.go
@@ -771,12 +771,16 @@ func (r *Replica) linkDisk(oldname, newname string) error {
 		return nil
 	}
 
+	destMetadata := r.diskPath(newname + metadataSuffix)
+	logrus.Infof("Deleting old disk metadata file %v", destMetadata)
+	if err := os.RemoveAll(destMetadata); err != nil {
+		return errors.Wrapf(err, "failed to delete old disk metadata file %v", destMetadata)
+	}
+
 	dest := r.diskPath(newname)
-	if _, err := os.Stat(dest); err == nil {
-		logrus.Infof("Old file %s exists, deleting", dest)
-		if err := os.Remove(dest); err != nil {
-			return err
-		}
+	logrus.Infof("Deleting old disk file %v", dest)
+	if err := os.RemoveAll(dest); err != nil {
+		return errors.Wrapf(err, "failed to delete old disk file %v", dest)
 	}
 
 	if err := os.Link(r.diskPath(oldname), dest); err != nil {
@@ -868,6 +872,12 @@ func (r *Replica) createDisk(name string, userCreated bool, created string, labe
 
 	if oldHead == "" {
 		newSnapName = ""
+	}
+
+	if newSnapName != "" {
+		if _, ok := r.diskData[newSnapName]; ok {
+			return fmt.Errorf("snapshot %v is already existing", newSnapName)
+		}
 	}
 
 	f, newHeadDisk, err := r.createNewHead(oldHead, newSnapName, created, size)


derekbit:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1696

derekbit:
> e2e https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1696

The e2e result looks good. It can be reviewed now.

pkg/controller/control.go
```diff
@@ -714,18 +715,28 @@ func (c *Controller) ReadAt(b []byte, off int64) (int, error) {
 }
 
 func (c *Controller) handleErrorNoLock(err error) error {
+	snapshotDiskAlreadyExist := false
+
 	if bErr, ok := err.(*BackendError); ok {
 		if len(bErr.Errors) > 0 {
 			for address, replicaErr := range bErr.Errors {
+				if strings.Contains(replicaErr.Error(), "is already existing") {
```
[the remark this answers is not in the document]
derekbit: The snapshots info is in Replica, so we cannot check if the snapshot is already existing or not before calling `handleErrorNoLock(c.backend.Snapshot)`.

Or, should we have another function `handleSnapshotErrorNoLock`?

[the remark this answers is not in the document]
derekbit: Or, having a function such `SnapshotExist` for check if the snapshot is existing is also an option.

pkg/replica/replica.go
```diff
@@ -870,6 +882,12 @@ func (r *Replica) createDisk(name string, userCreated bool, created string, labe
 		newSnapName = ""
 	}
 
+	if newSnapName != "" {
+		if _, ok := r.diskData[newSnapName]; ok {
+			return fmt.Errorf(fmt.Sprintf("snapshot %v is already existing", newSnapName))
```
c3y1huang, line 887:
```suggestion
			return fmt.Errorf("snapshot %v is already existing", newSnapName)
```

pkg/controller/control.go
```diff
@@ -714,18 +715,28 @@ func (c *Controller) ReadAt(b []byte, off int64) (int, error) {
 }
 
 func (c *Controller) handleErrorNoLock(err error) error {
+	snapshotDiskAlreadyExist := false
```
c3y1huang, line 718: Suggest moving this to the same layer where the variable is used.

pkg/replica/replica.go
```diff
@@ -773,10 +773,22 @@ func (r *Replica) linkDisk(oldname, newname string) error {
...
+		if _, err := os.Stat(destMetadata); err == nil {
+			logrus.Infof("Old disk metadata file %s exists, deleting", destMetadata)
+			if err := os.Remove(destMetadata); err != nil {
+				return err
+			}
+		} else if !os.IsNotExist(err) {
+			return errors.Wrapf(os.ErrExist, "failed to stat disk metadata file %v", destMetadata)
+		}
+
+		logrus.Infof("Old disk file %s exists, deleting", dest)
```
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

pkg/replica/replica.go
derekbit: Addressed. Thank you.

pkg/controller/control.go
derekbit: Addressed. Thank you.

pkg/replica/replica.go
```diff
@@ -773,10 +773,22 @@ func (r *Replica) linkDisk(oldname, newname string) error {
 
 	dest := r.diskPath(newname)
 	if _, err := os.Stat(dest); err == nil {
-		logrus.Infof("Old file %s exists, deleting", dest)
+		destMetadata := r.diskPath(newname + metadataSuffix)
+		if _, err := os.Stat(destMetadata); err == nil {
+			logrus.Infof("Deleting old disk metadata file %s", destMetadata)
+			if err := os.Remove(destMetadata); err != nil {
```
innobead, line 779: Just note: the disk meta should be the source of truth to ensure the atomicity of each disk (image + metafile).

https://github.com/longhorn/longhorn-engine/blob/c3a9bf904fa7b02737b93bd18efb24d5f32661c7/pkg/replica/replica.go#L187-L186

innobead (dismissed):
LGTM

pkg/replica/replica.go
shuo-wu: Can we blindly remove the existing meta file then the image file here? I mean, using `os.RemoveAll(filePath)` without `os.State()`

pkg/controller/control.go
```diff
@@ -713,21 +714,37 @@ func (c *Controller) ReadAt(b []byte, off int64) (int, error) {
...
-			}
-			// if we still have a good replica, do not return error
-			for _, r := range c.replicas {
-				if r.Mode == types.RW {
-					logrus.Errorf("Ignoring error because %s is mode RW: %v", r.Address, err)
-					err = nil
+			for _, replicaErr := range bErr.Errors {
+				if isSnapshotDiskExist(replicaErr) {
+					snapshotDiskIsExist = true
 					break
```
shuo-wu, line 730: Can we break without checking all replica errors?

pkg/controller/control.go
```diff
@@ -713,21 +714,37 @@ func (c *Controller) ReadAt(b []byte, off int64) (int, error) {
...
 				}
 			}
+
+			if !snapshotDiskIsExist {
```
shuo-wu, line 734: It's better to print error/warning logs when there are `snapshotDiskIsExist` errors

shuo-wu (commented):
1. Since there is a locking during engine snapshotting. Can we check the snapshot name before sending the requests to the replicas?
2. Last time in the debug session, we found that the rollback after the snapshotting failure (in replica side) is buggy. Does this PR fix it? Or will you create a separate PR for it?

derekbit:
> Since there is a locking during engine snapshotting. Can we check the snapshot name before sending the requests to the replicas?

I think yes, but we need a function such as `backend.SnapshotExist` for checking the existence of the snapshot.

> Last time in the debug session, we found that the rollback after the snapshotting failure (in replica side) is buggy. Does this PR fix it? Or will you create a separate PR for it?

After reviewing the logic, if we reject the request, we don't need a fix of the rollback.

pkg/replica/replica.go
derekbit: > Can we blindly remove the existing meta file then the image file here? I mean, using `os.RemoveAll(filePath)` without `os.State()`

Yes, we can replace `Stat...Remove` with `RemoveAll`.

pkg/controller/control.go
derekbit: Hmm...Probably some replicas have `Snapsho Exist` error and some replicas have other errors.
Should no break here.

pkg/controller/control.go
derekbit: The error log will be printed in the end of this function.

pkg/controller/control.go
derekbit: Addressed.  Thank you.

pkg/replica/replica.go
derekbit: Updated. Thank you.

shuo-wu:
> I think yes, but we need a function such as `backend.SnapshotExist` for checking the existence of the snapshot. In the controller, there is no knowledge about the replicas' snapshots, so the queries to replicas are inevitable. It will doubles the calls between engine and replicas (the first is `backend.SnapshotExist`, and the second is `backend.Snapshot`). Thus, I prefer the current design. WDYT?

OK. I am okay with keeping the current design.


> After reviewing the logic, if we reject the request, we don't need a fix of the rollback.

What if there are other kinds of errors that lead to the snapshotting failure?

derekbit:
> What if there are other kinds of errors that lead to the snapshotting failure?

Because the new snapshot name is not conflict with the previous snapshots anymore, the rollback caused by another failure can be done correctly.

shuo-wu:
IIRC, there is something wrong with the volume metafile info after the rollback. Do you mean that this issue does not exist when the snapshot failure is not caused by duplicate naming?
If YES then we don't need to take care of the rollback part.

derekbit:
> IIRC, there is something wrong with the volume metafile info after the rollback. Do you mean that this issue does not exist when the snapshot failure is not caused by duplicate naming? If YES then we don't need to take care of the rollback part.

Yes. The issue does not exist if there is no duplicate naming issue.
The volume meta is handled correctly in other failure cases.
https://github.com/longhorn/longhorn-engine/blob/master/pkg/replica/replica.go#L877

pkg/controller/control.go
```diff
@@ -713,19 +714,41 @@ func (c *Controller) ReadAt(b []byte, off int64) (int, error) {
...
+
+			for address, replicaErr := range bErr.Errors {
+				// The snapshot request using a existing snapshot's name might be caused by
+				// users and callers unexpectedly.
+				// We reject the request, so do not set the replica to ERR if the snapshot is already existing.
+				if _, ok := snapshotExistList[address]; !ok {
+					logrus.Errorf("Setting replica %s to ERR due to: %v", address, replicaErr)
+					c.setReplicaModeNoLock(address, types.ERR)
+				}
+			}
```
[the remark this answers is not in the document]
derekbit: Updated. Thank you.

shuo-wu (dismissed):

pkg/replica/replica.go
```diff
@@ -772,11 +772,15 @@ func (r *Replica) linkDisk(oldname, newname string) error {
 	}
 
 	dest := r.diskPath(newname)
-	if _, err := os.Stat(dest); err == nil {
-		logrus.Infof("Old file %s exists, deleting", dest)
-		if err := os.Remove(dest); err != nil {
-			return err
-		}
+	if err := os.RemoveAll(dest); err != nil {
+		logrus.Infof("Deleting old disk file %v", dest)
```
c3y1huang, line 776: This only shows on error, should we move this out of the if statement?

pkg/replica/replica.go
```diff
@@ -772,11 +772,15 @@ func (r *Replica) linkDisk(oldname, newname string) error {
...
+		return errors.Wrapf(err, "failed to delete old disk file %v", dest)
+	}
+
+	destMetadata := r.diskPath(newname + metadataSuffix)
+	if err := os.RemoveAll(destMetadata); err != nil {
+		logrus.Infof("Deleting old disk metadata file %v", destMetadata)
```
c3y1huang, line 782: This only shows on error, should we move this out of the if statement?

pkg/replica/replica.go
derekbit: Updated.  Thanks.

innobead (approved):

PR #727, opened 2022-07-25
Duplicated IP address of sotrage-network

mantissahz:
If longhorn use the setting of storage-network, and the second
network is contronllerd by ovs-cni, there is three
instance-managers of replica, and the instance-manager of replica
may have the same storage ip address.

longhorn/longhorn#4281

Signed-off-by: 王一知 <wangyizhi@inspur.com>


innobead (approved):

innobead:
Thanks @mantissahz

PR #736, opened 2022-08-12
Mark the replica ERR if its volume.meta is in invalid state

derekbit:
If there is any available replica, mark the replicas ERR if their volume.meta are in invalid state.

https://github.com/longhorn/longhorn/issues/4212

### DIFF
--- a/pkg/controller/control.go
+++ b/pkg/controller/control.go
@@ -466,7 +466,7 @@ func (c *Controller) StartFrontend(frontend string) error {
 // controller, and mark unmatch replica to ERR.
 func (c *Controller) checkReplicaRevCounterSettingMatch() error {
 	for _, r := range c.replicas {
-		if r.Mode == "ERR" {
+		if r.Mode == types.ERR {
 			continue
 		}
 		revCounterDisabled, err := c.backend.backends[r.Address].backend.IsRevisionCounterDisabled()
@@ -489,7 +489,7 @@ func (c *Controller) salvageRevisionCounterDisabledReplicas() error {
 	replicaCandidates := make(map[types.Replica]types.ReplicaSalvageInfo)
 	var lastModifyTime int64
 	for _, r := range c.replicas {
-		if r.Mode == "ERR" {
+		if r.Mode == types.ERR {
 			continue
 		}
 		repLastModifyTime, err := c.backend.backends[r.Address].backend.GetLastModifyTime()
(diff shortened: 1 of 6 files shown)


shuo-wu (commented):
Putting this state check in the revision counter check function looks a little bit weird... 
Will the replica process check the state during the starting stage?

derekbit:
> Putting this state check in the revision counter check function looks a little bit weird... Will the replica process check the state during the starting stage?

Agreed. But I don't know why the original state check is in the revision counter check funciton...

derekbit:
> Will the replica process check the state during the starting stage?

Actually, the check is in `c.backend.GetRevisionCounter(r.Address)`...

shuo-wu:
I may prefer to make this check a separate function, or a part of the replica `Open` function. But due to the complexity of the replica state, more investigation is needed. What do you think?

derekbit:
> 

Yes, I prefer a new function `checkReplicasState` for checking the state before the `checkReplicasRevisionCounter`.

>  due to the complexity of the replica state, more investigation is needed. 

Sure.

pkg/controller/control.go
```diff
@@ -660,6 +697,12 @@ func (c *Controller) Start(addresses ...string) error {
 	// For more details, see the following url:
 	// https://github.com/longhorn/longhorn/issues/1235
 	if !c.isUpgrade {
+		// Check if the replicas are in valid state.
+		// If not in a valid state, mark the replica as ERR.
+		if err := c.checkReplicasState(); err != nil {
+			return err
+		}
```
shuo-wu, line 704: I think we can move this as [a part of function `c.Start()`](https://github.com/longhorn/longhorn-engine/blob/7310e09cd6632276f29ae8659a1388b9ee8edbc7/pkg/controller/control.go#L605-L649). 
Then you can eliminate some duplications for `checkReplicasState`

derekbit: Updated. Thank you.

shuo-wu (dismissed):

innobead (changes requested):
LGTM, but just a few feedbacks to resolve.

pkg/controller/control.go
```diff
@@ -577,6 +578,23 @@ func (c *Controller) checkReplicasRevisionCounter() error {
 	return nil
 }
 
+func replicaIsInInvalidState(state string) bool {
```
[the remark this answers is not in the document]
derekbit: Updated. Thank you.

innobead (approved):

PR #740, opened 2022-08-29
Set replica to ERR if the size is mismatched with expected size

derekbit:
[Longhorn 3601](https://github.com/longhorn/longhorn/issues/3601)


proto/ptypes/controller.proto
```diff
@@ -59,7 +59,8 @@ message ControllerReplica {
 }
 
 message VolumeStartRequest {
-    repeated string replicaAddresses = 1;
+    int64 size = 1;
+    repeated string replicaAddresses = 2;
```
shuo-wu, line 63: NIT: It's better not to change the sequence number of the existing variables.

shuo-wu (commented):
Will this lead to any upgrade compatibility issues?

pkg/meta/version.go
```diff
@@ -2,7 +2,7 @@ package meta
 
 const (
 	// CLIAPIVersion used to communicate with user e.g. longhorn-manager
-	CLIAPIVersion    = 5
+	CLIAPIVersion    = 6
```
innobead, line 5: note: This is not be effective after 1.3, because the proxy has been introduced. However, it's still needed for releases < 1.3.

app/cmd/add_replica.go
```diff
@@ -21,6 +22,10 @@ func AddReplicaCmd() cli.Command {
 				Name:  "restore",
 				Usage: "Set this flag if the replica is being added to a restore/DR volume",
 			},
+			cli.StringFlag{
```
innobead, line 25: For >= 1.3, need to add this client implementation to proxy in instance manager code.

derekbit:
> Will this lead to any upgrade compatibility issues?

Yes. If the volume using the old engine image is stopped and then restart, the longhorn-manager cannot start the engine process because of the incompatible arguments. Any advise for the compatibility issue?

proto/ptypes/controller.proto
innobead: Exactly. 

For proto, it uses the same design guidance for other native language like C++ or interface definition language. The number is meaningful for backward compatibility.

innobead (commented):
The first commit message is not right. 

`Set replica to ERR if the size is mismatched with expected size` instead?

innobead (changes requested):
Some feedbacks and need to resolve other comments as well. Thanks.

derekbit:
> The first commit message is not right.
> 
> `Set replica to ERR if the size is mismatched with expected size` instead?

Yes, typo.

app/cmd/add_replica.go
innobead: cc @c3y1huang 

YES, you are right, the integration testing still relies on engine CLIs. We should probably refactor this, but it's fine to keep it for now.

proto/ptypes/controller.proto
derekbit: Updated. Thank you.

app/cmd/add_replica.go
```diff
@@ -21,6 +22,10 @@ func AddReplicaCmd() cli.Command {
...
+				Name:  "vol-size",
```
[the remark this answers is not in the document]
derekbit: Updated. Thank you.

pkg/meta/version.go
derekbit: Still bump the version because some test cases still rely on the version number.

innobead: Yeah, I see, as discussed here, https://github.com/longhorn/longhorn-engine/pull/740#discussion_r963031031.

innobead, line 5: Because we bump the CLI version, need to update the engine upgrade test image after merge. Chris will help with that.

integration/common/core.py
```diff
@@ -372,10 +373,10 @@ def start_no_frontend_volume(grpc_c, *grpc_r_list):
     grpc_c.volume_frontend_start(FRONTEND_TGT_BLOCKDEV)
 
     for grpc_r in grpc_r_list:
-        open_replica(grpc_r)
+        open_replica(grpc_r, SIZE_STR)
```
innobead, line 376: I found `def open_replica(grpc_client, backing_file=None)` did not change and the second arg is `backing_file`.

Is this correct?

derekbit: Ah, forgot updating one of the two open_replica().

integration/core/test_cli.py
```diff
@@ -106,18 +106,18 @@ def getNow():
     return datetime.datetime.utcnow().replace(microsecond=0).isoformat()
 
 
-def open_replica(grpc_client):
+def open_replica(grpc_client, size):
```
innobead, line 109: There are 2 `open_replica` (one in common, and the other in core package) . Should we consolidate and keep one only?

innobead (commented):
A few comments about testing, and suggest having the default value for `open_replica` or others w/ a new size arg, so you don't need to modify the test functions. WDYT? 

```python
open_replica(rpc_client, backing_file=None, size=SIZE_STR)
```

Ideally, specifying arguments explicitly is good for code maintenance, but this is for testing, so I think it's fine to have implicit default values.

pkg/meta/version.go
innobead: cc @longhorn/qa

integration/core/test_cli.py
derekbit: > A few comments about testing, and suggest having the default value for `open_replica` or others w/ a new size arg, so you don't need to modify the test functions. WDYT?
> 
> ```python
> open_replica(grpc_replica2)
> ```

Yes, good idea.

derekbit:
> A few comments about testing, and suggest having the default value for `open_replica` or others w/ a new size arg, so you don't need to modify the test functions. WDYT?
> 
> ```python
> open_replica(rpc_client, backing_file=None, size=SIZE_STR)
> ```

Yes, good idea.

integration/core/test_cli.py
derekbit: Merged the two functions.

integration/common/core.py
derekbit: Updated. Thank you.

joshimoo:
> the longhorn-manager cannot start the engine process because of the incompatible arguments. 

As a hack we used env vars instead of new parameters, but what we really need to do is change the engine binaries so that we can pass arbitrary parameters which are ignored instead of just terminating the process. 

If we fix this first, then in >= v1.4 we can fix a bunch of the env hacks we use as well.

We can also eventually make engines <X incompatible with longhorn-manager Y cc @innobead @shuo-wu 
A good candidate is to make any non proxy engine incompatible once we reach either v1.4 v1.5 that will allow us to remove a bunch of code (backwards compatibility etc for engine/instance-manager)

innobead:
> > Will this lead to any upgrade compatibility issues?
> 
> Yes. If the volume using the old engine image is stopped and then restart, the longhorn-manager cannot start the engine process because of the incompatible arguments. Any advise for the compatibility issue?

I missed this part.

It would be better to keep backward compatible; otherwise, users need to upgrade the engine image of the volume to start the engine. This will be an interrupt behavior. (we have a similar compatible issue before about backup creation w/ a provided backup name.)

Another way is to simulate compatible interface between versions to make any new added command options as optional, so if they are not provided, it will be fall back to the current behavior. Of course, this is not good enough in my view because I like strong contract instead.

innobead:
> > the longhorn-manager cannot start the engine process because of the incompatible arguments.
> 
> As a hack we used env vars instead of new parameters, but what we really need to do is change the engine binaries so that we can pass arbitrary parameters which are ignored instead of just terminating the process.

Agree, also if all communication can be over gRPC instead of CI, would be much naturally to do version compatibilty. 

In this case, @derekbit can we make the manager backward compatible with the old engine image when creating an engine process? at least, it should know what engine image it will use for that volume. 

Having backward compatibility is quite important for us if we would like to provide a non-interruptive user experience.

> 
> If we fix this first, then in >= v1.4 we can fix a bunch of the env hacks we use as well.

@joshimoo Could you elaborate what hacks we have now?

derekbit:
> > > the longhorn-manager cannot start the engine process because of the incompatible arguments.
> > 
> > 
> > As a hack we used env vars instead of new parameters, but what we really need to do is change the engine binaries so that we can pass arbitrary parameters which are ignored instead of just terminating the process.
> 
> Agree, also if all communication can be over gRPC instead of CLI (we use to create engine and replica process), would be much naturally to do version compatibilty.
> 
> In this case, @derekbit can we make the manager backward compatible with the old engine image when creating an engine process? at least, it should know what engine image it will use for that volume. WDYT?
> 
> Having backward compatibility is quite important for us if we would like to provide a non-interruptive user experience.
> 
> > If we fix this first, then in >= v1.4 we can fix a bunch of the env hacks we use as well.
> 
> @joshimoo Could you elaborate what hacks we have now?

Yes, I'm working on the compatibility issue.
I have one question. What do these mean? Especially, the xxxMinVersion.
```
InstanceManager.status.apiMinVersion
InstanceManager.status.apiVersion
InstanceManager.status.proxyApiMinVersion
InstanceManager.status.proxyApiVersion
```
cc @c3y1huang

PhanLe1010:
> Agree, also if all communication can be over gRPC instead of CLI (we use to create engine and replica process), would be much naturally to do version compatibility.

@innobead  Could you give some more detail about how it is much more natural to do version compatibility once all communication use gRPC?

innobead:
for @PhanLe1010 

Basically just use the protocol message compatibility characteristic.

"Protocol buffers provide a language-neutral, platform-neutral, extensible mechanism for serializing structured data in a forward-compatible and backward-compatible way." 

ref: https://developers.google.com/protocol-buffers/docs/overview

derekbit:
BTW, will we have any api/cli version deprecation roadmap?

derekbit:
The backward compatibility issue is addressed. Please help review the PR. Thank you.

shuo-wu (approved):

pkg/controller/control.go
```diff
@@ -613,25 +631,28 @@ func (c *Controller) Start(addresses ...string) error {
 
 	c.reset()
 
-	var fatalErr error
 	availableBackends := map[string]types.Backend{}
+	backendSizes := map[int64]struct{}{}
```
innobead, line 635: Why not just use list?

pkg/controller/control.go
```diff
@@ -642,20 +663,33 @@ func (c *Controller) Start(addresses ...string) error {
...
-			break
-		} else if c.sectorSize != newSectorSize {
-			availableBackends = map[string]types.Backend{}
-			fatalErr = fmt.Errorf("BUG: Backend sector sizes do not match %d != %d in the engine initiation phase", c.sectorSize, newSectorSize)
-			break
+		}
+
+		if c.sectorSize != newSectorSize {
+			logrus.Warnf("backend %v sector size does not match %d != %d in the engine initiation phase", address, c.sectorSize, newSectorSize)
+			continue
```
innobead, line 673: Do you know why the previous behavior will just error out if any backend is not matched? The new change looks sense to me by just bypass instead.

integration/common/cmd.py
```diff
@@ -3,8 +3,10 @@
 import subprocess
 from os import path
 
-from common.constants import RETRY_COUNTS, RETRY_INTERVAL
-
+from common.constants import (  # NOQA
```
innobead, line 6: Why adding `# NOQA` here?

pkg/frontend/socket/frontend.go
```diff
@@ -130,7 +130,7 @@ func (t *Socket) handleServerConnection(c net.Conn, rw types.ReaderWriterAt) {
 	server := dataconn.NewServer(c, NewDataProcessorWrapper(rw))
 	logrus.Infoln("New data socket connection established")
 	if err := server.Handle(); err != nil && err != io.EOF {
-		logrus.Errorln("Fail to handle socket server connection due to ", err)
+		logrus.Errorln("Failed to handle socket server connection due to ", err)
```
innobead, line 133: It's good to have a consistent pattern for messages 👍

innobead (changes requested):
In general, LGTM except a few comments required to fix.

integration/common/cmd.py
derekbit: It's not necessary here. Will remove it.

pkg/controller/control.go
derekbit: Well, I'm not really sure why the check was not implemented. I guess it might be due to that the case rarely happened and just for the simplicity.
=

pkg/controller/control.go
derekbit: Because I just want to check how many different sizes in the backens. Using list needs to iterate the values in the list for every new size.

pkg/controller/control.go
innobead: cc @shuo-wu @joshimoo

integration/common/cmd.py
derekbit: Updated. Thank you.

pkg/controller/control.go
```diff
@@ -595,7 +595,25 @@ func checkDeuplicteAddress(addresses ...string) error {
 	return nil
 }
 
-func (c *Controller) Start(addresses ...string) error {
+func determineCorrectVolumeSize(volumeSize, volumeCurrentSize int64, backendSizes map[int64]struct{}) int64 {
```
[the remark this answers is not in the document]
derekbit: Added the unit test for determineCorrectVolumeSize. Thank you.

derekbit:
All comments are addressed. It is ready for review. Thank you.
cc @innobead

innobead (approved):

PR #742, opened 2022-09-13
Unmap or Filesystem Trim Support

shuo-wu:
Longhorn/longhorn#836


innobead:
@shuo-wu Please update this outdated PR, then we will start the review soon.

shuo-wu:
Need to wait for https://github.com/longhorn/longhorn-engine/pull/762 merge before updating this PR

app/cmd/controller.go
```diff
@@ -85,6 +90,7 @@ func startController(c *cli.Context) error {
 	isUpgrade := c.Bool("upgrade")
 	disableRevCounter := c.Bool("disableRevCounter")
 	salvageRequested := c.Bool("salvageRequested")
+	unmapMarkSnapChainRemoved := c.Bool("unmap-mark-snap-chain-removed")
```
derekbit, line 93: Does longhorn-engine need to update the cli version due to the change?

pkg/types/types.go
```diff
@@ -36,13 +36,23 @@ const (
...
+type ReaderWriterUnmapperAt interface {
+	io.ReaderAt
+	io.WriterAt
+	UnmapperAt
+}
+
 type ReaderWriterAt interface {
 	io.ReaderAt
 	io.WriterAt
 }
```
derekbit, line 48: `ReaderWriterAt` is not useful. Why not just add `UnmapperAt` in the original `ReaderWriterAt` interface?

pkg/replica/client/client.go
```diff
@@ -378,6 +379,23 @@ func (c *ReplicaClient) SetRebuilding(rebuilding bool) error {
...
+	if err != nil {
+		return err
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), GRPCServiceCommonTimeout)
+	defer cancel()
+
+	if _, err := replicaServiceClient.UnmapMarkDiskChainRemovedSet(ctx, &ptypes.UnmapMarkDiskChainRemovedSetRequest{
+		Enabled: enabled,
+	}); err != nil {
+		return fmt.Errorf("failed to set UnmapMarkDiskChainRemoved flag to %v for replica %v: %v", enabled, c.replicaServiceURL, err)
```
derekbit, line 393: Use `errors.Wrapf()` instead.

pkg/controller/client/controller_client.go
```diff
@@ -210,6 +211,20 @@ func (c *ControllerClient) VolumeFrontendShutdown() error {
...
 
+func (c *ControllerClient) VolumeUnmapMarkSnapChainRemovedSet(enabled bool) error {
+	controllerServiceClient := c.getControllerServiceClient()
+	ctx, cancel := context.WithTimeout(context.Background(), GRPCServiceTimeout)
+	defer cancel()
+
+	if _, err := controllerServiceClient.VolumeUnmapMarkSnapChainRemovedSet(ctx, &ptypes.VolumeUnmapMarkSnapChainRemovedSetRequest{
+		Enabled: enabled,
+	}); err != nil {
+		return fmt.Errorf("failed to set UnmapMarkSnapChainRemoved to %v for volume %v: %v", enabled, c.serviceURL, err)
```
derekbit, line 222: `errors.Wrapf()`

app/cmd/volume.go
```diff
@@ -36,6 +36,27 @@ func ExpandCmd() cli.Command {
...
+	return cli.Command{
+		Name:      "unmap-mark-snap-chain-removed",
+		ShortName: "unmap-mark-snap",
+		Flags: []cli.Flag{
+			cli.BoolFlag{
+				Name: "enable",
+			},
+			cli.BoolFlag{
+				Name: "disable",
+			},
```
derekbit, line 49: After enabling the flag, will the following newly created snapshots be marked as `removed`?

pkg/backend/remote/remote.go
```diff
@@ -228,6 +228,37 @@ func (r *Remote) GetState() (string, error) {
...
+	replicaInfo, err := r.info()
+	if err != nil {
+		return false, err
+	}
+
+	return replicaInfo.UnmapMarkDiskChainRemoved, nil
+}
+
+func (r *Remote) SetUnmapMarkSnapChainRemoved(enabled bool) error {
+	logrus.Infof("Set UnmapMarkSnapChainRemoved of %s to : %v", r.name, enabled)
```
derekbit, line 241:
```suggestion
	logrus.Infof("Setting UnmapMarkSnapChainRemoved of %s to : %v", r.name, enabled)
```

pkg/backend/remote/remote.go
```diff
@@ -228,6 +228,37 @@ func (r *Remote) GetState() (string, error) {
...
+	defer conn.Close()
+	replicaServiceClient := ptypes.NewReplicaServiceClient(conn)
+
+	ctx, cancel := context.WithTimeout(context.Background(), client.GRPCServiceCommonTimeout)
+	defer cancel()
+
+	if _, err := replicaServiceClient.UnmapMarkDiskChainRemovedSet(ctx, &ptypes.UnmapMarkDiskChainRemovedSetRequest{
+		Enabled: enabled,
+	}); err != nil {
+		return fmt.Errorf("failed to set UnmapMarkDiskChainRemoved to %v for replica %v from remote: %v", enabled, r.replicaServiceURL, err)
```
derekbit, line 256: `errors.Wrapf()`

app/cmd/controller.go
shuo-wu: IIRC, you already bump up the version for v1.4.0 in other PRs, then I don't need to handle that again.

derekbit: Got it. Thank you.

pkg/types/types.go
shuo-wu: Sure. we can directly modify `ReaderWriterAt` to `ReaderWriterUnmapperAt`

app/cmd/volume.go
shuo-wu: If there is an unmap request immediately after the snapshot creation, then the answer is YES.
I guess you are worried about the backup creation. Yes this may lead to the following backup creation fail.

shuo-wu: Actually, I don't feel comfortable with the separation of the sync agent server and the replica server now. Due to the newly introduced feature like unmap or prune, the removed snapshot is not immutable anymore hence the rebuilding flow should be aware of it and be protected by the replica lock. Otherwise, things may mess up...

pkg/controller/multi_unmapper_at.go
```diff
@@ -0,0 +1,78 @@
...
+			n, err := u.UnmapAt(length, off)
+			if err != nil {
+				errs[index] = err
+			}
+
+			m.lock.Lock()
+			if unmappedSize == 0 {
+				unmappedSize = n
+			} else if unmappedSize != n {
+				logrus.Warnf("One of the MultiUnmapper get different size %v from others %v", n, unmappedSize)
```
shuo-wu, line 57: Not sure if we need to record this as an error.

app/cmd/volume.go
shuo-wu: Added a TODO comment in `func (c *Controller) UnmapAt(length uint32, off int64) (int, error)`

pkg/controller/multi_unmapper_at.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+	}
+
+	switch len(errors) {
+	case 0:
+		return "Unknown"
+	case 1:
+		return errors[0]
+	default:
+		return strings.Join(errors, "; ")
```
innobead, line 35: You can just keep `default`.

pkg/controller/multi_unmapper_at.go
```diff
@@ -0,0 +1,78 @@
...
+				errs[index] = err
+			}
+
+			m.lock.Lock()
+			if unmappedSize == 0 {
+				unmappedSize = n
+			} else if unmappedSize != n {
+				logrus.Warnf("One of the MultiUnmapper get different size %v from others %v", n, unmappedSize)
+			}
+			m.lock.Unlock()
```
innobead, line 59: `defer func() { m.lock.Unlock(); wg.Done() }` just next to `m.lock.Lock()` instead?

pkg/controller/multi_unmapper_at.go
```diff
@@ -0,0 +1,78 @@
...
+
+			wg.Done()
+		}(i, u)
+	}
+
+	wg.Wait()
+
+	var err error = nil
+	for i := range errs {
+		if errs[i] != nil {
```
innobead, line 69: You already make sure `errs` will not have nil value above, so can just create and return a `MultiUnmapperError` if `len(errs) > 0` instead of using this needless loop.

pkg/controller/replicator.go
```diff
@@ -143,16 +145,42 @@ func (r *replicator) WriteAt(p []byte, off int64) (int, error) {
...
-func (r *replicator) buildReadWriters() {
+func (r *replicator) UnmapAt(length uint32, off int64) (int, error) {
+	if !r.backendsAvailable {
+		return 0, ErrNoBackend
+	}
+
+	n, err := r.unmapper.UnmapAt(length, off)
+	if err != nil {
+		errs := map[string]error{
+			r.unmapperIndex[0]: err,
```
innobead, line 156: Should we need index check for `unmapperIndex[0]` or `unmapperIndex[index]` below to avoid any unexpected out of index issue? (I know you do the same as readerIndex and writeIndex)

proto/ptypes/replica.proto
```diff
@@ -24,6 +24,8 @@ service ReplicaService {
   rpc RebuildingSet(RebuildingSetRequest) returns (RebuildingSetResponse) {}
   rpc RevisionCounterSet(RevisionCounterSetRequest) returns
     (RevisionCounterSetResponse) {}
+rpc UnmapMarkDiskChainRemovedSet(UnmapMarkDiskChainRemovedSetRequest) returns
```
innobead, line 27: Wrong indent, need to reformat.

pkg/replica/replica.go
```diff
@@ -1097,6 +1099,53 @@ func (r *Replica) ReadAt(buf []byte, offset int64) (int, error) {
...
+				return 0, errors.Wrapf(err, "failed to mark snapshot disk %v as removed before unmap", disk.Name)
+			}
+		}
+		if disk.Removed {
+			unmappableDisks = append(unmappableDisks, r.diskPath(disk.Name))
+		}
+	}
+	r.info.Dirty = true
+	unmappedSize, err := r.volume.UnmapAt(unmappableDisks, length, offset)
+	r.Unlock()
```
innobead, line 1152: Should move `r.Unlock()` just next to `r.Lock()` to avoid some error return before `r.Unlock()`

If you don't want to unlock the lock until the end of this method, you can use an anonymous closure to run the above unmap operations and do the lock/unlock inside.

innobead (changes requested):
In general, LGTM. Some comments need to resolve.

app/cmd/replica.go
```diff
@@ -61,6 +61,11 @@ func ReplicaCmd() cli.Command {
 				Hidden: false,
 				Usage:  "To disable revision counter for every write",
 			},
+			cli.BoolFlag{
+				Name:   "unmap-mark-disk-chain-removed",
```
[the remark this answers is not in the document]
innobead: Should we make it consistent? It seems in replica, we use disk for snapshot. Given the context, I feel it should be fine for now.

pkg/controller/control.go
```diff
@@ -825,6 +825,35 @@ func (c *Controller) ReadAt(b []byte, off int64) (int, error) {
 	return n, err
 }
 
+func (c *Controller) UnmapAt(length uint32, off int64) (int, error) {
+	// TODO: Need to fail unmap requests
+	//  if the volume is purging snapshots or creating backups.
+	c.RLock()
+	if off < 0 || off+int64(length) > c.size {
+		err := fmt.Errorf("EOF: Unmap of %v bytes at offset %v is beyond volume size %v", length, off, c.size)
+		c.RUnlock()
```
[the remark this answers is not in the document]
shuo-wu: Yeah. Since we don't need to call `c.recordMetrics()` for this function. Directly using `defer ` is good enough.

pkg/controller/control.go
```diff
@@ -825,6 +825,35 @@ func (c *Controller) ReadAt(b []byte, off int64) (int, error) {
...
+	if c.hasWOReplica() {
+		c.RUnlock()
+		return 0, fmt.Errorf("can not unmap volume when there is WO replica")
+	}
+	if c.isExpanding {
+		c.RUnlock()
+		return 0, fmt.Errorf("can not unmap volume during expansion")
+	}
+
+	// startTime := time.Now()
```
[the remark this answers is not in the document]
shuo-wu: Currently the unmap operation is not included in the metrics. I will add a TODO anyway for it.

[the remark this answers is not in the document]
innobead: Sounds good, but suggest creating a ticket to avoid missing around.

pkg/controller/control.go
```diff
@@ -825,6 +825,35 @@ func (c *Controller) ReadAt(b []byte, off int64) (int, error) {
 	return n, err
 }
 
+func (c *Controller) UnmapAt(length uint32, off int64) (int, error) {
+	// TODO: Need to fail unmap requests
+	//  if the volume is purging snapshots or creating backups.
+	c.RLock()
```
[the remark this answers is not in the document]
shuo-wu: For the backup creation, if some removed files are unloaded to the remote target before the trim reclaiming it, the backup would contain some garbage data. And the checksum of a volume restored from this backup will be different from the original volume (after the trim). But I think this difference is acceptable.

For the purge, some blocks of the removed files may be retained. I am not sure if it's possible.

In general, it would not crash the data hence I didn't resolve this issue in the 1st version of the trim feature.

pkg/controller/multi_unmapper_at.go
shuo-wu: Got it. BTW, I switch to creating a lock inside this function instead of storing the lock in the struct.

pkg/controller/replicator.go
shuo-wu: Ideally, the controller should error out once there is no available replica (while it is still running and receiving the requests)
But just in case, we can add it.

pkg/replica/replica.go
shuo-wu: Modified. Using an anonymous closure instead.

pkg/controller/multi_unmapper_at.go
innobead: You can skip `case 1` because it will be just part of `default`.

pkg/controller/control.go
```diff
@@ -825,6 +885,36 @@ func (c *Controller) ReadAt(b []byte, off int64) (int, error) {
...
+	if c.hasWOReplica() {
+		return 0, fmt.Errorf("can not unmap volume when there is WO replica")
+	}
+	if c.isExpanding {
+		return 0, fmt.Errorf("can not unmap volume during expansion")
+	}
+
+	// startTime := time.Now()
+	n, err := c.backend.UnmapAt(length, off)
+	c.RUnlock()
```
innobead, line 907: This `c.RUnlock` is not needed.

pkg/controller/multi_unmapper_at.go
shuo-wu: Got it. Updated.

innobead (changes requested):

pkg/controller/multi_unmapper_at.go
shuo-wu: I cannot apply this change: `len(errs)` is always non-zero value: `errs := make([]error, len(m.unmappers))`. 
As a result, some of the entries are indeed nil.

innobead: oh, I see, then need to filter.

pkg/replica/replica.go
innobead: :point_up: this is good

innobead (approved):

innobead (approved):

PR #745, opened 2022-09-20
Fix data corruption due to COW and block size not being aligned during rebuilding replicas

PhanLe1010:
During rebuilding (while we have WO replica), if the incoming IO is not aligned, we read full 4096 bytes blocks from RW replicas then write that to all replicas. This should support all kinds of filesystem logical block sizes well as block volume mode.

longhorn/longhorn#4354

### DIFF
--- a/pkg/controller/control.go
+++ b/pkg/controller/control.go
@@ -19,6 +19,8 @@ import (
 	"github.com/longhorn/longhorn-engine/pkg/util"
 )
 
+const defaultSectorSize int64 = 4096
+
 type Controller struct {
 	sync.RWMutex
 	Name       string
@@ -750,13 +752,43 @@ func (c *Controller) WriteAt(b []byte, off int64) (int, error) {
 		return 0, err
 	}
 	startTime := time.Now()
-	n, err := c.backend.WriteAt(b, off)
+	var n int
+	var err error
+	if c.hasWOReplica() {
+		n, err = c.writeInWOMode(b, off)
+	} else {
+		n, err = c.writeInNormalMode(b, off)
+	}
 	c.RUnlock()
 	if err != nil {
 		return n, c.handleError(err)
 	}
 	c.recordMetrics(false, l, time.Since(startTime))
 	return n, err
+}
+
+func (c *Controller) writeInWOMode(b []byte, off int64) (int, error) {
+	bufLen := int64(len(b))
+	// buffer b is defaultSectorSize aligned
+	if (bufLen >= defaultSectorSize) && (off%defaultSectorSize == 0) && ((off+bufLen)%defaultSectorSize == 0) {
+		return c.backend.WriteAt(b, off)
+	}
+
+	readOffsetStart := (off / defaultSectorSize) * defaultSectorSize
+	readOffsetEnd := (((off + bufLen) / defaultSectorSize) + 1) * defaultSectorSize
+	readBuf := make([]byte, readOffsetEnd-readOffsetStart)
+	if _, err := c.backend.ReadAt(readBuf, readOffsetStart); err != nil {
+		return 0, errors.Wrap(err, "failed to retrieve aligned sectors from RW replicas")
+	}
+
+	startCut := off % defaultSectorSize
+	copy(readBuf[startCut:startCut+bufLen], b)
+
+	return c.backend.WriteAt(readBuf, readOffsetStart)
+}
+
+func (c *Controller) writeInNormalMode(b []byte, off int64) (int, error) {
+	return c.backend.WriteAt(b, off)
 }
 
 func (c *Controller) ReadAt(b []byte, off int64) (int, error) {


innobead:
This is creative and thinks from the opposite 👍 !

pkg/controller/control.go
```diff
@@ -759,6 +765,31 @@ func (c *Controller) WriteAt(b []byte, off int64) (int, error) {
 	return n, err
 }
 
+func (c *Controller) writeInWOMode(b []byte, off int64) (int, error) {
+	sectorSize := int64(4096)
```
innobead, line 769: use `defaultSectorSize` const for this value instead

pkg/controller/control.go
```diff
@@ -759,6 +765,31 @@ func (c *Controller) WriteAt(b []byte, off int64) (int, error) {
...
+	// buffer b is sectorSize aligned
+	if (bufLen >= sectorSize) && (off%sectorSize == 0) && ((off+bufLen)%sectorSize == 0) {
+		return c.backend.WriteAt(b, off)
+	}
+
+	readOffsetStart := (off / sectorSize) * sectorSize
+	readOffsetEnd := (((off + bufLen) / sectorSize) + 1) * sectorSize
+	readBuf := make([]byte, readOffsetEnd-readOffsetStart)
+	if _, err := c.backend.ReadAt(readBuf, readOffsetStart); err != nil {
+		return 0, err
```
innobead, line 780: Suggest an enriching message with this err like `failed to retrieve aligned sectors from RW replicas..`.

pkg/controller/control.go
```diff
@@ -759,6 +765,31 @@ func (c *Controller) WriteAt(b []byte, off int64) (int, error) {
...
+	sectorSize := int64(4096)
+	bufLen := int64(len(b))
+	// buffer b is sectorSize aligned
+	if (bufLen >= sectorSize) && (off%sectorSize == 0) && ((off+bufLen)%sectorSize == 0) {
+		return c.backend.WriteAt(b, off)
+	}
+
+	readOffsetStart := (off / sectorSize) * sectorSize
+	readOffsetEnd := (((off + bufLen) / sectorSize) + 1) * sectorSize
+	readBuf := make([]byte, readOffsetEnd-readOffsetStart)
```
innobead, line 778: Think if it's possible to reuse buffer here instead of reallocation.

innobead (approved):
In general, LGTM.

See others feedback.

pkg/controller/control.go
PhanLe1010: Each read will potentially require a different readBuff length. I am not sure how to reuse this one

pkg/controller/control.go
PhanLe1010: Thank you. Fixed

pkg/controller/control.go
PhanLe1010: Thank you. Fixed

PhanLe1010:
> Also, if we can add some unit tests to cover the block size calculation to determine if needs alignment correction before writing it down to the downstream replicas, that will be great.

Surely, I will work on that

pkg/controller/control.go
```diff
@@ -19,6 +19,8 @@ import (
 	"github.com/longhorn/longhorn-engine/pkg/util"
 )
 
+const defaultSectorSize int64 = 4096
```
derekbit, line 22: nit: many `defaultSectorSize` longhorn-engine codes, we can reorganize them in the future.

derekbit (approved):
LGTM

pkg/controller/control.go
shuo-wu: This function needs to read the unaligned part (of which the length is smaller than the `defaultSectorSize`) only, hence a buffer with a fixed length `defaultSectorSize` should be enough.

pkg/controller/control.go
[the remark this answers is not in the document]
innobead: @PhanLe1010 could you create a specific ticket for this? it's a later work. Thanks.

pkg/controller/control.go
PhanLe1010: 1. Since this function `writeInWOMode` is called concurrently by multiple go routines, I think it is not safe to pre-allocated and reuse the buffer
2. Suppose that the incoming IO is from offset 4000 to 9000, if we use fixed length 4096 for the readBuff, we would have to do 3 read IOs (0-4095, 4096-8191, 8192-12287) then followed by 3 write IOs. These read IO potentially will have to go over the network. The benefit is we save a very tiny RAM. Therefore, I think we shouldn't use fixed 4096 length for the readBuff. What do you think?

pkg/controller/control.go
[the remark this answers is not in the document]
PhanLe1010: https://github.com/longhorn/longhorn/issues/4599

pkg/controller/control.go
PhanLe1010: Ticket created https://github.com/longhorn/longhorn/issues/4599. Thanks

pkg/controller/control.go
PhanLe1010: ping @innobead and @shuo-wu

innobead: Agreed 👍

innobead (approved):
LGTM.

PR #746, opened 2022-09-21
Support customized iSCSI LUN block size

derekbit:
https://github.com/longhorn/longhorn/issues/4599

**TODO: update the go.mod go-iscsi-helper after review**

Signed-off-by: Derek Su [derek.su@suse.com](mailto:derek.su@suse.com)


pkg/sync/sync.go
```diff
@@ -340,14 +340,14 @@ func find(list []string, item string) int {
...
+func (t *Task) AddRestoreReplica(volumeSize, volumeCurrentSize, sectorSize int64, replica string) error {
 	volume, err := t.client.VolumeGet()
 	if err != nil {
 		return err
 	}
 
 	if volume.ReplicaCount == 0 {
-		return t.client.VolumeStart(volumeSize, volumeCurrentSize, replica)
+		return t.client.VolumeStart(volumeSize, volumeCurrentSize, sectorSize, replica)
 	}
```
shuo-wu, line 351: Actually, I don't think API `AddReplica` requires input `blockSize`. Instead, `t.client.VolumeGet()` should tell the info. (Maybe similar logic can be applied to the volume size inputs.)

derekbit: Becasue engine can call `add-replica` directly, the `volume.size` and `volume.blockSize` wiil be zero and lead to the failure of volume/replica size check in this case. So, I add the `volume.size` and `volume.blockSize` here.

shuo-wu: In the example you mentioned, the size recorded in the engine process should be the volume spec size rather than 0. ( the caller already tells the size during `VolumeStart`.) Is there anything I missed?

derekbit: Weird, I printed the value volume.size, and it showed 0.

shuo-wu: Hmmm. I am not sure if I misunderstand the volume-size-related modification you made last time. Will the engine process use `0` volume size during starting if there is no replica?

derekbit: > Will the engine process use 0 volume size during starting if there is no replica?

Yes, engine process uses 0 if there is no replica.
I also tested v1.3.1 (without my volume-size-related modification last time), and the size is still 0.

derekbit: @shuo-wu Any further concern for the implementation?

shuo-wu: I was thinking we can get a valid volume size from a running engine anyway but in fact NO. I may need to re-check the previous PR.
But for the block size part, all running engines (no matter if there are connected running replicas or not) should tell a valid value, right?

derekbit: > But for the block size part, all running engines (no matter if there are connected running replicas or not) should tell a valid value, right?

I'm not really sure for this part, because the size is set when there is at least one replica in original implementaton https://github.com/longhorn/longhorn-engine/blob/v1.2.x/pkg/controller/control.go#L657. Is it unexpected?

shuo-wu: I just checked the code:
Is `volumeSize` the volume spec size [here](https://github.com/longhorn/longhorn-engine/blob/bcde63a5bf2babdfe4fb90f1c049b69babdab134/pkg/controller/control.go#L601-L605)? If No, then there is no doubt that we have to input 2 size fields for API `AddReplica`.

derekbit: Yes. But the `Start()` will exit early if there is no replica ([code](https://github.com/longhorn/longhorn-engine/blob/bcde63a5bf2babdfe4fb90f1c049b69babdab134/pkg/controller/control.go#L622)), so the `c.size` won't be set

shuo-wu: OK... Do you think this makes sense? If it is fine for you, we just need to focus on the block size then.

derekbit: Hmm, not sure the purpose of the original design, but it looks make sense in the logic.

derekbit: I updated the block size part. You can take a look. Thank you.

pkg/util/util.go
```diff
@@ -30,9 +30,11 @@ var (
 	MaximumVolumeNameSize = 64
 	validVolumeName       = regexp.MustCompile(`^[a-zA-Z0-9][a-zA-Z0-9_.-]+$`)
 
-	cmdTimeout = time.Minute // one minute by default
-
 	HostProc = "/host/proc"
+
+	DefaultVolumeSectorSize  = int64(4096)
+	DefaultReplicaSectorSize = int64(512)
```
shuo-wu, line 36: I am thinking what value we should use for replica sector size... Still 512?

derekbit: 512 is fine in my test, but I cannot figure out why we need a 512 replica sector size before from the codes. Do you have any idea?

shuo-wu: No for me either. Actually, I prefer to use 4k here.

PR #747, opened 2022-09-22
[BACLKPORT][v1.3.2] Set replica to ERR if the size is mismatched with expected size

derekbit:
https://github.com/longhorn/longhorn/issues/3601
https://github.com/longhorn/longhorn/issues/4609

Signed-off-by: Derek Su [derek.su@suse.com](mailto:derek.su@suse.com)


innobead (approved):
LGTM

PR #759, opened 2022-09-29
Fix wrong aligment calculation and written bytes returned and add unit test for writeInWOMode

PhanLe1010:
longhorn/longhorn#4354

### DIFF
--- a/pkg/controller/control.go
+++ b/pkg/controller/control.go
@@ -19,8 +19,6 @@ import (
 	"github.com/longhorn/longhorn-engine/pkg/util"
 )
 
-const defaultSectorSize int64 = 4096
-
 type Controller struct {
 	sync.RWMutex
 	Name       string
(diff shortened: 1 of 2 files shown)


pkg/controller/control.go
```diff
@@ -768,23 +768,34 @@ func (c *Controller) WriteAt(b []byte, off int64) (int, error) {
 }
 
 func (c *Controller) writeInWOMode(b []byte, off int64) (int, error) {
-	bufLen := int64(len(b))
+	bufLen := len(b)
 	// buffer b is defaultSectorSize aligned
-	if (bufLen >= defaultSectorSize) && (off%defaultSectorSize == 0) && ((off+bufLen)%defaultSectorSize == 0) {
+	if (bufLen >= defaultSectorSize) && (off%defaultSectorSize == 0) && ((off+int64(bufLen))%defaultSectorSize == 0) {
 		return c.backend.WriteAt(b, off)
 	}
```
joshimoo, line 775: NIT: This can be simplified, since if bufLen == 0, we can either let it pass down or exit early here. Since I am not sure that there is any reason in continuing with the below operations in the case where the write has a length of 0?

Explanation:
off%defaultSectorSize == 0 --> implies that offset is a multiple of sector size
bufLen % defaultSectorSize == 0 --> implies that len is a multiple of sector size
therefore the sum of (len + offset) must also be a multiple of sector size.

So this can be simplified to:
(off%defaultSectorSize == 0) && (bufLen%defaultSectorSize == 0)

joshimoo: The simplification isn't as important, but I am concerned about the buflen == 0 case which would trigger the rmw cycle instead of the direct write. Since a quick check didn't show an exit condition.

joshimoo (commented):
just double checking the math at the moment.
are we covering the:

pre
4k
post

case?
Right now it seems that we read the complete offset + buffer length.
In the case where we have a 15mb

shuo-wu (dismissed):
In general LGTM

pkg/controller/control.go
```diff
@@ -768,23 +768,34 @@ func (c *Controller) WriteAt(b []byte, off int64) (int, error) {
...
 
 	readOffsetStart := (off / defaultSectorSize) * defaultSectorSize
-	readOffsetEnd := (((off + bufLen) / defaultSectorSize) + 1) * defaultSectorSize
+	var readOffsetEnd int64
+	if ((off + int64(bufLen)) % defaultSectorSize) == 0 {
```
derekbit, line 779: nit: we can save the result of `((off+int64(bufLen))%defaultSectorSize` in a variable and reuse it in the following check.

derekbit (dismissed):

pkg/controller/control.go
PhanLe1010: > pre-4k-post case?

Yeah, we have it 
```
	// Test case #3
	buf = makeByteSliceWithInitialData(4096+1024, newVal)
	off = 4096 - 512
```

> So this can be simplified to: (off%defaultSectorSize == 0) && (bufLen%defaultSectorSize == 0)

Thank you. I updated

> Test case len(buf) == 0 and off%defaulSectorSize==0 is added

```
	// Test case #8
	buf = makeByteSliceWithInitialData(0, newVal)
	off = 4096 * 3
```

derekbit (dismissed):
LGTM

pkg/controller/control_test.go
```diff
@@ -86,3 +91,208 @@ func (s *TestSuite) TestDetermineCorrectVolumeSize(c *C) {
...
+		case i < dataLength:
+			expectedData[i] = newVal
+		}
+	}
+	testsets = append(testsets, testCase{buf: buf, off: off, expectedData: expectedData})
+
+	// Test case #8
+	buf = makeByteSliceWithInitialData(0, newVal)
+	off = 4096 * 3
+	expectedData = makeByteSliceWithInitialData(dataLength, writeSourceInitVal)
```
shuo-wu, line 262: Isn't this `readSourceInitVal`?

PhanLe1010: I think no. the buff has 0 length. the fakeWriter.source shouldn't be touched so expected data remain as writeSourceInitVal

shuo-wu: Oh Yes. I misunderstood `writeSource` a little bit

shuo-wu (approved):

pkg/controller/control_test.go
PhanLe1010: But you are right that we would have problem in this new test case:
```
	// Test case #9
	buf = makeByteSliceWithInitialData(0, newVal)
	off = 4096 + 512
```

the fakeWriter.source shouldn't be touched therefore, I added the condition:

```
	if (bufLen == 0) || ((off%util.VolumeSectorSize == 0) && (bufLen%util.VolumeSectorSize == 0)) {
		return c.backend.WriteAt(b, off)
	}
```

LOL. unit test is very helpful

innobead (approved):
LGTM

shuo-wu (approved):

derekbit (approved):

joshimoo (approved):
Good job with the addition of the unit tests to validate the new logic :)

PR #762, opened 2022-10-04
Adjust the iSCSI initiator/target timeout and the engine-to-replica timeout

derekbit:
[Longhorn 4491](https://github.com/longhorn/longhorn/issues/4491)

### DIFF
--- a/app/cmd/controller.go
+++ b/app/cmd/controller.go
@@ -1,11 +1,11 @@
 package cmd
 
 import (
-	"errors"
-	"fmt"
 	"log"
+	"time"
 
 	"github.com/docker/go-units"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 
@@ -59,6 +59,11 @@ func ControllerCmd() cli.Command {
 				Hidden: false,
 				Usage:  "Start engine controller in a special mode only to get best replica candidate for salvage",
 			},
+			cli.Int64Flag{
+				Name:   "engine-replica-timeout",
+				Hidden: false,
+				Usage:  "In seconds. Timeout between engine and replica(s)",
+			},
 		},
 		Action: func(c *cli.Context) {
 			if err := startController(c); err != nil {
@@ -104,6 +109,11 @@ func startController(c *cli.Context) error {
 		return err
 	}
 
+	timeout := c.Int64("engine-replica-timeout")
+	engineReplicaTimeout := time.Duration(timeout) * time.Second
+	engineReplicaTimeout = controller.DetermineEngineReplicaTimeout(engineReplicaTimeout)
+	iscsiTargetRequestTimeout := controller.DetermineIscsiTargetRequestTimeout(engineReplicaTimeout)
+
 	factories := map[string]types.BackendFactory{}
 	for _, backend := range backends {
 		switch backend {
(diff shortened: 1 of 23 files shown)


pkg/controller/control.go
```diff
@@ -21,14 +21,16 @@ import (
...
+	Name                      string
+	size                      int64
+	sectorSize                int64
+	replicas                  []types.Replica
+	factory                   types.BackendFactory
+	backend                   *replicator
+	frontend                  types.Frontend
+	isUpgrade                 bool
+	iscsiTargetRequestTimeout int64
+	engineReplicaTimeout      time.Duration
```
innobead, line 33: Can we use `time.Duration` for all timeout properties to make type consistent? You can use `.Seconds()` to get the seconds later.

pkg/controller/init_frontend.go
```diff
@@ -1,21 +1,53 @@
...
 
-var (
-	Frontends = map[string]types.Frontend{}
+const (
+	defaultScsiTimeout       = 60 // SCSI device timeout, unit: seconds
+	defaultIscsiAbortTimeout = 15 // unit: seconds
+
+	defaultEngineReplicaTimeout = 8  // unit: seconds
+	minEngineReplicaTimeout     = 8  // unit: seconds
+	maxEngineReplicaTimeout     = 30 // unit: seconds
```
innobead, line 19: Probably we can just use `Time.Duration`, so we don't need to add these comments, but all of them are self-explanatory.

app/cmd/controller.go
```diff
@@ -104,6 +109,11 @@ func startController(c *cli.Context) error {
 		return err
 	}
 
+	engineReplicaTimeout := c.Int64("engine-replica-timeout")
+
+	engineReplicaTimeout = controller.DetermineEngineReplicaTimeout(engineReplicaTimeout)
+	iscsiTargetRequestTimeout := controller.DetermineIscsiTargetRequestTimeout(engineReplicaTimeout)
```
innobead, line 115: All related, so can put together w/o empty line.

```go
engineReplicaTimeout := c.Int64("engine-replica-timeout")
engineReplicaTimeout = controller.DetermineEngineReplicaTimeout(engineReplicaTimeout)
iscsiTargetRequestTimeout := controller.DetermineIscsiTargetRequestTimeout(engineReplicaTimeout)

```

pkg/controller/init_frontend.go
```diff
@@ -1,21 +1,53 @@
...
+		return tgt.New(devtypes.FrontendTGTISCSI, defaultScsiTimeout, defaultIscsiAbortTimeout, iscsiTargetRequestTimeout), nil
+	default:
+		return nil, fmt.Errorf("unsupported frontend type: %v", frontendType)
+	}
+}
+
+func DetermineEngineReplicaTimeout(timeout int64) int64 {
+	if timeout < minEngineReplicaTimeout ||
+		timeout > maxEngineReplicaTimeout {
+		return defaultEngineReplicaTimeout
```
innobead, line 41: Suggest a having a log about using the default timeout, because the value provided is not valid.

pkg/frontend/tgt/frontend.go
```diff
@@ -19,14 +19,17 @@ const (
...
 
-	isUp         bool
-	dev          longhorndev.DeviceService
-	frontendName string
+	isUp                      bool
+	dev                       longhorndev.DeviceService
+	frontendName              string
+	scsiTimeout               int64
+	iscsiAbortTimeout         int64
+	iscsiTargetRequestTimeout int64
```
innobead, line 27: Use `time.Duration`, then use the right method to convert to second.

innobead (changes requested):
LGTM, but some minor things needed to resolve.

pkg/controller/control.go
derekbit: Sure.

derekbit: Addressed. Thank you.

app/cmd/controller.go
derekbit: Addressed. Thank you.

pkg/frontend/tgt/frontend.go
derekbit: Addressed. Thank you.

pkg/controller/init_frontend.go
derekbit: Addressed. Thank you.

pkg/controller/init_frontend.go
derekbit: Addressed. Thank you.

innobead (approved):

PR #763, opened 2022-10-11
feat: Snapshot checksum calculation and verification

derekbit:
longhorn/longhorn#4210
longhorn/longhorn#3198


pkg/replica/hash.go
```diff
@@ -0,0 +1,220 @@
...
+
+func (t *SnapshotHashTask) hashSnapshot() (string, error) {
+	err := t.openSnapshot()
+	if err != nil {
+		logrus.Errorf("failed to open snapshot %v since %v", t.SnapshotNames, err)
+		return "", err
+	}
+	defer t.closeSnapshot()
+
+	h := sha256.New()
```
innobead, line 200: Do we consider using CRC instead of sha256 in terms of performance perspective? 
https://stackoverflow.com/a/996873

derekbit: CRC is a non-cryptographic hash algorithm and cannot provide collision resistance guarantee. This mean it is unable to detect the file corruption or content change in some cases. We can also use a faster cryptographic algorithm such as MD5 for the tradeoff between performance impact and detection capability.

Actually, I'm fine with the two solutions.

@joshimoo @shuo-wu @innobead WDYT

derekbit: IIRC, zfs uses CRC for block error detection. I think we can use CRC in the implementation here.

derekbit: Updated, thank you.

app/cmd/snapshot.go
```diff
@@ -166,6 +168,48 @@ func SnapshotCloneStatusCmd() cli.Command {
...
+		},
+	}
+}
+
+func SnapshotHashStatusCmd() cli.Command {
+	return cli.Command{
+		Name: "hash-status",
+		Flags: []cli.Flag{
+			cli.StringFlag{
+				Name:  "snapshot-name",
```
c3y1huang, line 201: Suggest using a command-line argument instead of as flag.

app/cmd/snapshot.go
```diff
@@ -415,3 +459,56 @@ func cloneSnapshotStatus(c *cli.Context) error {
...
+
+	names := c.StringSlice("snapshot-name")
+	if names == nil {
+		return errors.Wrap(err, "snapshot-name is required")
+	}
+
+	rehash := c.Bool("rehash")
+
+	concurrentLimit := c.Int("concurrent-limit")
+	if concurrentLimit < 0 {
```
c3y1huang, line 480: Will it make sense to auto-correct to `0` if `< 0`?

pkg/replica/hash.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+}
+
+func (t *SnapshotHashTask) openSnapshot() error {
+	t.Lock()
+	defer t.Unlock()
+
+	dir, err := os.Getwd()
+	if err != nil {
+		return errors.Wrapf(err, "cannot get working directory")
```
c3y1huang, line 113:
```suggestion
		return errors.Wrap(err, "cannot get working directory")
```

proto/ptypes/syncagent.proto
```diff
@@ -150,3 +153,20 @@ message SnapshotCloneStatusResponse {
...
+message SnapshotHashStatusRequest {
+  string snapshot_name = 1;
+}
+
+message SnapshotHashStatusResponse {
+  string state = 1;
+  int32 progress = 2;
+  string checksum = 3;
+  string error = 4;
+}
```
c3y1huang, line 172:
```suggestion
}

```

pkg/util/disk/disk.go
```diff
@@ -0,0 +1,56 @@
...
+	return map[string]string{
+		replicaExpansionLabelKey: strconv.FormatInt(size, 10),
+	}
+}
+
+func IsHeadDisk(diskName string) bool {
+	if strings.HasPrefix(diskName, VolumeHeadDiskPrefix) && strings.HasSuffix(diskName, VolumeHeadDiskSuffix) {
+		return true
+	}
+	return false
```
c3y1huang, line 55:
```suggestion
	return strings.HasPrefix(diskName, VolumeHeadDiskPrefix) && strings.HasSuffix(diskName, VolumeHeadDiskSuffix)
```

pkg/sync/rpc/server.go
```diff
@@ -1399,109 +1402,45 @@ func (s *SyncAgentServer) rmDisk(disk string) error {
...
-// remove deletes the object present at slice[index] and returns the remaining elements of slice yet maintaining
-// the original order of elements in the slice
-func (*BackupList) remove(b []*BackupInfo, index int) ([]*BackupInfo, error) {
-	if b == nil {
-		return nil, fmt.Errorf("empty list")
-	}
-	if index >= len(b) || index < 0 {
-		return nil, fmt.Errorf("BUG: attempting to delete an out of range index entry from backupList")
+func (s *SyncAgentServer) SnapshotHashStatus(ctx context.Context, req *ptypes.SnapshotHashStatusRequest) (*ptypes.SnapshotHashStatusResponse, error) {
+	if req.SnapshotName == "" {
```
c3y1huang, line 1431: Do we need to check this when it's also checked in `s.SnapshotHashList.Get`?

c3y1huang (changes requested):

app/cmd/snapshot.go
derekbit: Yeah, the auto-correctness is okay.

pkg/sync/rpc/server.go
shuo-wu: Do we need to get the status one by one? Unlike the backup creation, the snapshot hash can be handled in parallel.

pkg/sync/rpc/server.go
```diff
@@ -1505,3 +1529,137 @@ func (b *BackupList) Delete(backupID string) error {
...
+
+	status := hash.DoSnapshotHash(req.SnapshotNames, req.Rehash, req.ConcurrentLimit)
+
+	if err := s.SnapshotHashList.Delete(req.SnapshotNames[0]); err != nil {
+		return nil, errors.Wrapf(err, "failed to delete snapshot %v for from list", req.SnapshotNames[0])
+	}
+
+	if err := s.SnapshotHashList.Add(req.SnapshotNames[0], status); err != nil {
+		return nil, errors.Wrapf(err, "failed to add snapshot %v for hashing", req.SnapshotNames[0])
+	}
```
shuo-wu, line 1423: This part looks a little bit confusing with the first glance. Can we add a comment that describes how the syne agent server updates `SnapshotHashList`?
1. Typically, for this kind of operation, we will add the lock, update the list, use a defer function (to finish the handling) then start the actual operation.
2. Why does it handle the 1st entry of the list only?

pkg/sync/rpc/server.go
```diff
@@ -1505,3 +1529,137 @@ func (b *BackupList) Delete(backupID string) error {
...
+	return nil
+}
+
+func (s *SnapshotHashList) Refresh() error {
+	s.Lock()
+	defer s.Unlock()
+
+	var index, completed int
+
+	for index = len(s.tasks) - 1; index >= 0; index-- {
```
shuo-wu: Is this workflow actually a pulling mode notification? Typically we will use the pushing mode for this kind of progress update.

pkg/sync/sync.go
```diff
@@ -887,3 +894,112 @@ func CloneStatus(engineControllerClient *client.ControllerClient) (map[string]*S
...
+
+func (t *Task) HashSnapshot(snapshotNames []string, rehash bool, concurrentLimit int) error {
+	replicas, err := t.client.ReplicaList()
+	if err != nil {
+		return err
+	}
+
+	for _, r := range replicas {
+		if r.Mode != types.RW {
+			return fmt.Errorf("cannot hash a snapshot because %v is in %v mode", r.Address, r.Mode)
```
shuo-wu, line 864: I am thinking if the cmd should error out when there is a non-RW replica...

pkg/sync/sync.go
```diff
@@ -887,3 +894,112 @@ func CloneStatus(engineControllerClient *client.ControllerClient) (map[string]*S
...
+
+	for _, r := range replicas {
+		repClient, err := replicaClient.NewReplicaClient(r.Address)
+		if err != nil {
+			return nil, err
+		}
+		clients = append(clients, repClient)
+
+		if r.Mode != types.RW {
+			hashStatusMap[r.Address] = &SnapshotHashStatus{
```
shuo-wu, line 933: NIT: it's better to set `State` even if `Error` is not empty

pkg/replica/hash.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	snapshotDiskNames := []string{}
+	for _, name := range snapshotNames {
+		snapshotDiskNames = append(snapshotDiskNames, GenerateSnapshotDiskName(name))
+	}
+
+	r, err := NewReadOnly(dir, snapshotDiskNames[0], snapshotDiskNames, nil)
```
shuo-wu, line 121: Do we need to construct the struct `replica` for one file read? Can we directly open the sparse file (as a raw file)? This makes the modifications in `replica.go` look a little bit cumbersome.

shuo-wu (commented):
Will re-visit the PRs after addressing the below comments.

pkg/sync/rpc/server.go
derekbit: Yes, it's a pulling mode. Pulling mode is better. Do we have any pulling mode example in longhorn-engine?

pkg/replica/hash.go
derekbit: In some cases, it read a set of disk files, e.g. a snapshot disk file and a system-generated file.

pkg/sync/rpc/server.go
derekbit: You are right. Concurrent way is better. Thank you.

pkg/replica/hash.go
shuo-wu: a system-generated file?

pkg/sync/rpc/server.go
shuo-wu: I think the current progress update related implementation (backup creation for example) is more like the pushing mode rather than the pulling mode. (The progress and state will be updated inside the progress update function instead of relying on the sync agent server to pull it periodically.)

pkg/replica/hash.go
derekbit: After discussing with Shou, just hash non system-generated snapshot disk files.

pkg/sync/rpc/server.go
derekbit: Using waitgrounp instead, thank you,

proto/ptypes/syncagent.proto
derekbit: Updated, thank you.

pkg/util/disk/disk.go
derekbit: Updated, thank you.

app/cmd/snapshot.go
derekbit: Auto-correct ot 0 if < 0. Thank you.

app/cmd/snapshot.go
derekbit: Updated. Thank you.

pkg/hash/main.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	defer status.CloseSnapshot()
+
+	h := sha256.New()
+
+	size := status.GetSize()
+	blkCounts := size / defaultBlockSize
+	blkBuff := make([]byte, defaultBlockSize)
+
+	for i := int64(0); i < blkCounts; i++ {
```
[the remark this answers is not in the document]
derekbit: Check the ctime is not changed after hashing. Thank you.

pkg/sync/rpc/server.go
derekbit: Got it. So, the current method is same as the implemtaion of backup progress update.

proto/ptypes/syncagent.proto
```diff
@@ -28,6 +28,8 @@ service SyncAgentService {
   rpc SnapshotHash(SnapshotHashRequest) returns (google.protobuf.Empty) {}
   rpc SnapshotHashStatus(SnapshotHashStatusRequest) returns
     (SnapshotHashStatusResponse) {}
+  rpc SnapshotHashInProgressTasks(SnapshotHashInProgressTasksRequest) returns
```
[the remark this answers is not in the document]
derekbit: `SnapshotHashStatus` only returns single snapshot's status.
`SnapshotHashInProgressTasks` returns the maximum number of in-progress hashing tasks in the associated replicas' node. Longhorn-manager will check this value and determine if it can issue a new request.

pkg/sync/rpc/list.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	return nil
+}
+
+func (s *SnapshotHashList) Refresh() error {
+	s.Lock()
+	defer s.Unlock()
+
+	if err := s.cleanup(replica.ProgressStateComplete, MaxSnapshotHashTaskSize); err != nil {
```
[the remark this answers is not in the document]
derekbit: I think we don't need to limit the size. The failed or complete ones will be cleaned up in ADD operations.

pkg/replica/hash.go
```diff
@@ -0,0 +1,280 @@
...
+		offset := i * backupstore.DEFAULT_BLOCK_SIZE
+		if err := t.readSnapshot(offset, blkBuff); err != nil {
+			return "", err
+		}
+		written, err := h.Write(blkBuff)
+		if err != nil {
+			return "", err
+		}
+
+		t.Progress = int(100 * float64(offset+int64(written)) / float64(size))
```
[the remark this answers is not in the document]
derekbit: > Is there a lock protecting this?

I don't use a lock here. Any potential race condition here?

pkg/sync/sync.go
derekbit: If the replica is in WO state, the data might be changed. Do you think it should no error out here?

pkg/replica/hash.go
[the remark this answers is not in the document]
shuo-wu: The HashStatus API will read the progress while this goroutine keeps updating it... Isn't it a race condition?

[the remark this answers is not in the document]
derekbit: Well, the read operation won't be affected by the update.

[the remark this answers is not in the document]
shuo-wu: Sorry... Aren't the read and this update separate threads?

pkg/replica/hash.go
joshimoo: For knowledge sharing: CRC is a better approach for error detection, since it's less computationally expensive, but further it actually allows for detecting the individual bit/byte error based on the chosen CRC(size,polynominal) and the data size to be checked, it also can be done piece wise (different sections in parallel then merge).

It is correct that it's not secure against deliberate modifications and that the collision chance might be higher than with a cryptographic hash, but the chance for collisions can be increased/decreased by the choice of the CRC(size,polynominal) and the size of the data to be checked.

There is a 64bit package as well, but It doesn't look like the arm CPU has a dedicated instruction for it.
https://pkg.go.dev/hash/crc64@go1.19.2

Not sure of the top of my head on the intel side :)

pkg/replica/hash.go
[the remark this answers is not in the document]
derekbit: Updated, thank you.

derekbit:
Please help review again. Thank you.
cc @shuo-wu @c3y1huang @innobead

pkg/sync/sync.go
shuo-wu: I mean, postponing the hashing. If it makes thing complicated, we can ignore this now.

derekbit: Maybe requeue it in snapshot monitor?

derekbit: Implement the requeue the hash request in longhorn-manager 'ssnapshot monitor if encounter this error.

pkg/sync/sync.go
```diff
@@ -300,6 +311,24 @@ func (t *Task) markSnapshotAsRemoved(replicaInController *types.ControllerReplic
 	return nil
 }
 
+func (t *Task) cancelSnapshotHashTask(replicaInController *types.ControllerReplicaInfo, snapshot string) error {
+	if replicaInController.Mode != types.RW {
+		return fmt.Errorf("can only mark snapshot as removed from replica in mode RW, got %s", replicaInController.Mode)
```
shuo-wu, line 316: Wrong error message..

pkg/replica/hash.go
```diff
@@ -0,0 +1,340 @@
...
+}
+
+type SnapshotXattrHashInfo struct {
+	Method       string `json:"method"`
+	Checksum     string `json:"checksum"`
+	ModTime      string `json:"modTime"`
+	LastHashedAt string `json:"lastHashedAt"`
+}
+
+func NewSnapshotHashTask(ctx context.Context, cancel context.CancelFunc, snapshotName string, rehash bool) *SnapshotHashTask {
```
shuo-wu, line 66: NIT: it's better to rename this. `task` in this repo typically means the struct in `pkg/sync/sync.go`. This name is a little bit misleading...

pkg/replica/hash.go
```diff
@@ -0,0 +1,340 @@
...
+	}()
+
+	// Each node can have only snapshot hashing task at the same time.
+	// When the snapshot hashing task is initiated, the task will try to fetch the
+	// file lock (${fileLockDirectory}/hash), and other tasks need to wait for the lock.
+	// The file is unlocked after the task is completed.
+	fileLock, err := t.LockFile()
+	if err != nil {
+		return
+	}
```
shuo-wu, line 137: If the file locking failed (there is already one hashing in-progress, what will the caller handle it? Retry? Error out? Or stuck there?

pkg/replica/hash.go
```diff
@@ -0,0 +1,340 @@
...
+
+	err = SetSnapshotHashInfoToXattr(t.SnapshotName, checksum, modTime, lastHashedAt)
+	if err != nil {
+		return
+	}
+
+	err = t.isModTimeRemain(modTime)
+	if err != nil {
+		err = DeleteSnapshotHashInfoFromXattr(t.SnapshotName)
+	}
```
shuo-wu, line 178: NIT: execute this before `SetSnapshotHashInfoToXattr`

pkg/replica/hash.go
```diff
@@ -0,0 +1,340 @@
...
+
+	lastHashedAt := time.Now().UTC().Format(time.RFC3339)
+	checksum, err = hashSnapshot(t.Ctx, t.SnapshotName)
+	if err != nil {
+		return
+	}
+
+	if t.isSnapshotSilentlyCorrupted(checksum) {
+		t.StatusLock.Lock()
+		t.SilentlyCorrupted = true
```
shuo-wu, line 163: Why do you choose to use a separate flag instead of directly marking the state as `errror` when the mismatching is detected?

pkg/replica/hash.go
```diff
@@ -0,0 +1,340 @@
...
+}
+
+func (t *SnapshotHashTask) getSize() (int64, error) {
+	fileInfo, err := t.file.Stat()
+	if err != nil {
+		return -1, err
+	}
+
+	return fileInfo.Size(), nil
+}
```
shuo-wu, line 260: What's this for? BTW, this returns the apparent file size.

pkg/sync/rpc/server.go
```diff
@@ -1399,109 +1390,83 @@ func (s *SyncAgentServer) rmDisk(disk string) error {
...
+	return checksum, nil
+}
 
-	for _, info := range b.backups {
-		if info.backupID == backupID {
-			return info.backupStatus, nil
+func (s *SyncAgentServer) SnapshotHashStatus(ctx context.Context, req *ptypes.SnapshotHashStatusRequest) (*ptypes.SnapshotHashStatusResponse, error) {
+	task, err := s.SnapshotHashList.Get(req.SnapshotName)
+	if err != nil {
+		checksum, err := checkSnapshotHashStatusFromXattr(req.SnapshotName)
```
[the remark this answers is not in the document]
derekbit: By default, the hash status should be retrieved from the status in the `SnapshotHashList`. 

After finishing the hash task, if the data is somehow cleaned up by `SnapshotHashList`'s refresh the result can be read from xattr instead.

pkg/sync/rpc/server.go
```diff
@@ -1399,109 +1390,83 @@ func (s *SyncAgentServer) rmDisk(disk string) error {
...
+		if err != nil {
+			return &ptypes.SnapshotHashStatusResponse{
+				State: string(replica.ProgressStateError),
+				Error: err.Error(),
+			}, nil
```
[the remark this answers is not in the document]
derekbit: I think we don't need to update the state in the list because the `Get()` already failed.

pkg/replica/hash.go
derekbit: This is a blocking lock. If anothoer hashing in-progress, it will wait for the lock. If the error is caused by another issue such as EBADF, the hashing fails.

pkg/replica/hash.go
derekbit: The `error` records the reason why the checksum cannot be successfully got.  If the silently corruption happens, the checksum still can be got.

pkg/replica/hash.go
derekbit: Dead codes. Will remove it.

pkg/sync/rpc/list.go
```diff
@@ -0,0 +1,280 @@
...
+		return err
+	}
+
+	return s.cleanup(replica.ProgressStateError, MaxSnapshotHashTaskSize)
```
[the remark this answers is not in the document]
derekbit: > This cleanup means retaining at most limit count of hash entries that in the specified state, right?

Yes

> Besides, can we input retainingStateMap map[replica.ProgressState]bool instead?

Do you mean like this? Only call `cleanup` once?
```
retainingStateMap := map[replica.ProgressState]bool{
    replica.ProgressStateComplete: true,
    replica. ProgressStateError: true,
}

s.cleanup(retainingStateMap, MaxSnapshotHashTaskSize)
```

pkg/sync/sync.go
derekbit: Addressed. Thank you.

pkg/replica/hash.go
derekbit: `SetSnapshotHashInfoToXattr` should override the xattr, so the `DeleteSnapshotHashInfoFromXattr` might be not necessary.
`DeleteSnapshotHashInfoFromXattr` here is just to delete the useless information.

pkg/sync/rpc/list.go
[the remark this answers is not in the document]
shuo-wu: Yes. That's the idea.

pkg/replica/hash.go
shuo-wu: OK, got it. Then the next questions are, can we add one more state instead? Comparing this, which one is better?

derekbit: Good idea. Then, the slient corruption can be known quickly without the recalculation.

pkg/replica/hash.go
derekbit: > Can I say that all of the workers have a chance to send a hash request to a specific node? Then how about 100 nodes encountering the same scenario?

Yeah, in the case, some nodes might be idle. Howerever, if the volumes and replicas are evenly distributed, the idle cases should be reduced a lot.

pkg/sync/rpc/server.go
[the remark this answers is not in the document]
derekbit: I think `NotFound` is enough here. It implicitly indicates the calculation failed.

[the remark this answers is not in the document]
shuo-wu: Sorry...? You didn't add a NotFound error validation here...

pkg/replica/hash.go
shuo-wu: I mean, in the example mentioned above, there may be 500 goroutinues getting stuck in this locking...

derekbit: I see. Would you mind reducing the 5 task workers to 1?
Probably we can drop the request if the number of waiting tasks reaches to the limit. The number of waiting tasks can be recorded in the file lock's content.

shuo-wu: I think it's better to return an error here rather than keep waiting. Then the longhorn manager will requeue the task with a delay. But I am not sure if it would be too complicated.

derekbit: So, no waiting tasks and requeue the tasks that cannot get the lock. Correct?

shuo-wu: As long as there is no tons of task waiting there, I am OK with any other solutions.

derekbit: Addressed. Thank you.

shuo-wu: Can you point out the related modifications? BTW, how do you requeue the task if there is an existing in-progress task? I didn't find the code in the longhorn-manager PR.

derekbit: Before requesting, engine will check lock state of the replica's node ([codes](https://github.com/longhorn/longhorn-engine/blob/a5df8bb33eb46416120b20f525140abf2159e84d/pkg/sync/sync.go#L909)). If the lock file is held by another task, then engine returns the error `cannot request hashing snapshot because %s is hashing snapshot`.

In snashot monitor side, if the error `cannot request hashing snapshot` is received, then requeue the failed request ([codes](https://github.com/longhorn/longhorn-manager/blob/717cbb177393dca64d409e5c7d9d38cd085be7ec/controller/monitor/snapshot_monitor.go#L215))

shuo-wu: OK. The target sync agent server will check if other servers (of other replicas) are hashing... 
As we discussed before, can we move this to the instance manager part? cc @joshimoo

pkg/replica/hash.go
shuo-wu: Just for confirmation, you still prefer to use a separate field `SilentlyCorrupted` rather than introducing a new state for this case, right?

pkg/replica/hash.go
```diff
@@ -0,0 +1,400 @@
...
+	}
+	defer t.UnlockFile(fileLock)
+
+	modTime, err := GetSnapshotModTime(t.SnapshotName)
+	if err != nil {
+		return
+	}
+
+	// If the silent corruption is detected, don't need to recalculate the checksum.
+	// Just set SilentlyCorrupted to ture and return it.
```
shuo-wu, line 142: typo `ture`

pkg/replica/hash.go
```diff
@@ -0,0 +1,400 @@
...
+
+	// If the silent corruption is detected, the xattr will not be overrode.
+	// The scene will be preserved.
+	if t.isSnapshotSilentlyCorrupted(checksum) {
+		t.StatusLock.Lock()
+		t.SilentlyCorrupted = true
+		t.StatusLock.Unlock()
+		// Do the best to set SilentlyCorrupted to true.
+		// Never mind if failed to set it. The checksum will be recalculated and get the SilentlyCorrupted again.
+		if err := SetSilentlyCorruptedToXattr(t.SnapshotName); err != nil {
```
shuo-wu, line 180: NIT: can we use a single `SetSnapshotHashInfoToXattr` instead?

pkg/replica/hash.go
derekbit: Engine directly sends the requests to replicas' sync-agent servers, so handling the logic in IM-r is not available. If we implement the replica proxy in the future (i.e engine process->replica proxy->replica process), we can move it to IM-r and record the count in the IM-r. Let's improve it in the future. WDYT?

pkg/replica/hash.go
derekbit: Yeah. But because I don't want to change other values in the xattr, so the `GetSnapshotHashFromToXattr` and then `SetSnapshotHashInfoToXattr` are necessary. The silent corruption is not common, the two operations will not introduce negative imact.

pkg/replica/hash.go
derekbit: Updated. Thank you.

pkg/replica/hash.go
shuo-wu: Just revisit the logic. Why not directly use a defer function to blindly update the xattr? It will be updated when the modTime remains while there is a difference from the current value.

derekbit: Refactored this part as you mentioned. Thank you.

pkg/replica/hash.go
```diff
@@ -0,0 +1,413 @@
...
+			t.State = ProgressStateComplete
+
+			logrus.Infof("Snapshot %v checksum %v", t.SnapshotName, checksum)
+
+			// Checksum is not recalculated in this task, so return directly.
+			if lastHashedAt == "" {
+				return
+			}
+
+			err = SetSnapshotHashInfoToXattr(t.SnapshotName, &SnapshotXattrHashInfo{
```
shuo-wu, line 133: Can we leave this error unhandled then directly execute `t.isModTimeRemain(modTime)`?

derekbit: Yes, the error can be ignored. Then, next task will re-calculated and set it again.
Addressed. Thank you.

shuo-wu (approved):

pkg/replica/hash.go
```diff
@@ -0,0 +1,413 @@
...
+	modTime := info.ModTime
+
+	if modTime != currentModTime || checksum == "" {
+		return true, "", nil
+	}
+
+	return false, checksum, nil
+}
+
+func (t *SnapshotHashJob) isModTimeRemain(oldModTime string) error {
```
innobead, line 349: In general, `isXXX` is for returning bool, but of course can have an extra error returned. Here, I would suggest you return (bool, error), but change the code to below.

```go
newModTime, err := GetSnapshotModTime(t.SnapshotName)
if err != nil {
    return false, err
}

return oldModTime != newModTime, nil
```

pkg/replica/hash.go
```diff
@@ -0,0 +1,413 @@
...
+	}
+
+	if checksum != existingChecksum {
+		return true
+	}
+
+	return false
+}
+
+func GetSnapshotModTime(snapshotName string) (string, error) {
```
innobead, line 249: Again, if a function/method used internally, no need to export, so it should `getSnapshotModTime`. This way can make the code self-explanatory.

innobead (changes requested):
In general, LGTM. However, some comments need to resolve.

A few convention suggestions for any language. 
- About accessors, if a method/function/property/struct used internally, no need to export to make code self-explantory 
- Function name matters, `isXXX` for returning bool, `check/validate/verb/XXX` for returning an optional specific object. All of them can bundle an error for external info purpose

pkg/replica/hash.go
derekbit: `GetSnapshotModTime` is used by the functions in `rpc` and `replica` use

innobead: Then ok!

pkg/sync/rpc/list.go
```diff
@@ -0,0 +1,283 @@
...
+
+	purgedStates := map[replica.ProgressState]struct{}{
+		replica.ProgressStateComplete: {},
+		replica.ProgressStateError:    {},
+	}
+
+	return s.purgePartialRetained(purgedStates, MaxSnapshotHashJobSize)
+}
+
+func (s *SnapshotHashList) purgePartialRetained(purgedStates map[replica.ProgressState]struct{}, limit int) error {
```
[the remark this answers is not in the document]
derekbit: Yeah, list/array is engouh for this case.

pkg/sync/rpc/server.go
```diff
@@ -1399,109 +1391,110 @@ func (s *SyncAgentServer) rmDisk(disk string) error {
...
-	}
-	return nil, fmt.Errorf("backup not found %v", backupID)
+		return &ptypes.SnapshotHashStatusResponse{
+			State:    string(replica.ProgressStateComplete),
+			Checksum: checksum,
+		}, nil
+	}
+
+	task.StatusLock.Lock()
+	defer task.StatusLock.Unlock()
```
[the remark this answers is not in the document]
derekbit: Yes, should use read lock here.

pkg/sync/sync.go
```diff
@@ -887,3 +887,185 @@ func CloneStatus(engineControllerClient *client.ControllerClient) (map[string]*S
...
+
+func (t *Task) HashSnapshot(snapshotName string, rehash bool) error {
+	replicas, err := t.client.ReplicaList()
+	if err != nil {
+		return err
+	}
+
+	for _, r := range replicas {
+		if r.Mode != types.RW {
+			return fmt.Errorf(types.CannotRequestHashingSnapshotPrefix+" because %v is in %v mode", r.Address, r.Mode)
```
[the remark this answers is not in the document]
derekbit: longhorn-manager will check if the error messages contains `CannotRequestHashingSnapshotPrefix`, so I put it `CannotRequestHashingSnapshotPrefix` in `types/error.go`.

Putting `CannotRequestHashingSnapshotPrefix` in `sync.go` will introduce too many unnecessary dependencies. Then, building `longhorn-manager` will encounter the below error messages
```
+ go build -ldflags '-X github.com/longhorn/longhorn-manager/meta.Version=v1.3.0    -X github.com/longhorn/longhorn-manager/meta.GitCommit=78775fcc1694eb8706a5425101d75b2c5d26002d-dirty    -X github.com/longhorn/longhorn-manager/meta.BuildDate=2022-11-16T08:35:16+00:00    -extldflags -static -s' -o bin/longhorn-manager
package github.com/longhorn/longhorn-manager
        imports github.com/longhorn/longhorn-manager/app
        imports github.com/longhorn/longhorn-manager/api
        imports github.com/longhorn/longhorn-manager/controller
        imports github.com/longhorn/longhorn-manager/controller/monitor
        imports github.com/longhorn/longhorn-engine/pkg/sync
        imports github.com/longhorn/longhorn-engine/pkg/replica
        imports github.com/longhorn/longhorn-engine/pkg/backingfile
        imports github.com/longhorn/longhorn-engine/pkg/qcow: build constraints exclude all Go files in /go/src/github.com/longhorn/longhorn-manager/vendor/github.com/longhorn/longhorn-engine/pkg/qcow
```

pkg/sync/sync.go
```diff
@@ -887,3 +887,185 @@ func CloneStatus(engineControllerClient *client.ControllerClient) (map[string]*S
...
+			defer wg.Done()
+			repClient, err := replicaClient.NewReplicaClient(r.Address)
+			if err != nil {
+				syncErrorMap.Store(r.Address, err)
+				return
+			}
+			defer repClient.Close()
+			if err := repClient.SnapshotHash(snapshotName, rehash); err != nil {
+				syncErrorMap.Store(r.Address, err)
+			}
```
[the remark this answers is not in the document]
derekbit: Updated. Thank you.

app/cmd/snapshot.go
```diff
@@ -166,6 +169,45 @@ func SnapshotCloneStatusCmd() cli.Command {
...
+		},
+	}
+}
+
+func SnapshotHashCancelCmd() cli.Command {
+	return cli.Command{
+		Name: "hash-cancel",
+		Action: func(c *cli.Context) {
+			if err := cancelHashSnapshot(c); err != nil {
+				logrus.Fatalf("Error running cancel hashing snapshot command: %v", err)
```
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

pkg/replica/hash.go
```diff
@@ -0,0 +1,413 @@
...
+	return hex.EncodeToString(h.Sum(nil)), nil
+}
+
+type readerFunc func(p []byte) (n int, err error)
+
+func (rf readerFunc) Read(p []byte) (n int, err error) {
+	return rf(p)
+}
+
+func Copy(ctx context.Context, dst io.Writer, src io.Reader) error {
```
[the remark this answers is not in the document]
derekbit: Updated. Thank you.

pkg/sync/rpc/server.go
```diff
@@ -1399,109 +1391,110 @@ func (s *SyncAgentServer) rmDisk(disk string) error {
...
-	if err := b.Refresh(); err != nil {
-		return err
-	}
+		if err := s.SnapshotHashList.Add(req.SnapshotName, task); err != nil {
+			logrus.Errorf("failed to add snapshot %v for hashing", req.SnapshotName)
+			return
+		}
 
-	return nil
+		task.Execute()
```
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

pkg/sync/rpc/server.go
```diff
@@ -1399,109 +1391,110 @@ func (s *SyncAgentServer) rmDisk(disk string) error {
...
-	b.Unlock()
+	go func() {
+		ctx, cancel := context.WithCancel(context.Background())
+		task := replica.NewSnapshotHashJob(ctx, cancel, req.SnapshotName, req.Rehash)
 
-	if err := b.Refresh(); err != nil {
-		return err
-	}
+		if err := s.SnapshotHashList.Add(req.SnapshotName, task); err != nil {
+			logrus.Errorf("failed to add snapshot %v for hashing", req.SnapshotName)
```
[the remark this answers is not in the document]
derekbit: Updated. Thank you.

pkg/sync/rpc/server.go
```diff
@@ -1399,109 +1391,110 @@ func (s *SyncAgentServer) rmDisk(disk string) error {
...
+				State: string(replica.ProgressStateError),
+				Error: err.Error(),
+			}, nil
 		}
-	}
-	return nil, fmt.Errorf("backup not found %v", backupID)
+		return &ptypes.SnapshotHashStatusResponse{
+			State:    string(replica.ProgressStateComplete),
+			Checksum: checksum,
+		}, nil
```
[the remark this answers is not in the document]
derekbit: Updated. Thank you.

pkg/sync/rpc/server.go
```diff
@@ -1399,109 +1391,110 @@ func (s *SyncAgentServer) rmDisk(disk string) error {
...
 	}
-	return append(b[:index], b[index+1:]...), nil
-}
 
-// Refresh deletes all the old completed backups from the front. Old backups are the completed backups
-// that are created before MaxBackupSize completed backups
-func (b *BackupList) Refresh() error {
-	b.Lock()
-	defer b.Unlock()
+	task.CancelFunc()
```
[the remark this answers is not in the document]
derekbit: cancel method (`type CancelFunc func()`) does not have returned value, so the error handling is not necessary.

pkg/sync/rpc/list.go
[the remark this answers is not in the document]
derekbit: Updated, thank you.

pkg/sync/rpc/list.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+type SnapshotHashList struct {
+	sync.RWMutex
+	jobs []*SnapshotHashInfo
+}
+
+type SnapshotHashInfo struct {
+	snapshotName string
+	info         *replica.SnapshotHashJob
```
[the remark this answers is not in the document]
derekbit: Updated, thank you.

pkg/sync/rpc/list.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+type BackupInfo struct {
+	backupID     string
+	backupStatus *replica.BackupStatus
+}
+
+type SnapshotHashList struct {
+	sync.RWMutex
+	jobs []*SnapshotHashInfo
```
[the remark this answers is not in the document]
derekbit: Use `infos` in backup and snapshotHash lists.

pkg/replica/hash.go
```diff
@@ -0,0 +1,413 @@
...
+	// The scene will be preserved and only set silentlyCorrupted to true.
+	if t.isSnapshotSilentlyCorrupted(checksum) {
+		silentlyCorrupted = true
+
+		info, err := GetSnapshotHashInfoFromXattr(t.SnapshotName)
+		if err != nil && err != syscall.ENODATA {
+			return
+		}
+
+		if info == nil {
```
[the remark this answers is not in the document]
derekbit: Yes, should be `if info != nil {`. Fixed, thank you.

pkg/sync/sync.go
```diff
@@ -887,3 +887,185 @@ func CloneStatus(engineControllerClient *client.ControllerClient) (map[string]*S
...
+				return errors.Wrapf(err, "failed to create replica client to %v", r.Address)
+			}
+
+			lock.Lock()
+			clients = append(clients, repClient)
+			lock.Unlock()
+
+			if r.Mode != types.RW {
+				return fmt.Errorf("replica %v since it is in %v mode", r.Address, r.Mode)
+			}
```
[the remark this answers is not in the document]
derekbit: Fixed. Thank you.

pkg/sync/sync.go
```diff
@@ -887,3 +887,185 @@ func CloneStatus(engineControllerClient *client.ControllerClient) (map[string]*S
...
+			_ = client.Close()
+		}
+	}()
+
+	wg := sync.WaitGroup{}
+
+	for _, r := range replicas {
+		wg.Add(1)
+
+		go func(r *types.ControllerReplicaInfo) error {
```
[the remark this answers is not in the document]
derekbit: Yes, no need to return the error, because it should be saved in `SnapshotHashStatus`.

innobead (commented):
`Unify fatal log message` awesome!

It's good to do minor refactoring during the feature dev or bug fix.

innobead (approved):

PR #764, opened 2022-10-13
fix(qemu-img): Require -F with -b backing image

mantissahz:
New version of qemu-img requires `-F` flag for backing file format when we upgrade the base image version SLES from 15.3 to 15.4

https://gitlab.com/qemu-project/qemu/-/commit/497a30dbb065937d67f6c43af6dd78492e1d6f6d

longhorn/longhorn#4632

Signed-off-by: James Lu <james.lu@suse.com>

### DIFF
--- a/app/cmd/restore_to_file.go
+++ b/app/cmd/restore_to_file.go
@@ -240,7 +240,7 @@ func MergeSnapshotsToBackingFile(snapFilepath, backingFilepath string) error {
 
 func rebaseSnapshot(snapFilepath, backingFilepath string) error {
 	logrus.Infof("Start rebaseSnapshot %s -> %s", snapFilepath, backingFilepath)
-	_, err := iutil.ExecuteWithoutTimeout(QEMUImageBinary, []string{"rebase", "-u", "-b", backingFilepath, snapFilepath})
+	_, err := iutil.ExecuteWithoutTimeout(QEMUImageBinary, []string{"rebase", "-u", "-b", backingFilepath, "-F", DefaultOutputFormat, snapFilepath})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


app/cmd/restore_to_file.go
```diff
@@ -240,7 +240,7 @@ func MergeSnapshotsToBackingFile(snapFilepath, backingFilepath string) error {
 
 func rebaseSnapshot(snapFilepath, backingFilepath string) error {
 	logrus.Infof("Start rebaseSnapshot %s -> %s", snapFilepath, backingFilepath)
-	_, err := iutil.ExecuteWithoutTimeout(QEMUImageBinary, []string{"rebase", "-u", "-b", backingFilepath, snapFilepath})
+	_, err := iutil.ExecuteWithoutTimeout(QEMUImageBinary, []string{"rebase", "-u", "-b", backingFilepath, "-F", DefaultOutputFormat, snapFilepath})
```
derekbit, line 243: Is there any possible compatibility issue?

mantissahz: I think it's ok because this flag exists (after -b) for a long time, it just could be omitted before.
https://www.unix.com/man-page/suse/1/qemu-img/ this is the man page of qemu-img in 2010
Then I run this at 15.3 as well and it passed all tests.

shuo-wu: Probably NO. SLES 15.3 is using `qemu-img` version 5.2.0, in which the usage is already valid.

```
rebase [--object objectdef] [--image-opts] [-U] [-q] [-f fmt] [-t cache] [-T src_cache] [-p] [-u] -b backing_file [-F backing_fmt] filename
```

shuo-wu (approved):

innobead (approved):

innobead:
@mantissahz Please backport this to 1.2 and 1.3.

PR #769, opened 2022-11-11
Replica must be closed when rebuilding failed with `too many snapshots created`

mantissahz:
Replica banckend needs to be closed if the number of sanpshots is up to the limitation and replica rebuding is occurred.

longhorn/longhorn#3828

### DIFF
--- a/pkg/controller/control.go
+++ b/pkg/controller/control.go
@@ -181,10 +181,8 @@ func (c *Controller) Snapshot(name string, labels map[string]string) (string, er
 		name = util.UUID()
 	}
 
-	if remain, err := c.backend.RemainSnapshots(); err != nil {
+	if _, err := c.backend.RemainSnapshots(); err != nil {
 		return "", err
-	} else if remain <= 0 {
-		return "", fmt.Errorf("too many snapshots created")
 	}
 
 	created := util.Now()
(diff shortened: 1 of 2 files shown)


pkg/controller/control.go
```diff
@@ -323,15 +329,14 @@ func (c *Controller) addReplicaNoLock(newBackend types.Backend, address string,
 		if remain, err := c.backend.RemainSnapshots(); err != nil {
 			return err
 		} else if remain <= 0 {
-			return fmt.Errorf("too many snapshots created")
+			err = fmt.Errorf("too many snapshots created")
```
derekbit, line 332: IMO, the error messge should be from `RemainSnapshots`'s `err`. Using `remain <= 0` to chec the `too many snapshots created` eror is a bit of weird.

mantissahz: `RemainSnapshots()` only returns numbers of snapshot if the replica is in proper states. 
`RemainSnapshots` would be `maximumChainLength - len(r.activeDiskData)`
So maybe we check `remain <= 0` in `RemainSnapshots()`?
Does anyone know why `maximumChainLength` is 250?

innobead: I agree with Derek, because at the end you still need to handle `remain <= 0` as an error case, so the error should be good to cover in the `RemainSnapshots` method. However, this is about refactoring, can create a ticket to do it later.

About 250, no ideas, but I believe it's quite a big number for users and it will hurt user experience when the volume used by write heavy application.

innobead (approved):

innobead:
@mantissahz wants to improve https://github.com/longhorn/longhorn-engine/pull/769#discussion_r1021047493, so let's merge this after that part is fixed.

mantissahz:
Rebuild the engine image with refactored codes locally and do some testings.

innobead (approved):
LGTM

PR #772, opened 2022-11-24
Replace the modTime (mtime) with ctime in snapshot hash

derekbit:
https://github.com/longhorn/longhorn/issues/4934

### DIFF
--- a/pkg/replica/hash.go
+++ b/pkg/replica/hash.go
@@ -18,7 +18,6 @@ import (
 	"github.com/gofrs/flock"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	"golang.org/x/sys/unix"
 
 	diskutil "github.com/longhorn/longhorn-engine/pkg/util/disk"
 )
@@ -59,7 +58,7 @@ type SnapshotHashJob struct {
 type SnapshotXattrHashInfo struct {
 	Method            string `json:"method"`
 	Checksum          string `json:"checksum"`
-	ModTime           string `json:"modTime"`
+	ChangeTime        string `json:"changeTime"`
 	LastHashedAt      string `json:"lastHashedAt"`
 	SilentlyCorrupted bool   `json:"silentlyCorrupted"`
 }
@@ -108,7 +107,7 @@ func (t *SnapshotHashJob) UnlockFile(fileLock *flock.Flock) {
 
 func (t *SnapshotHashJob) Execute() (err error) {
 	var checksum string
-	var modTime string
+	var changeTime string
 	var lastHashedAt string
 	var silentlyCorrupted bool
 
@@ -131,12 +130,12 @@ func (t *SnapshotHashJob) Execute() (err error) {
 			SetSnapshotHashInfoToXattr(t.SnapshotName, &SnapshotXattrHashInfo{
 				Method:            defaultHashMethod,
 				Checksum:          checksum,
-				ModTime:           modTime,
+				ChangeTime:        changeTime,
 				LastHashedAt:      lastHashedAt,
 				SilentlyCorrupted: silentlyCorrupted,
 			})
 
-			remain, err := t.isModTimeRemain(modTime)
+			remain, err := t.isChangeTimeRemain(changeTime)
 			if !remain {
 				if err == nil {
 					err = fmt.Errorf("snapshot %v modification time is changed", t.SnapshotName)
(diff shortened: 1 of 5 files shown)


pkg/replica/hash.go
```diff
@@ -246,93 +243,141 @@ func (t *SnapshotHashJob) isSnapshotSilentlyCorrupted(checksum string) bool {
...
-	return fileInfo.ModTime().String(), nil
+	stat := fileInfo.Sys().(*syscall.Stat_t)
+	return time.Unix(int64(stat.Ctim.Sec), int64(stat.Ctim.Nsec)).String(), nil
 }
 
-func GetSnapshotHashInfoFromXattr(snapshotName string) (*SnapshotXattrHashInfo, error) {
-	xattrSnapshotHashValue := make([]byte, xattrSnapshotHashValueMax)
-	_, err := unix.Getxattr(diskutil.GenerateSnapshotDiskName(snapshotName), xattrSnapshotHashName, xattrSnapshotHashValue)
+func GetSnapshotHashInfoFromChecksumFile(snapshotName string) (*SnapshotXattrHashInfo, error) {
+	dir, err := os.Getwd()
```
innobead, line 257: Why `os.Getwd()` instead of `/var/lib/longhonr/..`?

pkg/replica/hash.go
```diff
@@ -246,93 +243,141 @@ func (t *SnapshotHashJob) isSnapshotSilentlyCorrupted(checksum string) bool {
...
+	if err := f.Close(); err != nil {
+		return err
+	}
+
+	return os.Rename(tmpPath, path)
 }
 
-func DeleteSnapshotHashInfoFromXattr(snapshotName string) error {
-	return unix.Removexattr(diskutil.GenerateSnapshotDiskName(snapshotName), xattrSnapshotHashName)
+func DeleteSnapshotHashInfoFromChecksumFile(snapshotName string) error {
```
innobead, line 328: Rename to `DeleteSnapshotHashInfoChecksumFile`

pkg/replica/hash.go
```diff
@@ -246,93 +243,141 @@ func (t *SnapshotHashJob) isSnapshotSilentlyCorrupted(checksum string) bool {
...
 		return err
 	}
+	defer f.Close()
+
+	if err := json.NewEncoder(f).Encode(&obj); err != nil {
+		return err
+	}
 
-	return unix.Setxattr(diskutil.GenerateSnapshotDiskName(snapshotName), xattrSnapshotHashName, xattrSnapshotHashValue, 0)
+	if err := f.Close(); err != nil {
```
innobead, line 321: You have `defer f.Close()` above already, why need here again?

innobead (changes requested):

pkg/replica/hash.go
```diff
@@ -255,47 +253,94 @@ func GetSnapshotChangeTime(snapshotName string) (string, error) {
 	return time.Unix(int64(stat.Ctim.Sec), int64(stat.Ctim.Nsec)).String(), nil
 }
 
-func GetSnapshotHashInfoFromXattr(snapshotName string) (*SnapshotXattrHashInfo, error) {
-	xattrSnapshotHashValue := make([]byte, xattrSnapshotHashValueMax)
-	_, err := unix.Getxattr(diskutil.GenerateSnapshotDiskName(snapshotName), xattrSnapshotHashName, xattrSnapshotHashValue)
+func GetSnapshotHashInfoFromChecksumFile(snapshotName string) (*SnapshotXattrHashInfo, error) {
+	dir, err := os.Getwd()
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot get working directory")
```
c3y1huang, line 259:
```suggestion
		return nil, errors.Wrap(err, "failed to get working directory when getting snapshot hash info")
```

pkg/replica/hash.go
```diff
@@ -255,47 +253,94 @@ func GetSnapshotChangeTime(snapshotName string) (string, error) {
...
-	return info, nil
+	return &info, nil
 }
 
-func SetSnapshotHashInfoToXattr(snapshotName string, info *SnapshotXattrHashInfo) error {
-	xattrSnapshotHashValue, err := json.Marshal(&SnapshotXattrHashInfo{
+func SetSnapshotHashInfoToChecksumFile(snapshotName string, info *SnapshotXattrHashInfo) error {
+	dir, err := os.Getwd()
+	if err != nil {
+		return errors.Wrap(err, "cannot get working directory")
```
c3y1huang, line 282:
```suggestion
		return errors.Wrap(err, "failed to get working directory when setting snapshot hash info")
```

pkg/replica/hash.go
derekbit: `os.Getwd()` can get the replica dir, which is the working dir of the sync-agent ([codes](https://github.com/longhorn/longhorn-engine/blob/master/app/cmd/replica.go#L156))

pkg/replica/hash.go
derekbit: Redundant. Removed.

pkg/replica/hash.go
derekbit: Updated, thank you.

pkg/replica/hash.go
derekbit: Updated, thank you.

pkg/replica/hash.go
derekbit: Updated, thank you.

pkg/replica/hash.go
derekbit: Updated, thank you.

innobead (approved):

PR #775, opened 2022-12-01
fix(replica): volume metafile deleted or empty

mantissahz:
Volume will be stuck in attaching and detaching loop when attaching a volume with volume metafile deleted or empty.

Try to rebuild the volume metafile when attaching a volume.

longhorn/longhorn#4846

Regression test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2601

Run a new Regression test: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2628/

### DIFF
--- a/pkg/replica/replica.go
+++ b/pkg/replica/replica.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path"
@@ -171,6 +172,11 @@ func construct(readonly bool, size, sectorSize int64, dir, head string, backingF
 	r.info.Size = size
 	r.info.SectorSize = sectorSize
 	r.volume.sectorSize = diskutil.VolumeSectorSize
+
+	// Try to recover volume metafile if deleted or empty.
+	if err := r.tryRecoverVolumeMetaFile(head); err != nil {
+		return nil, err
+	}
 
 	// Scan all the disks to build the disk map
 	exists, err := r.readMetadata()
@@ -986,6 +992,50 @@ func (r *Replica) readMetadata() (bool, error) {
 	}
 
 	return len(r.diskData) > 0, nil
+}
+
+func (r *Replica) tryRecoverVolumeMetaFile(head string) error {
+	valid, err := r.checkValidVolumeMetaData()
+	if err != nil {
+		return err
+	}
+	if valid {
+		return nil
+	}
+
+	if head != "" {
+		r.info.Head = head
+	}
+
+	if r.info.Head == "" {
+		files, err := ioutil.ReadDir(r.dir)
+		if err != nil {
+			return err
+		}
+		for _, file := range files {
+			if strings.Contains(file.Name(), types.VolumeHeadName) {
+				r.info.Head = file.Name()
+				break
+			}
+		}
+	}
+
+	logrus.Warnf("Recovering volume metafile %v, and replica info: %+v", r.diskPath(volumeMetaData), r.info)
+	return r.writeVolumeMetaData(true, r.info.Rebuilding)
+}
+
+func (r *Replica) checkValidVolumeMetaData() (bool, error) {
+	err := r.unmarshalFile(r.diskPath(volumeMetaData), &r.info)
+	if err == nil {
+		return true, nil
+	}
+
+	// recover metadata file that does not exist or is empty
+	if os.IsNotExist(err) || errors.Is(err, io.EOF) {
+		return false, nil
+	}
+
+	return false, err
 }
 
 func (r *Replica) readDiskData(file string) error {
--- a/pkg/replica/server.go
+++ b/pkg/replica/server.go
@@ -113,7 +113,19 @@ func (s *Server) Status() (State, Info) {
 		info, err := ReadInfo(s.dir)
 		if os.IsNotExist(err) {
 			return Initial, Info{}
-		} else if err != nil {
+		}
+
+		replica := Replica{dir: s.dir}
+		volumeMetaFileValid, vaildErr := replica.checkValidVolumeMetaData()
+		if vaildErr != nil {
+			logrus.Errorf("Failed to check if volume metedata is valid in replica directory %s: %v", s.dir, err)
+			return Error, Info{}
+		}
+		if !volumeMetaFileValid {
+			return Initial, Info{}
+		}
+
+		if err != nil {
 			logrus.Errorf("Failed to read info in replica directory %s: %v", s.dir, err)
 			return Error, Info{}
 		}


pkg/replica/replica.go
```diff
@@ -221,6 +221,10 @@ func construct(readonly bool, size, sectorSize int64, dir, head string, backingF
 		}
 	}
 
+	if err := r.checkVolumeMetaFile(); err != nil {
```
innobead, line 224: If the volume.meta is empty, doesn't it mean this function will fail at the above step `exists, err := r.readMetadata()` earlier before running this recovery? If yes, why this will work?

innobead (changes requested):

pkg/replica/replica.go
mantissahz: Because file which is empty had been removed at (s *Server) Status()(State, Info)

pkg/replica/replica.go
```diff
@@ -973,6 +1008,9 @@ func (r *Replica) readMetadata() (bool, error) {
 
 	for _, file := range files {
 		if file.Name() == volumeMetaData {
+			if file.Size() == 0 {
+				continue
```
innobead, line 1012: Suggest having a log for this intended operation.

pkg/replica/replica.go
```diff
@@ -232,6 +236,37 @@ func construct(readonly bool, size, sectorSize int64, dir, head string, backingF
 	return r, r.writeVolumeMetaData(true, r.info.Rebuilding)
 }
 
+func (r *Replica) checkVolumeMetaFile() (err error) {
```
derekbit, line 239: Suggest separating the `checkVolumeMetaFile` to tw functions `checkVolumeMetaFile` and `createVolumeMetaFile`. Have the functions more clear and specific.

pkg/replica/server.go
```diff
@@ -113,7 +115,19 @@ func (s *Server) Status() (State, Info) {
...
+		if errors.Is(err, io.EOF) {
+			volumeMetaFilePath := (&Replica{dir: s.dir}).diskPath(volumeMetaData)
+			fileInfo, statErr := os.Stat(volumeMetaFilePath)
+			if statErr != nil {
+				logrus.Errorf("Failed to read info in replica directory %s: %v", s.dir, err)
+				return Error, Info{}
+			} else if fileInfo.Size() == 0 {
+				os.Remove(volumeMetaFilePath)
+				return Initial, Info{}
+			}
```
derekbit, line 128:
```suggestion
			if statErr != nil {
				logrus.WithError(err).Errorf("Failed to read info in replica directory %s", s.dir)
				return Error, Info{}
			}
			
			if fileInfo.Size() == 0 {
				if err := os.Remove(volumeMetaFilePath); err != nil {
				    log the error.... 
				 }
				return Initial, Info{}
			}
```

pkg/replica/server.go
```diff
@@ -113,7 +115,19 @@ func (s *Server) Status() (State, Info) {
...
+		}
```
innobead, line 129: Should not do any `change` operation in `get`-style function/method.

pkg/replica/replica.go
innobead: Suggest recovering the file before `readMetadata`, so you don't need to change other places.

pkg/replica/replica.go
```diff
@@ -973,6 +1008,9 @@ func (r *Replica) readMetadata() (bool, error) {
 
 	for _, file := range files {
 		if file.Name() == volumeMetaData {
+			if file.Size() == 0 {
+				continue
+			}
```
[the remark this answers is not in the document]
mantissahz: Do not know why file why be empty now.

pkg/replica/replica.go
innobead, line 239: Suggest renaming `tryRecoverVolumeMetaFile`

innobead (changes requested):

pkg/replica/replica.go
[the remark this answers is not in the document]
derekbit: I see. Probably can add a comment it for letting others know the `Size()==0` is due to some errors.

pkg/replica/replica.go
```diff
@@ -232,6 +236,37 @@ func construct(readonly bool, size, sectorSize int64, dir, head string, backingF
...
 }
 
+func (r *Replica) checkVolumeMetaFile() (err error) {
+	filePath := r.diskPath(volumeMetaData)
+	fileInfo, err := os.Stat(filePath)
+	if err != nil && !os.IsNotExist(err) {
+		return err
+	}
+
+	if fileInfo != nil && fileInfo.Size() != 0 {
```
innobead, line 246: Don't just use `fileInfo.Size()` to determine if this needs to recover, but check if the content can be decoded to `replica.Info`, so this recovery method can tackle all cases w/ invalid content.

pkg/replica/replica.go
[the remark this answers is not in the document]
innobead: note: in general, for any `intended` operation to tackle some unexpected cases, we should have a log to make it explicit.

pkg/replica/replica.go
mantissahz: Moved. still need to check `io.EOF` caused by empty meta file at server.Status() but it will not delete the file, handle empty file in `tryRecoverVolumeMetaFile`

pkg/replica/replica.go
mantissahz: Modified. And changed the method name.

pkg/replica/replica.go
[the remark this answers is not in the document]
mantissahz: Removed this check. Metadata file will be recovered before `readMetadata`

pkg/replica/server.go
mantissahz: Modified. Thanks.

pkg/replica/server.go
mantissahz: Removed `os.Remove(volumeMetaFilePath)`. Thanks.

pkg/replica/replica.go
mantissahz: Modified, use `r.unmarshalFile()` to check if the content can be decoded to `replica.Info` but only check metadata file is empty or does not exist now

pkg/replica/replica.go
```diff
@@ -988,6 +994,37 @@ func (r *Replica) readMetadata() (bool, error) {
...
+
+		for _, file := range files {
+			if strings.Contains(file.Name(), types.VolumeHeadName) {
+				r.info.Head = file.Name()
+				break
+			}
+		}
+	}
+
+	logrus.Warnf("Recover volume metafile %v", filePath)
```
innobead, line 1024: `Recovering ...`

pkg/replica/server.go
```diff
@@ -113,7 +115,21 @@ func (s *Server) Status() (State, Info) {
...
+
+			fileInfo, statErr := os.Stat(volumeMetaFilePath)
+			if statErr != nil {
+				logrus.Errorf("Failed to read info in replica directory %s: %v", s.dir, err)
+				return Error, Info{}
+			}
+
+			if fileInfo.Size() == 0 {
+				return Initial, Info{}
+			}
```
innobead, line 130: Suggest change to below.

```go
replica := Replica{dir: s.dir}
if er := replica.checkValidVolumeMetaData(); err != nil {
    return Initial, Info{}
}
```

In `checkValidVolumeMetaData`, you can try to decode the file to see if it's valid.

pkg/replica/replica.go
```diff
@@ -988,6 +994,37 @@ func (r *Replica) readMetadata() (bool, error) {
...
 }
 
+func (r *Replica) tryRecoverVolumeMetaFile(head string) error {
+	filePath := r.diskPath(volumeMetaData)
+
+	err := r.unmarshalFile(filePath, &r.info)
+	// recover metadata file that does not exist or is empty
+	if !os.IsNotExist(err) && !errors.Is(err, io.EOF) {
+		return err
+	}
```
innobead, line 1004: Replace by `checkValidVolumeMetaData` and if not valid, try to do the following recovery.

pkg/replica/replica.go
mantissahz: Modified. Thanks.

pkg/replica/server.go
mantissahz: Modified. Add `checkValidVolumeMetaData` to check if volume metadata file is valid.

pkg/replica/replica.go
mantissahz: Replaced. Thanks.

innobead (changes requested):

pkg/replica/replica.go
```diff
@@ -988,6 +994,51 @@ func (r *Replica) readMetadata() (bool, error) {
...
+		r.info.Head = head
+	}
+
+	if r.info.Head == "" {
+		files, err := ioutil.ReadDir(r.dir)
+		if err != nil {
+			return err
+		}
+
+		for _, file := range files {
```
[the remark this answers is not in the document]
mantissahz: No. Do you think `for _, file := range files {` should align with which indentation?

pkg/replica/replica.go
```diff
@@ -988,6 +994,51 @@ func (r *Replica) readMetadata() (bool, error) {
...
+			if strings.Contains(file.Name(), types.VolumeHeadName) {
+				r.info.Head = file.Name()
+				break
+			}
+		}
+	}
+
+	logrus.Warnf("Recovering volume metafile %v", r.diskPath(volumeMetaData))
```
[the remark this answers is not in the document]
mantissahz: Added. Thanks.

pkg/replica/server.go
```diff
@@ -113,7 +113,18 @@ func (s *Server) Status() (State, Info) {
 		info, err := ReadInfo(s.dir)
 		if os.IsNotExist(err) {
 			return Initial, Info{}
-		} else if err != nil {
+		}
+
+		replica := Replica{dir: s.dir}
+		volumeMetaFileValid, vaildErr := replica.checkValidVolumeMetaData()
+		if vaildErr != nil {
+			return Error, Info{}
```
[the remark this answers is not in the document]
mantissahz: Added. Thanks.

pkg/replica/replica.go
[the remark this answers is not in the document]
innobead: Oh probably misread. All good.

pkg/replica/replica.go
```diff
@@ -988,6 +994,50 @@ func (r *Replica) readMetadata() (bool, error) {
...
+		}
+		for _, file := range files {
+			if strings.Contains(file.Name(), types.VolumeHeadName) {
+				r.info.Head = file.Name()
+				break
+			}
+		}
+	}
+
+	logrus.WithField("replicaInfo", r.info).Warnf("Recovering volume metafile %v", r.diskPath(volumeMetaData))
```
innobead, line 1023: Log with fields is more for general info over the scope instead.

Could you just use %v to print r.info part of the message?

innobead (changes requested):
LGTM only one left.

pkg/replica/replica.go
mantissahz: Yes, Modified. Thanks.

innobead (approved):

PR #777, opened 2022-12-02
Fix the misleading log messages

derekbit:
[Longhorn 4980](https://github.com/longhorn/longhorn/issues/4980)

### DIFF
--- a/pkg/replica/replica.go
+++ b/pkg/replica/replica.go
@@ -352,14 +352,15 @@ func (r *Replica) hardlinkDisk(target, source string) error {
 	}
 
 	if _, err := os.Stat(r.diskPath(target)); err == nil {
-		logrus.Infof("Old file %s exists, deleting", target)
+		logrus.Infof("Removing old file %s", target)
 		if err := os.Remove(r.diskPath(target)); err != nil {
 			return errors.Wrapf(err, "failed to remove %s", target)
 		}
 	}
 
+	logrus.Infof("Hard linking %v to %v", source, target)
 	if err := os.Link(r.diskPath(source), r.diskPath(target)); err != nil {
-		return fmt.Errorf("failed to link %s to %s", source, target)
+		return fmt.Errorf("failed to hard link %s to %s", source, target)
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


pkg/replica/replica.go
```diff
@@ -757,12 +758,19 @@ func (r *Replica) rmDisk(name string) error {
 		return nil
 	}
 
+	logrus.Infof("Removing disk %v", name)
+
 	lastErr := os.Remove(r.diskPath(name))
+	if lastErr != nil {
+		logrus.WithError(lastErr).Errorf("Failed to remove dis file %v", name)
```
innobead, line 765: `Failed to remove disk file `

pkg/replica/replica.go
```diff
@@ -757,12 +758,19 @@ func (r *Replica) rmDisk(name string) error {
...
+	}
 	if err := os.Remove(r.diskPath(name + diskutil.DiskMetadataSuffix)); err != nil {
 		lastErr = err
+		logrus.WithError(lastErr).Errorf("Failed to remove disk metadata file %v", name+diskutil.DiskMetadataSuffix)
```
innobead, line 769: Can we do this way?

```go
    diskMetaPath := r.diskPath(name + diskutil.DiskMetadataSuffix)
	if err := os.Remove(diskMetaPath); err != nil {
		lastErr = err
		logrus.WithError(lastErr).Errorf("Failed to remove disk metadata file %v", diskMetaPath)
	}
	
	diskChecksumPath = r.diskPath(name + diskutil.DiskChecksumSuffix)
	if err := os.RemoveAll(diskChecksumPath); err != nil {
		lastErr = err
		logrus.WithError(lastErr).Errorf("Failed to remove disk checksum file %v", diskChecksumPath)
	}
```

innobead (changes requested):
In general, LGTM. Some comments needed to resolve.

pkg/replica/replica.go
derekbit: Updated. Thank you.

innobead (approved):

PR #779, opened 2022-12-02
[BACKPORT][v1.2.7] fix(replica): volume metafile deleted or empty

mantissahz:
Volume will be stuck in attaching and detaching loop when attaching a volume with volume metafile deleted or empty.

Try to rebuild the volume metafile when attaching a volume.

longhorn/longhorn#4846
https://github.com/longhorn/longhorn/issues/4983

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -34,6 +34,8 @@ const (
 
 	EngineFrontendBlockDev = "tgt-blockdev"
 	EngineFrontendISCSI    = "tgt-iscsi"
+
+	VolumeHeadName = "volume-head"
 )
 
 type ReaderWriterAt interface {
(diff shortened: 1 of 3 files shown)


pkg/types/types.go
```diff
@@ -34,6 +34,8 @@ const (
 
 	EngineFrontendBlockDev = "tgt-blockdev"
 	EngineFrontendISCSI    = "tgt-iscsi"
+
+	VolumeHeadName = "volume-head"
```
innobead, line 38: The original commit doesn't have this change. I guess you did not use cherry-pick; otherwise, it should be a cherry-pick commit in the message.

mantissahz: Actually I do use `cherry-pick` but this change `VolumeHeadName = "volume-head"` https://github.com/longhorn/longhorn-engine/commit/64a19ede4aa457863c07226eba77b1baadf193ef did not backport to v1.3.x and v1.2.x, then I added it manually.

PR #780, opened 2022-12-05
Return exit code ENODATA(61) if there is no backend when starting controller

derekbit:
[Longhorn 1330](https://github.com/longhorn/longhorn/issues/1330)
    
### DIFF
--- a/app/cmd/controller.go
+++ b/app/cmd/controller.go
@@ -1,7 +1,9 @@
 package cmd
 
 import (
-	"log"
+	"os"
+	"strings"
+	"syscall"
 	"time"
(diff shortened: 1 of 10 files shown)


innobead (approved):
LGTM

For API error handling between caller and callee, it's better to rely on returned code or constant variable/enum which can be reused on the caller side to verify the returned error. 

It seems here what we return is mostly implementing the error interface, so the constant-var validation seems simpler from a refactoring perspective.  (This refactoring can be planned in the future)

derekbit:
@Mergify backport v1.3.x

PR #781, opened 2022-12-09
Online expansion support

shuo-wu:
longhorn/longhorn#1674

### DIFF
--- a/pkg/controller/control.go
+++ b/pkg/controller/control.go
@@ -284,6 +284,9 @@ func (c *Controller) startExpansion(size int64) (err error) {
 	if c.hasWOReplica() {
 		return fmt.Errorf("cannot do expansion since there is WO(rebuilding) replica")
 	}
+	if size%diskutil.VolumeSectorSize != 0 {
+		return fmt.Errorf("failed to expand volume size %v, because it is not multiple of volume sector size %v", size, diskutil.VolumeSectorSize)
+	}
 	if c.size > size {
 		return fmt.Errorf("controller cannot be expanded to a smaller size %v", size)
 	} else if c.size == size {
--- a/pkg/controller/control.go
+++ b/pkg/controller/control.go
@@ -213,6 +213,14 @@ func (c *Controller) Expand(size int64) error {
 	go func(size int64) {
 		expanded := false
 		defer func() {
+			// Frontend expansion involves in the iSCSI session rescanning, which will wait for the in-fly io requests complete.
+			// Hence there will be a deadlock once we use the lock to protect this frontend expansion.
+			if c.frontend != nil {
+				if err := c.frontend.Expand(size); err != nil {
+					logrus.WithError(err).Error("Failed to expand the frontend")
+					expanded = false
+				}
+			}
 			c.finishExpansion(expanded, size)
 		}()
 
@@ -252,13 +260,6 @@ func (c *Controller) Expand(size int64) error {
 			}
 		}
 
-		if c.frontend != nil {
-			if err := c.frontend.Expand(size); err != nil {
-				logrus.WithError(err).Error("Failed to expand the frontend")
-				return
-			}
-		}
-
 		expanded = true
 	}(size)
 
@@ -288,9 +289,6 @@ func (c *Controller) startExpansion(size int64) (err error) {
 	} else if c.size == size {
 		return fmt.Errorf("controller %v is already expanded to size %v", c.Name, size)
 	}
-	if c.frontend != nil && c.frontend.State() == types.StateUp {
-		return fmt.Errorf("controller %v doesn't support on-line expansion, frontend: %v", c.Name, c.frontend.FrontendName())
-	}
 
 	c.isExpanding = true
 	return nil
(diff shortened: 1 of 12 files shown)


pkg/controller/control.go
```diff
@@ -213,6 +213,14 @@ func (c *Controller) Expand(size int64) error {
 	go func(size int64) {
 		expanded := false
 		defer func() {
+			// Frontend expansion involves in the iSCSI session rescanning, which will wait for the in-fly io requests complete.
+			// Hence there will be a deadlock once we use the lock to protect this frontend expansion.
+			if c.frontend != nil {
+				if err := c.frontend.Expand(size); err != nil {
+					logrus.WithError(err).Error("Failed to expand the frontend")
+					expanded = false
+				}
```
derekbit, line 222: return here if any error?

shuo-wu: No. We need to rely on the below function `c.finishExpansion(expanded, size)` to update flag `c.isExpanding`

pkg/controller/control.go
```diff
@@ -213,6 +213,14 @@ func (c *Controller) Expand(size int64) error {
...
+			}
```
innobead, line 223: Any potential side effect if no lock?

pkg/controller/control.go
```diff
@@ -283,14 +284,14 @@ func (c *Controller) startExpansion(size int64) (err error) {
 	if c.hasWOReplica() {
 		return fmt.Errorf("cannot do expansion since there is WO(rebuilding) replica")
 	}
+	if size%diskutil.VolumeSectorSize != 0 {
+		return fmt.Errorf("controller expanding size %v is not multiple for volume sector size %v", size, diskutil.VolumeSectorSize)
```
innobead, line 288: `"failed to expand volume size %v, because it is not multiple of volume sector size %v"`

pkg/replica/replica.go
```diff
@@ -1136,6 +1136,10 @@ func (r *Replica) SetUnmapMarkDiskChainRemoved(enabled bool) {
 }
 
 func (r *Replica) Expand(size int64) (err error) {
+	if size%diskutil.VolumeSectorSize != 0 {
+		return fmt.Errorf("replica expanding size %v is not multiple for volume sector size %v", size, diskutil.VolumeSectorSize)
```
innobead, line 1140: `failed to expend volume replica size %v, because it is not multiple of volume sector size %v"`

innobead (changes requested):
In general, LGTM. One question and a few message format change requests.

pkg/controller/control.go
innobead: cc @derekbit

shuo-wu: The frontend expansion consists of 2 parts:
1. The target LUN expansion. We just need to update the field `size` for the LUN. No data flow gets involved. This should be fine.
2. The iSCSI initiator refresh/rescan. The iSCSI should handle the data flow by itself rather than requiring the vendor/users to control it. And as I mentioned, adding the engine lock for this would lead to the deadlock issue.

Hence I don't think we need a lock here.

innobead: Clarified, all good.

pkg/controller/control.go
derekbit: Got it. Thank you.

pkg/controller/control.go
```diff
@@ -301,6 +302,9 @@ func (c *Controller) finishExpansion(expanded bool, size int64) {
 	c.Unlock()
```
derekbit, line 302:
```suggestion
	defer c.Unlock()
```

derekbit: BTW, we need to backport the fix to v1.3.x and v1.2.x.

shuo-wu: Sure. Will do it after the expansion PR merge.

innobead (approved):
LGTM

pkg/controller/control.go
PhanLe1010: Will this one be a halfway expansion? 

Next time when volume detaches and reattaches, will the size of replicas be different than the size of the engine?

pkg/controller/control.go
```diff
@@ -213,6 +213,14 @@ func (c *Controller) Expand(size int64) error {
 	go func(size int64) {
 		expanded := false
 		defer func() {
+			// Frontend expansion involves in the iSCSI session rescanning, which will wait for the in-fly io requests complete.
+			// Hence there will be a deadlock once we use the lock to protect this frontend expansion.
```
PhanLe1010, line 217: Could you please explain more about how a deadlock would happen here? Thank you

shuo-wu: 1. The online expansion flow adds a lock for this frontend expansion.
1. New r/w requests will be blocked once reaching the controller.
2. The frontend expansion will try to rescan/refresh the iSCSI initiator, which actually needs to flush and wait for all in-fly r/w requests. But there are some requests getting stuck (in step 2), deadlock

pkg/controller/control.go
shuo-wu: Actually there is a separate issue for this: longhorn/longhorn#3601.
The key point is that we cannot expand the frontend if there is no replica succeeding the expansion. Otherwise, you will see tgt r/w errors, which may need to data loss/

pkg/controller/control.go
PhanLe1010: Ah. Do you mean if we do [this](https://github.com/shuo-wu/longhorn-engine/blob/7376a90ceeb70caf205e10352c40dae054dd5dae/pkg/controller/control.go#L239-L240) before [this](https://github.com/shuo-wu/longhorn-engine/blob/7376a90ceeb70caf205e10352c40dae054dd5dae/pkg/controller/control.go#L215), the deadlock will happen?

PhanLe1010: Thank you for the explantion!

PR #794, opened 2022-12-14
Support metrics for IOPS and bandwidth real time

derekbit:
Longhorn/longhorn#2406


pkg/types/types.go
```diff
@@ -152,8 +152,8 @@ const (
 )
 
 type Metrics struct {
-	Bandwidth    RWMetrics // in byte
-	TotalLatency RWMetrics // in microsecond(us)
+	Throughput   RWMetrics // in byte
+	TotalLatency RWMetrics // in nanoseconds
```
shuo-wu, line 156: Using a more accurate naming and a finer unit is good. But will these lead to compatibility issues?

pkg/controller/control.go
```diff
@@ -1078,12 +1083,35 @@ func (c *Controller) metricsStart() {
...
+			c.metricsLock.Unlock()
 		}
 	}()
 }
 
-func (c *Controller) GetLatestMetics() *types.Metrics {
-	return c.latestMetrics
+func (c *Controller) GetLatestMetics() *ptypes.Metrics {
+	c.metricsLock.RLock()
+	defer c.metricsLock.RUnlock()
```
shuo-wu, line 1096: Will this impact the rw performance? Almost No?

shuo-wu (commented):
Actually, I have an extra question for the metrics part. 
The operation trim currently is not recorded by the metrics, should we add it now? If YES, should we launch a separate option or just consider trim as a kind of special write?

pkg/types/types.go
derekbit: I've checked the MetricGet method. It's not integrated into longhorn-manager, so there is not compatibility issue, right?

pkg/controller/control.go
derekbit: I think no. I think it is a memory copy operation. But, I can benchmark and check it.

derekbit:
> Actually, I have an extra question for the metrics part. The operation trim currently is not recorded by the metrics, should we add it now? If YES, should we launch a separate option or just consider trim as a kind of special write?

I'm fine with this idea. And, we can open a new ticket for the trim part. WDYT? 
cc @innobead

pkg/controller/control.go
shuo-wu: Each r/w request received by the controller will call `recordMetrics`, it may affect the latency a little bit. You can have a simple benchmark if needed.

derekbit: Sure. Will update the result later.

shuo-wu:
> I'm fine with this idea, but I rarely see the trim metric in other projects.
We can open a new ticket for the trim part or implement it in this ticket. WDYT?

Sure. We can just ignore recording this operation or handle it separately.

integration/tox.ini
```diff
@@ -4,11 +4,6 @@ envlist=flake8, py3
 [testenv]
 deps=-rrequirements.txt
 changedir={toxinidir}
-commands=py.test core data instance --durations=20 {posargs}
+commands=py.test -vs core/test_controller.py::test_metrics
```
innobead, line 7: Should not change this to ignore the original tests.

integration/core/test_controller.py
```diff
@@ -163,18 +163,21 @@ def test_controller_expand(grpc_controller_client):  # NOQA
     assert v.replicaCount == 0
 
 
-def test_metric(grpc_controller_client):  # NOQA
-    replies = grpc_controller_client.metric_get()
-
+def test_metrics(grpc_controller_client):  # NOQA
```
innobead, line 166: This test looks weird. Why there will be an exception probably happened? Also, why 5 iterations.

integration/data/test_basic_ops.py
```diff
@@ -42,31 +44,35 @@ def test_basic_rw(dev):  # NOQA
...
-                assert metric.iOPS != 0
+                assert metrics.readThroughput != 0
+                assert metrics.writeThroughput != 0
+                assert metrics.readIOPS != 0
+                assert metrics.writeIOPS != 0
                 break
-            except StopIteration:
-                time.sleep(1)
+            except Exception as e:
+                if iter < METRICS_RETRY_COUNTS:
```
innobead, line 72: Why need retry count? is it because the runtime record not updated yet?

integration/data/test_basic_ops.py
```diff
@@ -42,31 +44,35 @@ def test_basic_rw(dev):  # NOQA
...
+def test_rw_with_metric(grpc_replica1, grpc_replica2, grpc_controller):  # NOQA
+    rw_dev = get_dev(grpc_replica1, grpc_replica2,
+                     grpc_controller)
 
-    replies = grpc_controller.metric_get()
-    # skip the first metric since its fields are 0
-    next(replies).metric
-
-    for i in range(0, 5):
+    for i in range(0, 1):
```
innobead, line 51: Why need two iteration?

integration/data/test_basic_ops.py
```diff
@@ -18,6 +18,8 @@
     VOLUME_NAME, ENGINE_NAME
 )
 
+METRICS_RETRY_COUNTS = 10
+METRICS_RETRY_INTERVAL = 0.1
```
innobead, line 22: Only used in the metric test function, so it's OK to declare there.

pkg/controller/control.go
```diff
@@ -1078,12 +1083,35 @@ func (c *Controller) metricsStart() {
 	go func() {
 		for {
 			time.Sleep(1 * time.Second)
+			c.metricsLock.Lock()
```
innobead, line 1086: Why we need `metricsStart` instead of keeping `c.metrics` only w/o `c.latestMetrics`?

pkg/controller/control.go
innobead: Why not just deep copy `c.metrics` here w/o read lock, then return the calculated copied?

pkg/controller/control.go
```diff
@@ -1062,14 +1065,16 @@ func (c *Controller) FrontendState() string {
 }
 
 func (c *Controller) recordMetrics(isRead bool, dataLength int, latency time.Duration) {
```
innobead, line 1067: Because `recordMetrics` is part of write and read method, will it impact the IO? If yes, probably we should record metrics asynchronously out from write/read like using a goruntine to do that. (metrics can be best efforts)

innobead (changes requested):

integration/core/test_controller.py
derekbit: I guess the purpose of this test is to check the values are always 0 if there is no IO. I feel it is not necessary.
I feel `data/test_basic_ops.py::test_rw_with_metric` is enough. I prefer to remove this test. WDYT?

innobead: Agreed.

integration/data/test_basic_ops.py
derekbit: https://github.com/longhorn/longhorn-engine/pull/794#discussion_r1050728159

integration/data/test_basic_ops.py
innobead: No, I mean what is the original case having 5 iterations?

pkg/controller/control.go
derekbit: > Because recordMetrics is part of write and read method, will it impact the IO? 

According to my experience, I think no. I'm running the benchmark, wll update it later.

pkg/controller/control.go
derekbit: Need to calculate the average latency here, not just copy here.

integration/data/test_basic_ops.py
derekbit: Well...I think just the first iteration is to check the previous grpc streaming works as expected. It can be removed here, because we don't use streaming here.

The second iteration is due to that the metrics are not updated yet.

integration/core/test_controller.py
derekbit: Removed.

integration/data/test_basic_ops.py
derekbit: Updated the test case.

integration/data/test_basic_ops.py
derekbit: Removed. Thank you.

pkg/controller/control.go
derekbit: Don't see the performance difference between the two cases.

**Enable metrics**
```
        Random:          32,123 / 15,537
    Sequential:          29,562 / 27,833

Bandwidth in KiB/sec (Read/Write)
        Random:        347,891 / 413,658
    Sequential:        451,056 / 434,144

Latency in ns (Read/Write)
        Random:        458,593 / 393,689
    Sequential:        379,152 / 400,432
```

**Disable metrics**
```
IOPS (Read/Write)
        Random:          32,320 / 15,545
    Sequential:          29,348 / 27,903

Bandwidth in KiB/sec (Read/Write)
        Random:        348,873 / 414,618
    Sequential:        451,903 / 435,478

Latency in ns (Read/Write)
        Random:        461,520 / 397,417
    Sequential:        378,810 / 402,162
```

cc @shuo-wu @innobead

innobead:
@derekbit the first commit `[vendor: update MetricGet to MetricsGet](https://github.com/longhorn/longhorn-engine/pull/794/commits/fff43fd8230cf4394db99d4a8f06b3e2eadc37bd)` is about proto update instead of vendor update.

proto/ptypes/controller.proto
```diff
@@ -136,14 +136,15 @@ message VersionDetailGetReply {
...
-    uint64 writeBandwidth = 2;
+message Metrics {
+    uint64 readThroughput = 1;
+    uint64 writeThroughput = 2;
     uint64 readLatency = 3;
     uint64 writeLatency = 4;
-    uint64 iOPS = 5;
+    uint64 readIOPS = 5;
+    uint64 writeIOPS = 6;
 }
```
innobead, line 146: It's ok to change the properties, because metrics does not be used before. (literally, we should not change existing fields but rather add new ones. )

innobead (approved):
LGTM

innobead:
@derekbit Need to backport to 1.4.x.

PR #795, opened 2022-12-14
Disable tcmalloc

derekbit:
Longhorn/longhorn#5068


innobead (commented):
Then we should backport this to 1.3?

derekbit:
> Then we should backport this to 1.3?

Yes. Will backport it.

derekbit:
Unfortunately, hit the same error although the tcmalloc is disabled...
cc @innobead

PR #796, opened 2022-12-14
Fix the canceling logic and the upgrade related issues for online expansion

shuo-wu:
longhorn/longhorn#1674

### DIFF
--- a/pkg/controller/control.go
+++ b/pkg/controller/control.go
@@ -270,30 +270,33 @@ func (c *Controller) startExpansion(size int64) (err error) {
 	c.Lock()
 	defer c.Unlock()
 
-	defer func() {
-		if err != nil {
-			err = fmt.Errorf("failed to start expansion: %v", err)
-			c.lastExpansionFailedAt = time.Now().UTC().Format(time.RFC3339)
-			c.lastExpansionError = err.Error()
-		}
-	}()
-
 	if c.isExpanding {
 		return fmt.Errorf("controller expansion is in progress")
 	}
-	if c.hasWOReplica() {
-		return fmt.Errorf("cannot do expansion since there is WO(rebuilding) replica")
-	}
+
+	defer func() {
+		if c.isExpanding {
+			c.lastExpansionFailedAt = ""
+			c.lastExpansionError = ""
+		} else if err != nil {
+			c.lastExpansionFailedAt = time.Now().String()
+			c.lastExpansionError = errors.Wrap(err, "controller failed to start expansion").Error()
+		}
+	}()
+
 	if size%diskutil.VolumeSectorSize != 0 {
-		return fmt.Errorf("failed to expand volume size %v, because it is not multiple of volume sector size %v", size, diskutil.VolumeSectorSize)
+		return fmt.Errorf("requested expansion size %v not multiple of volume sector size %v", size, diskutil.VolumeSectorSize)
+	}
+	if c.size == size {
+		logrus.Infof("controller %v is already expanded to size %v", c.Name, size)
+		return nil
 	}
 	if c.size > size {
 		return fmt.Errorf("controller cannot be expanded to a smaller size %v", size)
-	} else if c.size == size {
-		return fmt.Errorf("controller %v is already expanded to size %v", c.Name, size)
 	}
 
 	c.isExpanding = true
+
 	return nil
 }
 
--- a/pkg/replica/replica.go
+++ b/pkg/replica/replica.go
@@ -590,7 +590,8 @@ func (r *Replica) writeVolumeMetaData(dirty, rebuilding bool) error {
 	info := r.info
 	info.Dirty = dirty
 	info.Rebuilding = rebuilding
-	return r.encodeToFile(&info, volumeMetaData)
+	_, err := r.encodeToFile(&info, volumeMetaData)
+	return err
 }
 
 func (r *Replica) isBackingFile(index int) bool {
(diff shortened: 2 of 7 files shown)


pkg/replica/replica.go
```diff
@@ -667,39 +665,38 @@ func (r *Replica) openFile(name string, flag int) (types.DiffDisk, error) {
...
-				if err := os.Remove(r.diskPath(newHeadName)); err != nil {
-					rollbackErr = err
-				}
-			}
-			if _, err := os.Stat(r.diskPath(newHeadName + diskutil.DiskMetadataSuffix)); err == nil {
-				if err := os.Remove(r.diskPath(newHeadName + diskutil.DiskMetadataSuffix)); err != nil {
-					rollbackErr = types.CombineErrors(rollbackErr, err)
-				}
+		rollbackFunc = func() error {
+			f.Close()
```
derekbit, line 686: Does `f.Close()` need to be in the `rollbackFunc`?

shuo-wu: If everything works fine, `f` will be added into `r.volume.files`. In this case, we don't need to close it.
Otherwise, we need to close it before the rollback cleanup.

derekbit: Got it. Thanks for the clarification.

derekbit (approved):
LGTM

pkg/replica/replica.go
```diff
@@ -613,40 +614,37 @@ func (r *Replica) close() error {
...
+			return os.RemoveAll(r.diskPath(tmpFileName))
+		}
+
 		if err != nil {
-			if _, err := os.Stat(r.diskPath(tmpFileName)); err == nil {
-				if err := os.Remove(r.diskPath(tmpFileName)); err != nil {
-					rollbackErr = err
-				}
-			}
+			err = types.GenerateFunctionErrorWithRollback(err, os.RemoveAll(r.diskPath(tmpFileName)))
```
innobead, line 630:
```go
			err = types.GenerateFunctionErrorWithRollback(err, rollbackFunc())
```

pkg/replica/replica.go
```diff
@@ -613,40 +614,37 @@ func (r *Replica) close() error {
...
 		}
-		err = types.GenerateFunctionErrorWithRollback(err, rollbackErr)
 	}()
 
 	f, err := os.Create(r.diskPath(tmpFileName))
 	if err != nil {
-		return err
+		return rollbackFunc, errors.Wrapf(err, "failed to create the tmp file for file %v", r.diskPath(file))
```
innobead, line 636: After returning the error, actually `tmpFileName` has been deleted in the defer function. Knowing you want to accumulate the errors, but why not just use the returned error?

pkg/replica/replica.go
```diff
@@ -709,38 +706,53 @@ func (r *Replica) createNewHead(oldHead, parent, created string, size int64) (f
 		UserCreated: false,
 		Created:     created,
 	}
-	err = r.encodeToFile(&newDisk, newHeadName+diskutil.DiskMetadataSuffix)
-	return f, newDisk, err
+	subRollbackFunc, err = r.encodeToFile(&newDisk, newHeadName+diskutil.DiskMetadataSuffix)
```
innobead, line 709: As I mentioned above, what `subRollbackFun` will do has been done in the defer function of `encodeToFile`, so you should use the returned error.

```go
types.CombineErrors(errorHappenedBeforeDefer, r.rmDisk(newHeadName))
```

innobead (changes requested):

pkg/replica/replica.go
```diff
@@ -709,38 +706,53 @@ func (r *Replica) createNewHead(oldHead, parent, created string, size int64) (f
...
-		return nil
+func (r *Replica) linkDisk(oldName, newName string) (rollbackFunc func() error, err error) {
+	if oldName == "" {
+		return nil, nil
 	}
 
-	destMetadata := r.diskPath(newname + diskutil.DiskMetadataSuffix)
-	logrus.Infof("Deleting old disk metadata file %v", destMetadata)
+	defer func() {
+		rollbackFunc = func() error {
```
innobead, line 719: This rollbackFunc is ok, but it has not been executed after `linkDisk`.

pkg/replica/replica.go
shuo-wu: 1. Yes, the error will be accumulated. It's for helping understand the whole failure chain.
2. The rollback functions are not for accumulating errors. Instead, they allow the upper caller to be atomic: the upper layer either succeeds to execute all functions, or fails in the middle then does rollback for the previous succeeded parts so that everything looks like unchanged.
  1. If this kind of function failed, we will do rollback at the end of (in the defer function) the current function. Then the upper layer is typically no longer need to re-run the rollback function.
  2. In other words, the returned rollback function will be executed (by the upper layer) when the **current function succeeded** but the **subsequent executions failed**.

innobead: oh I think I missed that purpose. 

The point is `In other words, the returned rollback function will be executed (by the upper layer) when the current function succeeded but the subsequent executions failed.`.

pkg/replica/replica.go
innobead: Clarified by @shuo-wu

shuo-wu: Oh Yes. Here `subRollbackFunc` will be executed twice if this `r.encodeToFile` failed...
I am thinking we can set `rollbackFunc` to `nil` when the original function errors out. Then we don't need to worry about the duplicate execution issued by the upper layer.

innobead: Sounds good to me.

innobead (commented):
In general, LGTM. Waitinf for minor fixes, then we can merge this.

pkg/controller/control.go
```diff
@@ -270,30 +270,33 @@ func (c *Controller) startExpansion(size int64) (err error) {
...
+			c.lastExpansionError = errors.Wrap(err, "controller failed to start expansion").Error()
 		}
 	}()
 
-	if c.isExpanding {
-		return fmt.Errorf("controller expansion is in progress")
-	}
-	if c.hasWOReplica() {
-		return fmt.Errorf("cannot do expansion since there is WO(rebuilding) replica")
-	}
```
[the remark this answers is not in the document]
shuo-wu: Right now online expansion is like a special snapshotting. Hence it's fine to allow it. See the modified test `test_ha_recovery_with_expansion` in [the PR](https://github.com/longhorn/longhorn-tests/pull/1223/files) for details

innobead (approved):

PR #799, opened 2022-12-16
[BACKPORT][v1.3.3] Disable tcmalloc

derekbit:
Longhorn/longhorn#5096
https://github.com/longhorn/longhorn/issues/5097


innobead:
Build failure is due to `ioutil` removed from https://go.dev/doc/go1.16#ioutil, but we still rely on `go get github.com/rancher/trash`, so we need to update the build or just update the go version in 1.3. It should be fine for the second. (let's tackle this PR later)

image

PR #802, opened 2022-12-21
[BACKPORT][v1.3.3] fix(replica): Replica must be closed when rebuild

mantissahz:
Replica banckend needs to be closed if the number of snapshots is up to the limitation and replica rebuiding is occurred.

Ref: longhorn/longhorn#3828, https://github.com/longhorn/longhorn/issues/5116

### DIFF
--- a/pkg/controller/control.go
+++ b/pkg/controller/control.go
@@ -181,10 +181,8 @@ func (c *Controller) Snapshot(name string, labels map[string]string) (string, er
 		name = util.UUID()
 	}
 
-	if remain, err := c.backend.RemainSnapshots(); err != nil {
+	if _, err := c.backend.RemainSnapshots(); err != nil {
 		return "", err
-	} else if remain <= 0 {
-		return "", fmt.Errorf("too many snapshots created")
 	}
 
 	created := util.Now()
@@ -221,11 +219,8 @@ func (c *Controller) Expand(size int64) error {
 		// Should block R/W during the expansion.
 		c.Lock()
 		defer c.Unlock()
-		if remain, err := c.backend.RemainSnapshots(); err != nil {
+		if _, err := c.backend.RemainSnapshots(); err != nil {
 			logrus.Errorf("Cannot get remain snapshot count before expansion: %v", err)
-			return
-		} else if remain <= 0 {
-			logrus.Errorf("Cannot do expansion since too many snapshots created")
 			return
 		}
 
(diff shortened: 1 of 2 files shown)


Dockerfile.dapper
```diff
@@ -43,7 +43,7 @@ RUN if [ ${ARCH} == "s390x" ]; then \
 ENV GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm64=arm64 GOLANG_ARCH_s390x=s390x GOLANG_ARCH=GOLANG_ARCH_${ARCH} \
     GOPATH=/go PATH=/go/bin:/usr/local/go/bin:${PATH} SHELL=/bin/bash
 RUN wget -O - https://storage.googleapis.com/golang/go1.14.15.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local && \
-    go get github.com/rancher/trash && go get -u golang.org/x/lint/golint
+    go get github.com/rancher/trash
```
innobead, line 46: Why remote golint?

mantissahz: Changed back, and keep it failed when getting `golint`, It should be a go version issue.

innobead:
Error is as below because we always get the latest of `golang.org/x/lint/golint` which is using the new `io.ReadAll` but it's only available since Go 1.16. 

Two ways:
- Upgrade go version (don't prefer this for stable branch)
- Replace golint with https://golangci-lint.run/
- Go get the specific commit of golint

```go
# golang.org/x/tools/go/gcexportdata
/go/src/golang.org/x/tools/go/gcexportdata/gcexportdata.go:97:9: undefined: io.ReadAll
The command '/bin/sh -c wget -O - https://storage.googleapis.com/golang/go1.14.15.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local &&     go get github.com/rancher/trash && go get -u golang.org/x/lint/golint' returned a non-zero code: 2
time="2022-12-21T05:31:58Z" level=fatal msg="exit status 2"
[0m
```

innobead:
> Error is as below because we always get the latest of `golang.org/x/lint/golint` which is using the new `io.ReadAll` but it's only available since Go 1.16.
> 
> Two ways:
> 
> * Upgrade go version (don't prefer this for stable branch)
> * Replace golint with [golangci-lint.run](https://golangci-lint.run/)
> * Go get the specific commit of golint

@mantissahz as discussed, please try to the thrid first by having another PR for 1.3. branch. Thanks.

innobead (approved):

PR #805, opened 2022-12-21
chore(image): getting golint failed when building image

mantissahz:
Move the GO111MODULE=on forward to getting golint, and fixed the golint version.


innobead:
@mantissahz Can we make the command inline in the previous RUN instead of having a separate run layer?

mantissahz:
>Can we make the command inline in the previous RUN instead of having a separate run layer?

Possibly NOT, `GO111MODULE=on` will cause `go get github.com/rancher/trash`
Let me find out why it happened.

innobead (approved):

innobead:
@mantissahz Please fix this to the 1.3.x branch of other repos as well. Thanks.

PR #808, opened 2022-12-23
Don't mark the backing image as removed

derekbit:
Longhorn/longhorn#5129

Also, the fix make the trim operation supports volumes with backing images. (Longhorn/longhorn#5128)

### DIFF
--- a/pkg/replica/replica.go
+++ b/pkg/replica/replica.go
@@ -1265,7 +1265,12 @@ func (r *Replica) UnmapAt(length uint32, offset int64) (n int, err error) {
 		// For list `unmappableDisks`, the first entry is the volume head,
 		// the second one is the parent of the first entry...
 		unmappableDisks := []string{r.diskPath(r.activeDiskData[len(r.activeDiskData)-1].Name)}
-		for idx := len(r.activeDiskData) - 2; idx > 0; idx-- {
+		indexOfVolumeHeadParent := len(r.activeDiskData) - 2
+		indexOfLastSnapshotDiskFile := 1
+		if r.isBackingFile(int(backingFileIndex)) {
+			indexOfLastSnapshotDiskFile = int(backingFileIndex) + 1
+		}
+		for idx := indexOfVolumeHeadParent; idx >= indexOfLastSnapshotDiskFile; idx-- {
 			disk := r.activeDiskData[idx]
 			if len(r.diskChildrenMap[disk.Name]) > 1 {
 				break


derekbit:
e2e
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2782/console

pkg/replica/replica.go
```diff
@@ -1265,7 +1265,8 @@ func (r *Replica) UnmapAt(length uint32, offset int64) (n int, err error) {
 		// For list `unmappableDisks`, the first entry is the volume head,
 		// the second one is the parent of the first entry...
 		unmappableDisks := []string{r.diskPath(r.activeDiskData[len(r.activeDiskData)-1].Name)}
-		for idx := len(r.activeDiskData) - 2; idx > 0; idx-- {
+		indexOfVolumeHeadParent := len(r.activeDiskData) - 2
+		for idx := indexOfVolumeHeadParent; idx > int(backingFileIndex); idx-- {
```
innobead, line 1269: We should only skip the first disk (index:1) when the replica is based on a backing image only, right?

derekbit: Yes, only skip index 1 (backing image) and index 0 (empty struct) if a backing image exists.

innobead: yeah, after updating this part, LGTM. cc @shuo-wu for this change/limitation.

innobead (approved):

innobead (approved):
LGTM

pkg/replica/replica.go
```diff
@@ -1265,7 +1265,12 @@ func (r *Replica) UnmapAt(length uint32, offset int64) (n int, err error) {
...
 		// the second one is the parent of the first entry...
 		unmappableDisks := []string{r.diskPath(r.activeDiskData[len(r.activeDiskData)-1].Name)}
-		for idx := len(r.activeDiskData) - 2; idx > 0; idx-- {
+		indexOfVolumeHeadParent := len(r.activeDiskData) - 2
+		indexOfLastSnapshotDiskFile := 1
+		if r.isBackingFile(int(backingFileIndex)) {
+			indexOfLastSnapshotDiskFile = int(backingFileIndex) + 1
+		}
+		for idx := indexOfVolumeHeadParent; idx >= indexOfLastSnapshotDiskFile; idx-- {
 			disk := r.activeDiskData[idx]
```
innobead, line 1274: Do you know why we keep the first of activeDiskData is empty? any context for this current design?

innobead:
@Mergifyio backport v1.4.x

pkg/replica/replica.go
derekbit: No idea from the [commit history](https://github.com/longhorn/longhorn-engine/commit/cb06cb8af65ace8e4b9a861f80a6301bf91a679c#diff-6e6694d5b35c75c9d0850d82cbd57a1a0bf3f9fac601cb25f7dacfa8682fe0a5R121). Keeping the first entry empry seems not necessary from the implementation perspsective.

innobead: yeah, very confusing. Let's keep it as is but can refactor it later.

PR #814, opened 2022-12-29
control: Clear error when the expansion succeeded

shuo-wu:
Longhorn/longhorn#5169

### DIFF
--- a/pkg/controller/control.go
+++ b/pkg/controller/control.go
@@ -310,8 +310,14 @@ func (c *Controller) finishExpansion(expanded bool, size int64) {
 	defer c.Unlock()
 
 	if expanded {
-		logrus.Infof("Controller succeeded to expand from size %v to %v", c.size, size)
+		if c.lastExpansionError != "" {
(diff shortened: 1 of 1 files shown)


innobead (approved):

pkg/controller/control.go
```diff
@@ -310,8 +310,14 @@ func (c *Controller) finishExpansion(expanded bool, size int64) {
 	defer c.Unlock()
 
 	if expanded {
-		logrus.Infof("Controller succeeded to expand from size %v to %v", c.size, size)
+		if c.lastExpansionError != "" {
+			logrus.Infof("Controller succeeded to expand from size %v to %v but there are some replica expansion failures: %v", c.size, size, c.lastExpansionError)
```
innobead, line 314: Should we make the message as below to avoid any confusion about successful expansion?

"Controller finished to expand from ...."

shuo-wu: Word "finished" is vague... And to avoid confusion, I add one more comment "but..."

innobead: Ok, I am fine with that.

innobead:
@Mergifyio backport v1.4.x

PR #816, opened 2022-12-29
Fix the expansion error issue

shuo-wu:
longhorn/longhorn#5169

### DIFF
--- a/pkg/controller/control.go
+++ b/pkg/controller/control.go
@@ -49,8 +49,11 @@ type Controller struct {
 	latestMetrics *types.Metrics
 	metrics       *types.Metrics
 
-	lastExpansionError    string
+	// lastExpansionFailedAt indicates if the error belongs to the recent expansion
 	lastExpansionFailedAt string
+	// lastExpansionError indicates the error message.
+	// It may exist even if the expansion succeeded. For exmaple, some of the replica expansions failed.
+	lastExpansionError string
 
 	fileSyncHTTPClientTimeout int
 }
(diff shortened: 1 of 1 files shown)


pkg/controller/control.go
```diff
@@ -250,7 +253,7 @@ func (c *Controller) Expand(size int64) error {
 
 		expansionSuccess, errsNeedToBeHandled, errsForRecording := c.backend.Expand(size)
 		if errsForRecording != nil {
-			c.lastExpansionFailedAt = time.Now().UTC().Format(time.RFC3339)
+			c.lastExpansionFailedAt = time.Now().UTC().Format(time.RFC3339Nano)
```
innobead, line 256: Why nano? for consistency?

innobead (changes requested):
Don't we still need to do `c.lastExpansionFailedAt = ""` when the expansion succeed?

pkg/controller/control.go
shuo-wu: In the manager part, we will use `time.Now()` to do the comparison. And `time.Now()` is using nanosecond precision.

shuo-wu:
> Don't we still need to do c.lastExpansionFailedAt = "" when the expansion succeed?

Just for recording, so that the manager will print out an event that explains why the replica becomes failed.

pkg/controller/control.go
innobead: I feel we should have a common library to use in every module, so we don't need to remember what time unit we are using.

shuo-wu:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2819/

pkg/controller/control.go
shuo-wu: YES. I was thinking if we need a util repo that covers all kinds of common functions. For example:
1. Rich error marshaling and unmarshaling
2. Address handling. (prefix, format, etc)
3. Timestamp handling.
4. Checksum handling.
5. UUID
......

shuo-wu: YES. I was thinking if we need a util repo that covers all kinds of common functions. For example:
1. Rich error marshaling and unmarshaling
2. Address handling. (prefix, format, etc)
3. Timestamp handling.
4. Checksum handling.
5. UUID
......

shuo-wu: YES. I was thinking if we need a util repo that covers all kinds of common functions. For example:
1. Rich error marshaling and unmarshaling
2. Address handling. (prefix, format, etc)
3. Timestamp handling.
4. Checksum handling.
5. UUID
......

innobead: then, let's create it. https://github.com/longhorn/common-libs

innobead:
@Mergifyio backport v1.4.x

PR #817, opened 2022-12-29
test: empty last error message if volume expansion works

innobead:


innobead:
No need, because we will rely on the original logic, https://github.com/longhorn/longhorn-engine/pull/816.

PR #819, opened 2023-01-03
integration: improve filesystem_trim_test

shuo-wu:
Longhorn/longhorn#5173

### DIFF
--- a/integration/data/test_snapshot.py
+++ b/integration/data/test_snapshot.py
@@ -1,5 +1,6 @@
 import random
 import subprocess
+import time
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


integration/data/test_snapshot.py
```diff
@@ -716,42 +718,45 @@ def filesystem_trim_test(dev, address, volume_name, engine_name, fs_type):  # NO
...
     # create snapshot1 with fs and file 1
     file_path1 = mnt_path + "/test1"
     test1_checksum = write_filesystem_file(length, file_path1)
     subprocess.check_call(nsenter_cmd + ["sync"])
     snap1 = cmd.snapshot_create(address)
 
+    # it is hard to predict the exact blocks the filesystem would consume
+    # when writing a file. but writing a 3Ki file requires at least 3 blocks:
+    # the data itself, inode info update, and free space info (for xfs)
+    info = wait_for_snap_size_no_less_than_value(
```
innobead, line 743: Is the snapshot synchronous? If it's, why need an extra wait here? I know you want to assume the size is expected, but if it's not expected, at the end the test case will still fail in a later step. So, adding the wait here seems not really helpful?

shuo-wu: I am afraid of the data is not completely flushed into the volume head file before snapshotting. But YES you are right, I should check and wait for the volume head file size before snapshotting, rather than do it after snapshotting.

shuo-wu: Updated the PR. Added the volume head size check before snapshot creation, which help guarantee that the data is flushed into the file before snapshotting.

innobead (approved):

innobead:
@Mergifyio backport v1.4.x

PR #820, opened 2023-01-03
Improve backup performance with faster compression algorithm and multiple threads

derekbit:
https://github.com/longhorn/longhorn/issues/5189

Depends on https://github.com/longhorn/backupstore/pull/94


pkg/sync/backup.go
```diff
@@ -53,27 +51,27 @@ func (t *Task) CreateBackup(backupName, snapshot, dest, backingImageName, backin
...
-	replicas, err := t.client.ReplicaList()
+	replica, err := t.findAvailableReplica()
 	if err != nil {
 		return nil, err
 	}
 
+	return t.createBackup(replica, backupName, snapshot, dest, volume.Name, backingImageName, backingImageChecksum, labels, credential)
+}
+
+func (t *Task) findAvailableReplica() (*types.ControllerReplicaInfo, error) {
```
shuo-wu, line 64: NIT: `findRWReplica()`

pkg/backup/main.go
```diff
@@ -22,7 +22,7 @@ var (
 	log     = logrus.WithFields(logrus.Fields{"pkg": "backup"})
 )
 
-type BackupParameters struct {
```
shuo-wu: NIT: Retain the previous naming. This won't be used in the restore function

shuo-wu (dismissed):
LGTM. Good refactoring! 
Don't miss the backward compatibility handling on the manager side.

pkg/backup/main.go
derekbit: Updated. Thank you.

shuo-wu (dismissed):

derekbit:
Dependencies are updated and ready for review.

pkg/backup/main.go
```diff
@@ -21,6 +22,18 @@ var (
 	log     = logrus.WithFields(logrus.Fields{"pkg": "backup"})
 )
 
+type CreateBackupParameters struct {
```
innobead, line 25: ref: https://stackoverflow.com/a/45886055 

I think the original naming should be fine. Or if you want it more precise, you can use `BackupCreateParameter`. 

nit: I feel we can do this pattern for the parameter for DoBackupRestore as well. This is not a priority. We can refactor over time.

innobead (approved):
LGTM.

pkg/backup/main.go
derekbit: `BackupCreateParameter` sounds good.

PR #821, opened 2023-01-09
fix: restore backup failed behind internet proxy

mantissahz:
proto: add geting backup information apis

Move `backupstore.InspectBackup` to `SyncAgentServer` then instance manager have to get backup information by a replicaclient.

Need to update longhorn-instance-manager after merged.

Regression test: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2865

longhorn/longhorn#5054


derekbit:
Can you elaborate more on this issue?

innobead:
> Can you elaborate more on this issue?

Exactly, we should know the real cause before fixing this.

mantissahz:
>>Can you elaborate more on this issue?
>Exactly, we should know the real cause before fixing this.

This issue is about that the S3 request (`backupstore.InspectBackup` to get backup information from the S3 backup target) failed to be sent to the S3 backup target through the internet proxy in the air-gap/private network environment.

Actually I did not find the exact root cause about why instance-manager can not send requests to S3 through internet proxy (HTTP_PROXY was set but ) but engine can work, I would do more testings on this to find out the real root cuase.

innobead:
> Actually I did not find the exact root cause about why instance-manager can not send requests to S3 through internet proxy (HTTP_PROXY was set but ) but engine can work, I would do more testings on this to find out the real root cause.

@shuo-wu have any idea?

mantissahz:
Some updates:

`backupstore.InspectBackup` would call `GetBackupStoreDriver` to get the http client from  [GetClient](https://github.com/longhorn/backupstore/blob/master/http/client.go#L31) when initializing s3 driver.
The restoring a backup would work well if setting the transport proxy `customTransport.Proxy = [correct proxy url]` after [ http.DefaultTransport.(*http.Transport).Clone()](https://github.com/longhorn/backupstore/blob/master/http/client.go#L41-L47) in `GetClient`.
That means `http.ProxyFromEnvironment` might be modified somewhere or there is some reason that makes it returns a invalid url or nothing

innobead:
> Some updates:
> 
> `backupstore.InspectBackup` would call `GetBackupStoreDriver` to get the http client from [GetClient](https://github.com/longhorn/backupstore/blob/master/http/client.go?rgh-link-date=2023-01-10T08%3A49%3A14Z#L31) when initializing s3 driver. The restoring a backup would work well if setting the transport proxy `customTransport.Proxy = [correct proxy url]` after [ http.DefaultTransport.(*http.Transport).Clone()](https://github.com/longhorn/backupstore/blob/master/http/client.go?rgh-link-date=2023-01-10T08%3A49%3A14Z#L41-L47) in `GetClient`. That means `http.ProxyFromEnvironment` might be modified somewhere or there is some reason that makes it returns a invalid url or nothing

Do you mean this?
https://github.com/longhorn/longhorn-manager/blob/60df31ab85f5f561cc012685b0f40ae7642cb29d/engineapi/backups.go#L242-L251

In ``, the credentials are passed as project environment variables already, why the inspect command doesn't respect them only?
https://github.com/longhorn/longhorn-manager/blob/60df31ab85f5f561cc012685b0f40ae7642cb29d/engineapi/backups.go#L109-L115

https://github.com/longhorn/longhorn-manager/blob/60df31ab85f5f561cc012685b0f40ae7642cb29d/engineapi/backups.go#L74-L107

cc @shuo-wu @c3y1huang

mantissahz:
>Do you mean this?
https://github.com/longhorn/longhorn-manager/blob/60df31ab85f5f561cc012685b0f40ae7642cb29d/engineapi/backups.go#L242-L251
>In ExecuteEngineBinary, the credentials are passed as process environment variables already, so why the inspect command doesn't respect them only but other commands work?
https://github.com/longhorn/longhorn-manager/blob/60df31ab85f5f561cc012685b0f40ae7642cb29d/engineapi/backups.go#L109-L115
> https://github.com/longhorn/longhorn-manager/blob/60df31ab85f5f561cc012685b0f40ae7642cb29d/engineapi/backups.go#L74-L107

No, it is not about `ExecuteEngineBinary`.
It failed when calling [backupstore.InspectBackup](https://github.com/longhorn/longhorn-engine/blob/master/pkg/sync/backup.go#L246) and it will call `GetBackupStoreDriver` to initialize and test the s3 client.

innobead:
> No, it is not about `ExecuteEngineBinary`. It failed when calling [backupstore.InspectBackup]

Yeah, it's failed in the proxy method `BackupRestore`. In the proxy `BackupRestore`, before calling the backup restore against each replica, we will call `backupstore.InspectBackup(backup)` which just uses an HTTP client behind to get the backup info, so need to pass the credentials to it or either set the PROXY env vars to the IM-E (caller), but the second option is not we want actually.

I feel credential management for backup is not well managed. Need to review and improve it. 

So I would suggest you just fix the command instead of moving the inspect function to the replica process (sync agent server).

innobead:
cc @PhanLe1010 I think no need to spend time reproducing because the root cause has been found and the context is clear now.

cc @longhorn/qa

PhanLe1010:
Yes, thanks @innobead

shuo-wu:
Based on your statement, it seems that [this credential config](https://github.com/longhorn/longhorn-instance-manager/blob/a2efd12a5bfed9ec43b162563cfcb23d9ca963d6/pkg/proxy/backup.go#L151-L165) does not work correctly, right?

mantissahz:
Actually [credential config](https://github.com/longhorn/longhorn-instance-manager/blob/a2efd12a5bfed9ec43b162563cfcb23d9ca963d6/pkg/proxy/backup.go#L151-L165) is set and  HTTP_PROXY and HTTPS_PROXY are correct.

mantissahz:
`http.ProxyFromEnvironment` will be fixed after grpc server is running.
Modifying environment variables `HTTP_PROXY` and `HTTPs_Proxy` will not be updated.
Let's give a method to `http.Transport.Proxy` instead of `http.ProxyFromEnvironment`.

https://github.com/longhorn/backupstore/pull/97

PR #827, opened 2023-01-11
CIFS Backup Store Support

derekbit:
https://github.com/longhorn/longhorn/issues/3599


derekbit:
Dependencies are updated.

pkg/util/util.go
```diff
@@ -265,33 +264,12 @@ func CheckBackupType(backupTarget string) (string, error) {
 }
 
 func GetBackupCredential(backupURL string) (map[string]string, error) {
```
innobead, line 266: Suggest moving `GetBackupCredential` to util/credential.go as well.

innobead (changes requested):
LGTM, just one minor suggestion.

pkg/util/util.go
derekbit: Addressed. Thank you.

innobead (approved):

shuo-wu (approved):

PR #824, opened 2023-01-10
vendor: update sparse-tools

derekbit:
Longhorn/longhorn#5130

Signed-off-by: Derek Su <derek.su@suse.com>

PR #828, opened 2023-01-12
[BACKPORT][v1.3.x] vendor: update sparse-tools (backport #824) 

derekbit:
Longhorn/longhorn#5130

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/gorilla/websocket v1.4.2 // indirect
 	github.com/longhorn/backupstore v0.0.0-20220913112826-5f5c95274f2a
 	github.com/longhorn/go-iscsi-helper v0.0.0-20220805034259-7b59e22574bb
-	github.com/longhorn/sparse-tools v0.0.0-20220323120706-0bd9b4129826
+	github.com/longhorn/sparse-tools v0.0.0-20230110055233-727da1d6548d
 	github.com/mattn/go-colorable v0.1.4 // indirect
 	github.com/mattn/go-runewidth v0.0.5-0.20181218000649-703b5e6b11ae // indirect
 	github.com/moby/moby v1.11.1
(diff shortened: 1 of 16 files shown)


innobead:
PR check failed?

derekbit:
> PR check failed?

Just noticed the some apis' parameters are changed. Sould we backport to v1.3.x?

innobead:
> > PR check failed?
> 
> Just noticed the some apis' parameters are changed. Sould we backport to v1.3.x?

You are right, it's about the new checksum introduction as below. 

Unfortunately, Golang does not have function overloading or function default parameters to cause this incompatible change. The only solution seems just to branch the sparse_tools library or next time, we should create a new method instead of changing the existing method, but this way is really old school like C and it will need the caller to do change accordingly.  

```
# github.com/longhorn/longhorn-engine/pkg/sync/rpc
pkg/sync/rpc/server.go:343:27: not enough arguments in call to sparse.SyncFile
	have (string, string, number, bool)
	want (string, string, int, bool, bool)
pkg/sync/rpc/server.go:374:30: not enough arguments in call to sparse.SyncContent
	have (string, *replica.Replica, int64, string, number, bool)
	want (string, sparse.ReaderWriterAt, int64, string, int, bool, bool)
```

derekbit:
> Unfortunately, Golang does not have function overloading or function default parameters to cause this incompatible change. The only solution seems just to branch the sparse_tools library or next time, we should create a new method instead of changing the existing method, but this way is really old school like C and it will need the caller to do change accordingly.

One idea is to make lib's api parameters like
```
func mount(souurce, mountpoint string, mountOptions MountOptions)

type MountOptions struct {
   vers string
   ro bool
   ...
}
```

`souurce` and `mountpoint` are always necessary.
The fields in `MountOptions` are optional for different versions.

innobead:
yeah, this is what I thought as well. 

Should better design the library to make it backward compatible enough, but based on Golang's capability, A new method or object/structure parameter is the way to adopt.

PR #838, opened 2023-02-09
build: use go 1.17

innobead:


innobead:
@Mergifyio backport v1.4.x

PR #849, opened 2023-02-15
vendor: update go-iscsi-helper

derekbit:
Longhorn/longhorn#5383


innobead (approved):
LGTM

innobead:
Just noticed go-iscsi-helper gets diverse based on different release branches like right now, we have master/1.3.x/1.2.x. The reason for that is because some feature-specific changes for different releases with coupling into this library, especially `LonghornDevice` like trim, expansion, etc.

However, ideally, it will be better to keep the library compatible w/o any branches to support all sustainable releases. For any changes based on interface, we should leverage the inheritance of OO to override the contract implementation instead of changing the underlining implementation. For example, for the newly added `SetLunThinProvisioning` in 1.4, we should have another device implementation like `LonghornDevice2` to override the `Start` implementation instead of changing `CreateTarget` method directly which caused the branch diverse due to new things introduced to this common library. 

image

PR #851, opened 2023-02-16
meta: update CLIAPIVersion to 8

derekbit:
Update CLIAPIVersion to 8 because of the newly introduced compression enhancement.

Longhorn/longhorn#5376


innobead:
CLI has been changed due to https://github.com/longhorn/longhorn-engine/commit/8e096311 for `backup create`.

innobead (approved):

PR #853, opened 2023-02-16
[BACKPORT][v1.4.x] vendor: update sparse-tools

derekbit:
Revert "sync file: buffering read disk files for replica rebuild"

After the rebuilding, the memory usage of the instance-manager-r increases. This is not harmful. However, to avoid uncertainties caused by the checks of pods' memory usages, we revert the commit.

Longhorn/longhorn#5246


innobead (dismissed):

innobead:
@derekbit the title is wrong. It should be for sparse-tools update.

derekbit:
Fixed. Thank you.

innobead (approved):

PR #854, opened 2023-02-16
Clean up unused mount points

derekbit:
Longhorn/longhorn#5391

deps on https://github.com/longhorn/backupstore/pull/102


derekbit:
Updated in https://github.com/longhorn/longhorn-engine/pull/855 instead.

PR #863, opened 2023-03-06
chore(vendor): update backupstore

derekbit:


innobead (approved):

innobead (changes requested):
Should update the go to 1.17 as well. (ideally, we should not update the base go version for patch release, but it seems we missed updating it when releasing 1.4.0, so it should be fine this time)

innobead (approved):

PR #868, opened 2023-03-15
Separate service unavailable and replica error

derekbit:
When an engine had no available backend, the controller emitted a message stating "...no available backend...". Afterward, the volume controller blindly marked all replicas as failed.

To create a RWX volume, it was first created on node A, and then the sharemanager resource was created with node A as the owner. The sharemanager controller then created a sharemanager pod, which might be located on node B. After the pod was created, the sharemanager resource's owner was transferred to node B, causing the volume to detach from node A and attach to node B.

However, if replicas were deleted because of detachment, the in-progress engine instance's startup would fail due to the "...no available backend..." error. This caused all replicas to be marked as failed (healthyAt is empty and failedAt is set), and they could not be salvaged anymore. As a result, the volume became faulty.

To avoid the blindly marking failed operation, we should separate the recoverable and unrecoverable error. For a recoverable error, for example, the replica instance is deleted, the replica is not really failed, the replica should not be marked as failed directly.

To prevent blindly marking replicas as failed, recoverable and unrecoverable errors can be separated. The manifest corruption can be regarded as a unrecoverable error. For a recoverable error, for instance, if a replica instance is deleted and the controller is disconnected with it, it should not be marked as failed directly since the replica are somehow stopped rather than corrupted. In this case, avoiding marking all replicas failed can help the volume not trapped in a faulty state and reattach to a node.

Longhorn/longhorn#5537

### DIFF
--- a/pkg/controller/control.go
+++ b/pkg/controller/control.go
@@ -4,12 +4,14 @@ import (
 	"fmt"
 	"net"
 	"regexp"
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 
 	iutil "github.com/longhorn/go-iscsi-helper/util"
 
@@ -694,6 +696,15 @@ func determineCorrectVolumeSize(volumeSize, volumeCurrentSize int64, backendSize
 	return volumeCurrentSize
 }
 
+func isBackendServiceUnavailable(errorCodes map[string]codes.Code) bool {
+	for _, code := range errorCodes {
+		if code == codes.Unavailable {
+			return true
+		}
+	}
+	return false
+}
+
 func (c *Controller) Start(volumeSize, volumeCurrentSize int64, addresses ...string) error {
 	c.Lock()
 	defer c.Unlock()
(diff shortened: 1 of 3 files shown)


derekbit:
> But what if the replica is running for just several seconds then get failed? For example, the work directory is read-only, I haven't tested this example if this is possible but we cannot guarantee the similar issues won't happen...

If the replica server can error out and does not die accidentally, the error is regarded as unrecoverable. Then, the replicas can still be marked as failed.

innobead (commented):
just for reference to see when the controller determine replicas failed. 

[/controller/volume_controller.go#L550-L566](https://github.com/longhorn/longhorn-manager/blob/a3c94b56f11fd744998dc0191466ce98775e1918/controller/volume_controller.go#L550-L566)
```go
	if e.Status.CurrentState != longhorn.InstanceStateRunning {
		// If a replica failed at attaching stage before engine become running,
		// there is no record in e.Status.ReplicaModeMap
		engineInstanceCreationCondition := types.GetCondition(e.Status.Conditions, longhorn.InstanceConditionTypeInstanceCreation)
		isNoAvailableBackend := strings.Contains(engineInstanceCreationCondition.Message, fmt.Sprintf("exit status %v", int(syscall.ENODATA)))
		for _, r := range rs {
			if isNoAvailableBackend || (r.Spec.FailedAt == "" && r.Status.CurrentState == longhorn.InstanceStateError) {
				log.Warnf("Replica %v that not in the engine mode map is marked as failed, current state %v, engine name %v, active %v",
					r.Name, r.Status.CurrentState, r.Spec.EngineName, r.Spec.Active)
				e.Spec.LogRequested = true
				r.Spec.LogRequested = true
				r.Spec.FailedAt = vc.nowHandler()
				r.Spec.DesireState = longhorn.InstanceStateStopped
			}
		}
		return nil
	}
```

pkg/controller/control.go
```diff
@@ -694,6 +696,15 @@ func determineCorrectVolumeSize(volumeSize, volumeCurrentSize int64, backendSize
...
 }
 
+func isBackendServiceUnavailable(errorCodes map[string]codes.Code) bool {
+	for _, code := range errorCodes {
+		if code == codes.Unavailable {
+			return true
+		}
+	}
+	return false
+}
```
innobead, line 706: For this implementation, we will not regard all replicas failed if there is at least one is unable to reach. It seems not granular enough due to the current design limitation, but ideally we should be able to determine each replica individually. 

My question is shouldn't we check all error codes of every replica instead of only one?

innobead:
GRPC UNAVAILABLE definition.

> The service is currently unavailable. This is most likely a transient condition, which can be corrected by retrying with a backoff. Note that it is not always safe to retry non-idempotent operations.

innobead:
> f the replica server can error out and does not die accidentally, the error is regarded as unrecoverable. Then, the replicas can still be marked as failed.

Yes, this is just a current error handling.

pkg/controller/control.go
innobead: > My question is shouldn't we check all error codes of every replica instead of only one?

I think it should be fine because individual replica should be able to sync their status from the instance manager. 

image

innobead (approved):

pkg/controller/control.go
innobead: @derekbit Can you confirm this?

derekbit: If one or more than one of the replicas failed caused by service unavailable, the all replicas won't be marked as error.
The volume controller will retry the attachment. Once it succeeds, the engine monitor will set the real failed replicas to ERR state, and later the volume controller marked them as failed.

innobead: Is it possible to become endless retry?

derekbit: Do you mean if the replicas' services are always unavailable? If yes, it will be at attach/detach loop.

innobead: Yeah but it seems fair because it's just the result due to unexpected environment situation.

derekbit: Yeah, I think the behavior looks reasonable.

innobead:
@Mergifyio backport v1.3.x v1.4.x

PR #870, opened 2023-03-16
Fix deadlock when encountering a failed UnmapAt operation

derekbit:
Longhorn/longhorn#5575

### DIFF
--- a/pkg/controller/control.go
+++ b/pkg/controller/control.go
@@ -906,21 +906,26 @@ func (c *Controller) UnmapAt(length uint32, off int64) (int, error) {
 	// TODO: Need to fail unmap requests
 	//  if the volume is purging snapshots or creating backups.
 	c.RLock()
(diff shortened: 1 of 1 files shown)


innobead (commented):
I didn't see the change here different than before, since the original defer will be run eventually?

innobead (approved):
Wait, it's happening in `return n, c.handleError(err)`. 

[/pkg/controller/control.go#L975-L979](https://github.com/longhorn/longhorn-engine/blob/8b4cf86f3392afd88802f0d477116fadfd2d48ce/pkg/controller/control.go#L975-L979)
```go
func (c *Controller) handleError(err error) error {
	c.Lock()
	defer c.Unlock()
	return c.handleErrorNoLock(err)
}
```

So bad :disappointed: 

LGTM.

innobead:
@Mergifyio backport v1.4.x

PR #875, opened 2023-03-22
vendor: update backupstore

derekbit:
Depends on https://github.com/longhorn/backupstore/pull/112
Longhorn/longhorn#5628


innobead:
@derekbit this will change nfs mount to use soft mode, but we didn't tackle https://github.com/longhorn/longhorn/issues/5662 yet. Probably we should draft this backport first until 5662 has been tackled?

derekbit:
> @derekbit this will change nfs mount to use soft mode, but we didn't tackle [longhorn/longhorn#5662](https://github.com/longhorn/longhorn/issues/5662) yet. Probably we should draft this backport first until 5662 has been tackled?
> 
> Wait this is the update to longhorn-engine instead, so it should be fine.

https://github.com/longhorn/longhorn/issues/5662 depends on https://github.com/longhorn/longhorn/issues/5628, so need to update this first.

innobead (approved):

PR #877, opened 2023-03-24
chore: the io/ioutil package is deprecated from Go 1.16

jenting:


innobead (approved):
LGTM. The min required version is Go 1.17, so we will only backport to version 1.4.

innobead:
@Mergifyio backport v1.4.x

PR #879, opened 2023-03-26
[BACKPORT][v1.3.3] vendor: update backupstore

derekbit:
https://github.com/longhorn/longhorn/issues/5654


derekbit:
Won't apply to v1.3.x because the feature will change the backup behavior.

PR #882, opened 2023-03-28
Improve the error handling of the backup failure

derekbit:
The purpose of this pull request is to enhance the error handling mechanism in case of backup failures. This improvement is specifically targeted at backups that use an NFS backup store that is mounted in `soft` mode. In this mode, operations such as "mount" and "write" are not retried infinitely.

Depends on https://github.com/longhorn/backupstore/pull/117

Longhorn/longhorn#5662


pkg/replica/backup.go
```diff
@@ -154,24 +158,33 @@ type BackupStatus struct {
...
+func (rb *BackupStatus) UpdateBackupStatus(snapID, volumeID string, state string, progress int, url string, errString string) error {
 	id := diskutil.GenerateSnapshotDiskName(snapID)
 	rb.lock.Lock()
 	defer rb.lock.Unlock()
-	if err := rb.assertOpen(id, volumeID); err != nil {
-		logrus.Errorf("Returning Error from UpdateBackupProgress")
-		return err
+
+	if rb.volumeID != volumeID || rb.SnapshotID != id {
+		return fmt.Errorf("invalid volume [%s] and snapshot [%s], not volume [%s], snapshot [%s]", rb.volumeID, rb.SnapshotID, volumeID, id)
```
shuo-wu, line 184: NIT: Abstract this as a function, e.g., `assertVolumeSnapshot`

shuo-wu (commented):
>  However, this meant that any stuck caused by an NFS
  connection during backupstore initialization could not be detected by the Longhorn manager

Can you explain more about this statement?

pkg/replica/backup.go
derekbit: Addressed. Thank you.

pkg/sync/rpc/server.go
```diff
@@ -670,7 +670,7 @@ func (s *SyncAgentServer) BackupCreate(ctx context.Context, req *ptypes.BackupCr
 		return nil, err
 	}
 
-	backupID, replicaObj, err := backup.DoBackupCreate(&backup.CreateBackupParameters{
+	replicaObj, backupConfig, err := backup.DoBackupInit(&backup.CreateBackupParameters{
```
shuo-wu, line 673: NIT: Rename `replicaObj`. e.g., to `backupStatus` or `backupInfo`

shuo-wu (commented):
Great explanation! Can you add more descriptions in the comments?
The core is that, `backupstore.CreateDeltaBlockBackup(replicaBackup.Name, config)` contains NFS soft mount, which may fail forever now. We should make sure the sync agent server recorded the backup status before executing this function, right?

pkg/replica/backup.go
```diff
@@ -154,24 +158,33 @@ type BackupStatus struct {
...
+	if backupName == "" {
+		backupName = butil.GenerateName("backup")
+	}
 	return &BackupStatus{
+		Name:        backupName,
 		backingFile: backingFile,
-		State:       ProgressStateInProgress,
+		State:       ProgressStateStarting,
+		volumeID:    volumeID,
+		SnapshotID:  diskutil.GenerateSnapshotDiskName(snapID),
```
innobead, line 174: Why not just use the original `snapID`?

pkg/sync/rpc/server.go
```diff
@@ -682,17 +682,23 @@ func (s *SyncAgentServer) BackupCreate(ctx context.Context, req *ptypes.BackupCr
 		Labels:               req.Labels,
 	})
 	if err != nil {
-		logrus.WithError(err).Errorf("Failed to create backup %v", req.BackupName)
+		logrus.WithError(err).Errorf("Error initializing backup %v", req.BackupName)
```
innobead, line 685: Suggest log `Failed to initialize backup %v`

pkg/sync/rpc/server.go
```diff
@@ -682,17 +682,23 @@ func (s *SyncAgentServer) BackupCreate(ctx context.Context, req *ptypes.BackupCr
...
-	resp := &ptypes.BackupCreateResponse{
-		Backup:        backupID,
-		IsIncremental: replicaObj.IsIncremental,
+	if err := s.BackupList.BackupAdd(replicaObj.Name, replicaObj); err != nil {
+		return nil, errors.Wrapf(err, "failed to add the backup object %v", replicaObj.Name)
+	}
+
+	err = backup.DoBackupCreate(replicaObj, backupConfig)
+	if err != nil {
+		logrus.WithError(err).Errorf("Error creating backup %v", req.BackupName)
```
innobead, line 695: `Failed to create backup..`

pkg/sync/rpc/server.go
```diff
@@ -725,7 +725,7 @@ func (s *SyncAgentServer) BackupStatus(ctx context.Context, req *ptypes.BackupSt
 }
 
 func (*SyncAgentServer) BackupRemove(ctx context.Context, req *ptypes.BackupRemoveRequest) (*empty.Empty, error) {
-	cmd := reexec.Command("sbackup", "delete", req.Backup)
+	cmd := reexec.Command("backup", "delete", req.Backup)
```
innobead, line 734: Is there any place using `BackupRemove`? I thought only longhorn-manager will do backup deletion but why sync agent in replica has this capability as well?

pkg/replica/backup.go
```diff
@@ -154,24 +158,33 @@ type BackupStatus struct {
...
 
-func (rb *BackupStatus) UpdateBackupStatus(snapID, volumeID string, progress int, url string, errString string) error {
+func (rb *BackupStatus) UpdateBackupStatus(snapID, volumeID string, state string, progress int, url string, errString string) error {
 	id := diskutil.GenerateSnapshotDiskName(snapID)
 	rb.lock.Lock()
 	defer rb.lock.Unlock()
-	if err := rb.assertOpen(id, volumeID); err != nil {
-		logrus.Errorf("Returning Error from UpdateBackupProgress")
+
+	if err := rb.assertVolumeSnapshot(id, volumeID); err != nil {
```
innobead, line 183: nit: It is good to make imperative check in a meaningful method, but `assert` is usually used for testing purpose instead to assert something true.

innobead (commented):
In general, LGTM, but just some questions/comments.

Also, need to update the dep if the backupstore PR is ready.

pkg/sync/rpc/server.go
derekbit: No, `BackupRemove` is not used. longhorn-manager calls `backup rm...` rather than `BackupRemove` to delete a backup.

derekbit:
> assertVolumeSnapshot

Yeah, right!

pkg/sync/rpc/server.go
derekbit: Updated. Thank you.

pkg/replica/backup.go
derekbit: Updated. Thank you.

pkg/sync/rpc/server.go
derekbit: Updated. Thank you.

pkg/replica/backup.go
derekbit: `GenerateSnapshotDiskName` adds the prefix `volume-snap-` to snapID.

pkg/sync/rpc/server.go
derekbit: Updated. Thank you.

derekbit:
> In general, LGTM, but just some questions/comments.
> 
> Also, need to update the dep if the backupstore PR is ready.

Waiting for the review of https://github.com/longhorn/backupstore/pull/117.
cc @innobead

pkg/replica/backup.go
```diff
@@ -154,24 +158,33 @@ type BackupStatus struct {
...
 		backingFile: backingFile,
-		State:       ProgressStateInProgress,
+		State:       ProgressStateStarting,
+		volumeID:    volumeID,
+		SnapshotID:  diskutil.GenerateSnapshotDiskName(snapID),
 	}
 }
 
-func (rb *BackupStatus) UpdateBackupStatus(snapID, volumeID string, progress int, url string, errString string) error {
+func (rb *BackupStatus) UpdateBackupStatus(snapID, volumeID string, state string, progress int, url string, errString string) error {
```
shuo-wu, line 178: This function looks weird now. It receives an input state but sometimes it determines the state by itself...

pkg/replica/backup.go
```diff
@@ -19,6 +20,7 @@ type ProgressState string
 const (
 	snapBlockSize = 2 << 20 // 2MiB
 
+	ProgressStateStarting   = ProgressState("starting")
```
shuo-wu, line 23: NIT: Use the const defined in `backupstore` instead.

pkg/sync/rpc/server.go
innobead: ref: [/controller/backup_controller.go#L310-L314](https://github.com/longhorn/longhorn-manager/blob/3b1d6252cd8389e8465ee4b38cf93d6543f8c372/controller/backup_controller.go#L310-L314)
```go
			backupURL := backupstore.EncodeBackupURL(backup.Name, backupVolumeName, backupTargetClient.URL)
			if err := backupTargetClient.BackupDelete(backupURL, backupTargetClient.Credential); err != nil {
				log.WithError(err).Error("Error deleting remote backup")
				return err
			}
```

innobead: Keeping in sync server seems being confused somehow if it's not used. However, this is an exported interface, so let's just keep it as is.

pkg/backup/main.go
```diff
@@ -69,70 +69,72 @@ func RequiredMissingError(name string) error {
...
-		DeltaOps:        replicaBackup,
+		DestURL:         params.DestURL,
+		DeltaOps:        backup,
 		Labels:          labelMap,
-	})
+	}
+	return backup, config, nil
+}
+
+func DoBackupCreate(replicaBackup *replica.BackupStatus, config *backupstore.DeltaBackupConfig) error {
```
innobead, line 130: nit: Ideally, for each exported function/method in any level (component or lib), I would recommend having logging at the beginning. For example, `Staring  backup creation...`.

note: Since we use go, so it's not possible to add this kind of logs automatically by using meta programming or AOP at runtime or compile time.

pkg/replica/backup.go
innobead: It seems you already added states into backupstore, so let's take this time to fix this. WDYT?

pkg/replica/backup.go
```diff
@@ -154,24 +158,33 @@ type BackupStatus struct {
 	BackupURL     string
 	State         ProgressState
 	IsIncremental bool
+	IsOpened      bool
 }
 
-func NewBackup(backingFile *backingfile.BackingFile) *BackupStatus {
+func NewBackup(name, volumeID, snapID string, backingFile *backingfile.BackingFile) *BackupStatus {
```
innobead, line 164: The returned type seems inconsistent with the function. `NewXX` is more used for initializing an instance with type `XX`. In this case, `NewBackup` is more suitable for creating a backup instance instead of backup status. 

Because this is the current code, let's keep it as is for now.

pkg/replica/backup.go
innobead: Indeed, so I would suggest at least having a comment to explain this progress first-respected behavior.

innobead (changes requested):
In general, LGTM. 

Some comments needed to resolve.

pkg/replica/backup.go
derekbit: Add a comment as @innobead's suggestion.

pkg/replica/backup.go
derekbit: Reuse the state in `backupstore`.
But some codes in `pkg/replica` are using `ProgressStateXXX`, so I still keep them and move to `pkg/replica/replica.go`.

pkg/backup/main.go
```diff
@@ -69,70 +69,72 @@ func RequiredMissingError(name string) error {
 	return fmt.Errorf("cannot find valid required parameter: %v", name)
 }
 
-func DoBackupCreate(backupParams *CreateBackupParameters) (string, *replica.BackupStatus, error) {
+func DoBackupInit(params *CreateBackupParameters) (*replica.BackupStatus, *backupstore.DeltaBackupConfig, error) {
 	var err error
 
-	if backupParams.VolumeName == "" || backupParams.SnapshotName == "" || backupParams.DestURL == "" {
-		return "", nil, fmt.Errorf("missing input parameter")
+	if params.VolumeName == "" || params.SnapshotName == "" || params.DestURL == "" {
```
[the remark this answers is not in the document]
derekbit: Updated. Thank you.

pkg/backup/main.go
derekbit: Added. Thank you.

pkg/replica/backup.go
innobead: I feel some state constants are quite messy in the current code. NP, let's keep it for now.

pkg/backup/main.go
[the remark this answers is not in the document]
derekbit: Just found `params.BackupName` can be empty. If it is empty, `backupName` is generated automatically.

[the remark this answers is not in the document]
innobead: Then, why do we need to provide backup name if it can be generated automatically?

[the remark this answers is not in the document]
derekbit: Looks for the unit tests.
Backup name is not empty if the request is from longhorn-manager

derekbit:
cc @innobead

innobead (approved):

PR #902, opened 2023-05-11
Engine identity validation

ejweber:
longhorn/longhorn#5845

This PR provides the base mechanisms to allow for engine identity verification during normal Longhorn operation (assuming the higher level components like longhorn-instance-manager and longhorn-manager make use of them).


ejweber:
This PR is ready for review.

I [ran the Longhorn manager core tests on it](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4328/) with the expectation that all tests would pass (all changes should be "opt in"). There were four failures related to the fact that there is no test image like `longhornio/longhorn-test:upgrade-test...` with the new API version these changes specify:

- tests.test_backing_image.test_engine_live_upgrade_with_backing_image
- tests.test_engine_upgrade.test_engine_offline_upgrade
- tests.test_engine_upgrade.test_engine_live_upgrade
- tests.test_engine_upgrade.test_engine_live_upgrade_with_intensive_data_writing

It looks like code to generate a new image lives [here](https://github.com/longhorn/longhorn-tests/blob/6538ac9c7ca207b4f39dcddd982a8a19b31b1166/build_engine_test_images/scripts/generate_images.sh#L1-L75), but it is a GitHub bot that normally does the push. Is this a process I should kick off somehow?

PhanLe1010:
It looks like the test-engine image is built by [this Terraform script(https://github.com/longhorn/longhorn-tests/blob/6538ac9c7ca207b4f39dcddd982a8a19b31b1166/build_engine_test_images/terraform/aws/ubuntu/main.tf#L211-L227). Do you know how to trigger that script @yangchiu @chriscchien

yangchiu:
I triggered the pipeline from https://ci.longhorn.io/job/private/job/build-engine-test-images/, but it said `latest engine test images have already published` and didn't do anything. 

@chriscchien Can you help to elaborate more on how to set this right? Thanks!

app/cmd/replica.go
```diff
@@ -115,7 +116,14 @@ func startReplica(c *cli.Context) error {
 		}
 	}
 
-	volumeName := c.String("volume-name")
+	// The --volume-name flag for the replica command (added before the identity validation feature) and the
+	// --volume-name global flag are redundant. Since both are optional, proceed with either one, but do not proceed if
+	// they are different.
+	volumeName, err := resolveVolumeName(c.String("volume-name"), c.GlobalString("volume-name"))
```
shuo-wu, line 122: Can you ignore flag `--volume-name`  in command `replica` and rely on the global option only?
Same below. Then there is no need to add `resolveVolumeName()`

proto/ptypes/interceptor.go
```diff
@@ -0,0 +1,74 @@
...
+			// Only refuse to serve if both client and server provide validation information.
+			if ok && volumeName != "" {
+				log := logrus.WithFields(logrus.Fields{"method": info.FullMethod,
+					"clientVolumeName": incomingVolumeName[0], "serverVolumeName": volumeName})
+				if incomingVolumeName[0] != volumeName {
+					log.Error("Invalid gRPC metadata")
+					return nil, status.Errorf(codes.FailedPrecondition, "Incorrect volume name; check %s address",
+						serverType)
+				}
+				log.Debug("Valid gRPC metadata")
```
shuo-wu, line 36: NIT: I think `log.Trace` is enough. Same below

proto/ptypes/interceptor.go
```diff
@@ -0,0 +1,74 @@
...
+		md, ok := metadata.FromIncomingContext(ctx)
+		if ok {
+			incomingVolumeName, ok := md["volume-name"]
+			// Only refuse to serve if both client and server provide validation information.
+			if ok && volumeName != "" {
+				log := logrus.WithFields(logrus.Fields{"method": info.FullMethod,
+					"clientVolumeName": incomingVolumeName[0], "serverVolumeName": volumeName})
+				if incomingVolumeName[0] != volumeName {
+					log.Error("Invalid gRPC metadata")
+					return nil, status.Errorf(codes.FailedPrecondition, "Incorrect volume name; check %s address",
```
shuo-wu, line 33: NIT: We can add `incomingVolumeName[0]` and `volumeName` to the error info. Same below.

shuo-wu (approved):
In general LGTM

app/cmd/replica.go
ejweber: I was hesitant to do this for backwards compatibility reasons, though I will admit I'm not sure I have a perfect handle on exactly what those reasons are.

Older versions of longhorn-manager will want to pass the replica-command-specific --volume-name flag. New versions of longhorn-manager will be aware of the global --volume-name flag. The resolveVolumeName function ensures that we can correctly respond to both.

1. Is there any scenario in which an older longhorn-manager will talk to an updated longhorn-engine? (I think the answer is no, since longhorn-manager always upgrades first and then longhorn-engine is upgraded asynchronously, but this was my concern.)
2. Should I remove the replica-command-specific --volume-name flag entirely?

I have not made any changes in response to this comment yet.

ejweber:
> I triggered the pipeline from https://ci.longhorn.io/job/private/job/build-engine-test-images/, but it said latest engine test images have already published and didn't do anything.
>
> @chriscchien Can you help to elaborate more on how to set this right? Thanks!

I looks like that automation accept only works against the https://github.com/longhorn/longhorn-engine repo and accepts an optional commit ID (defaults to tip of master). Since these changes aren't yet in that repo, it seems we cannot run the automation until after merge.

app/cmd/replica.go
shuo-wu: I don't think we need to consider older version longhorn manager pods. Unlike the instance manager or engine image, all old Longhron manager pods will be evicted (daemonset rolling upgrade percentage is 100%) before launching the first new longhorn manager pod. And the new instance manager launching & volume upgrade happen after the new longhorn manager ready.
As long as the new longhorn managers are aware of using different flags (global flag for new engines, cmd-specific flag for old engine) for different versions of engines, it should work.

ejweber: Makes sense. I removed the function and ignored the command-specific flag for my new use case. Once I have a longhorn-manager PR merged that only uses the global flag for the new engine, I will come back and remove the command-specific flag entirely (per the code comments).

app/cmd/replica.go
```diff
@@ -116,18 +118,14 @@ func startReplica(c *cli.Context) error {
...
-	}
+	volumeName := c.GlobalString("volume-name")
 	replicaInstanceName := c.String("replica-instance-name")
 	dataServerProtocol := c.String("data-server-protocol")
 
 	controlAddress, dataAddress, syncAddress, syncPort, err :=
-		util.GetAddresses(volumeName, address, types.DataServerProtocol(dataServerProtocol))
+		// TODO (5845): Remove replica command-specific volume-name flag when its use is completely replaced in
+		// longhorn-manager.
+		util.GetAddresses(c.String("volume-name"), address, types.DataServerProtocol(dataServerProtocol))
```
shuo-wu, line 128: Why will you use `c.String("volume-name")` rather than variable `volumeName` here?

chriscchien:
> > I triggered the pipeline from https://ci.longhorn.io/job/private/job/build-engine-test-images/, but it said latest engine test images have already published and didn't do anything.
> > @chriscchien Can you help to elaborate more on how to set this right? Thanks!
> 
> I looks like that automation accept only works against the https://github.com/longhorn/longhorn-engine repo and accepts an optional commit ID (defaults to tip of master). Since these changes aren't yet in that repo, it seems we cannot run the automation until after merge.

Yes the engine upgrade test images all locate in https://hub.docker.com/r/longhornio/longhorn-test/ and built by [pipeline](https://ci.longhorn.io/job/private/job/build-engine-test-images/) based on latest longhorn-engine information
```
$ sudo docker run longhornio/longhorn-engine:master-head longhorn version --client-only
{
	"clientVersion": {
		"version": "d1d6d308",
		"gitCommit": "d1d6d308d37a5b465bc1c2824da2e0702830de7d",
		"buildDate": "2023-06-27T15:51:32+00:00",
		"cliAPIVersion": 8,
		"cliAPIMinVersion": 3,
		"controllerAPIVersion": 4,
		"controllerAPIMinVersion": 3,
		"dataFormatVersion": 1,
		"dataFormatMinVersion": 1
	},
	"serverVersion": null
}
```
With those `APIVserions`, pipeline will generate below longhornio/longhorn-test test images for e2e cases use by this [script](https://github.com/longhorn/longhorn-tests/blob/master/build_engine_test_images/scripts/generate_images.sh) :
- upgrade-test.8-3.4-3.1-1
- version-test.7-2.4-3.1-1
- version-test.9-4.4-3.1-1
- version-test.2-2.4-3.1-1
- version-test.8-3.4-3.1-1
- version-test.9-9.4-3.1-1
 
However, in test script the `APIVserions` come from custom engine and if the custom engine has different `APIVserions` with master-head,  related engine upgrade test case can not pass because test images not exist.

In conclusion, if engine `APIVserions` changed in PR, currently the engine upgrade cases only can be tested after merged, thank you.

app/cmd/replica.go
ejweber: My intention is for this to be temporary. Until longhorn-manager changes eventually merge, longhorn-manager will not be aware of the global flag on master-head and will continue to provide the replica command specific volume-name flag. This might cause annoying test failures, etc. I am thinking we can merge this longhorn-engine PR, allowing me to merge the longhorn-instance-manager and longhorn-manager PRs. Then I can come back and completely remove the command-specific flag per the code comments.

ejweber:
I have never observed it on my local machine, but the new integration tests executed by drone have failed a couple of times because the engine and/or replica server are not yet listening on the expected socket. I added a commit that takes advantage of existing functions with built-in retries.

PhanLe1010:
I drew this graph to assist myself while reviewing this PR. I thought it might be useful for other team members as well. Please correct me if you see anything wrong.

longhorn_grpc_communication

From the graph, there are some points that I notice for this PR:
1. I think [the SyncAgentGRPCServer](https://github.com/ejweber/longhorn-engine/blob/f4414d2e32cd06e94bffdcc798172b56477604e0/app/cmd/sync_agent.go#L90) should also have this identity verification similar as the ControllerGRPCServer and the ReplicaGRPCServer because the sync agent also doing tasks related to volume's snapshots and data. It would be dangerous for the syncAgent to perform the operation on the wrong data. 
2. SyncAgentServer should also include volume name and instance name in GRPC metadata when it talk to ReplicaServer because we don't want to talk to a wrong ReplicaServer. For example, operations like [this](https://github.com/ejweber/longhorn-engine/blob/f4414d2e32cd06e94bffdcc798172b56477604e0/pkg/sync/rpc/server.go#L955), [this](https://github.com/ejweber/longhorn-engine/blob/f4414d2e32cd06e94bffdcc798172b56477604e0/pkg/sync/rpc/server.go#L973), [this](https://github.com/ejweber/longhorn-engine/blob/f4414d2e32cd06e94bffdcc798172b56477604e0/pkg/sync/rpc/server.go#L1270), [this 
 ](https://github.com/ejweber/longhorn-engine/blob/f4414d2e32cd06e94bffdcc798172b56477604e0/pkg/sync/rpc/server.go#L1270-L1271), [this](https://github.com/ejweber/longhorn-engine/blob/f4414d2e32cd06e94bffdcc798172b56477604e0/pkg/sync/rpc/server.go#L1340), and [this](https://github.com/ejweber/longhorn-engine/blob/f4414d2e32cd06e94bffdcc798172b56477604e0/pkg/sync/rpc/server.go#L1361)

PhanLe1010 (commented):
I drew this graph to assist myself while reviewing this PR. I thought it might be useful to share with the team as well. Please correct me if you see anything wrong.

longhorn_grpc_communication

From the graph, there are some points that I notice for this PR:
1. I think [the SyncAgentGRPCServer](https://github.com/ejweber/longhorn-engine/blob/f4414d2e32cd06e94bffdcc798172b56477604e0/app/cmd/sync_agent.go#L90) should also have this identity verification similar as the ControllerGRPCServer and the ReplicaGRPCServer because the sync agent also doing tasks related to volume's snapshots and data. It would be dangerous for the syncAgent to perform the operation on the wrong data. 
2. SyncAgentServer should also include volume name and instance name in GRPC metadata when it talk to ReplicaServer because we don't want to talk to a wrong ReplicaServer. For example, operations like [this](https://github.com/ejweber/longhorn-engine/blob/f4414d2e32cd06e94bffdcc798172b56477604e0/pkg/sync/rpc/server.go#L955), [this](https://github.com/ejweber/longhorn-engine/blob/f4414d2e32cd06e94bffdcc798172b56477604e0/pkg/sync/rpc/server.go#L973), [this](https://github.com/ejweber/longhorn-engine/blob/f4414d2e32cd06e94bffdcc798172b56477604e0/pkg/sync/rpc/server.go#L1270), [this 
 ](https://github.com/ejweber/longhorn-engine/blob/f4414d2e32cd06e94bffdcc798172b56477604e0/pkg/sync/rpc/server.go#L1270-L1271), [this](https://github.com/ejweber/longhorn-engine/blob/f4414d2e32cd06e94bffdcc798172b56477604e0/pkg/sync/rpc/server.go#L1340), and [this](https://github.com/ejweber/longhorn-engine/blob/f4414d2e32cd06e94bffdcc798172b56477604e0/pkg/sync/rpc/server.go#L1361)

app/cmd/replica.go
shuo-wu: For this kind of change that involves multiple repos and compatibility issues, we will merge all PRs at the same time. There is no need to worry about temporary mismatchings between the engine PR merge and the manager PR merge. So please directly use the global flags in this PR.

ejweber: I currently import the longhorn-engine commits on my own branch in my longhorn-instance-manager and longhorn-manager PRs. This allows me to test those changes, since the commits aren't in the longhorn/longhorn-engine repo yet.  My intention was to change each to reference the correct. My intention was to fix these imports after the longhorn-engine changes merged and before the longhorn-instance-manager and longhorn-manager changes were approved. Please let me know what the best practice here is.

PhanLe1010: Agree with @shuo-wu about the using the global flags in this PR directly. 

In general, since `longhorn-engine` is imported by `longhorn-manager` and `longhorn-instance-manager`, I think we can use this flow:
1. Update the `longhorn-engine` PR repo as if  `longhorn-manager` and `longhorn-instance-manager` are already updated
2. In  `longhorn-manager` and `longhorn-instance-manager` PRs, use the replace directive in `go.mod` to point to the in-progress engine PR
3. When all `longhorn-engine`, `longhorn-manager`, `longhorn-instance-manager` are approved and ready for merging. 
    1. `longhorn-engine` PR will be merged first 
    2. In  `longhorn-manager` and `longhorn-instance-manager` PRs, edit `go.mod` to point to the correct engine commits 
    3. Merger  `longhorn-manager` and `longhorn-instance-manager` PRs
    4. This PRs merging process should happen without a few hours so there is no need to worry about temporary missmatchings

ejweber:
> 1. I think [the SyncAgentGRPCServer](https://github.com/ejweber/longhorn-engine/blob/f4414d2e32cd06e94bffdcc798172b56477604e0/app/cmd/sync_agent.go#L90) should also have this identity verification similar as the ControllerGRPCServer and the ReplicaGRPCServer because the sync agent also doing tasks related to volume's snapshots and data. It would be dangerous for the syncAgent to perform the operation on the wrong data.

Good catch. We already initialize the SyncAgentServiceClient with the client side interceptor [here](https://github.com/longhorn/longhorn-engine/blob/394e5918cf83e9726547e77c4cbb08b6bff627e5/pkg/replica/client/client.go#L113-L118), but I missed the server side interceptor. This is coming in the next commit.

> SyncAgentServer should also include volume name and instance name in GRPC metadata when it talk to ReplicaServer because we don't want to talk to a wrong ReplicaServer. For example, operations like [this](https://github.com/ejweber/longhorn-engine/blob/f4414d2e32cd06e94bffdcc798172b56477604e0/pkg/sync/rpc/server.go#L955), [this](https://github.com/ejweber/longhorn-engine/blob/f4414d2e32cd06e94bffdcc798172b56477604e0/pkg/sync/rpc/server.go#L973), [this](https://github.com/ejweber/longhorn-engine/blob/f4414d2e32cd06e94bffdcc798172b56477604e0/pkg/sync/rpc/server.go#L1270), [this](https://github.com/ejweber/longhorn-engine/blob/f4414d2e32cd06e94bffdcc798172b56477604e0/pkg/sync/rpc/server.go#L1270-L1271)
, [this](https://github.com/ejweber/longhorn-engine/blob/f4414d2e32cd06e94bffdcc798172b56477604e0/pkg/sync/rpc/server.go#L1340), and [this](https://github.com/ejweber/longhorn-engine/blob/f4414d2e32cd06e94bffdcc798172b56477604e0/pkg/sync/rpc/server.go#L1361)

Another good catch. I missed these dials. Fixing in the coming commit.

I made these changes but am adding integration tests to confirm them. I didn't get this done before wrapping up today.

- [ ] Add integration tests that use sync agent server/client.

integration/rpc/sync/sync_agent_client.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def replica_rebuild_status(self):
+        return self.stub.ReplicaRebuildStatus(empty_pb2.Empty())
+    
+    def sync_files(self, from_address, sync_file_info_tuples, fast_sync,
+                   file_sync_http_client_timeout):
+
+        sync_file_info_list = []
+        for tuple in sync_file_info_tuples:
+            info = common_pb2.SyncFileInfo(from_file_name="some_name",
```
james-munson, line 32: We don't even bother looking at the tuple, which could be anything, but we make a hard-coded info - the same one for each item in the array of tuples, if there are more than one.

james-munson (approved):
Just one nit in the python test code.  Otherwise LGTM.

app/cmd/sync_agent.go
```diff
@@ -87,10 +84,7 @@ func startSyncAgent(c *cli.Context) error {
 		return errors.Wrap(err, "failed to listen")
 	}
 
-	server := grpc.NewServer()
-	ptypes.RegisterSyncAgentServiceServer(server, syncagentrpc.NewSyncAgentServer(start, end, replicaAddress,
-		volumeName, replicaInstanceName))
-	reflection.Register(server)
```
shuo-wu: NIT: Is this necessary? All other server launching are still using this old pattern

generate_grpc.sh
```diff
@@ -17,6 +17,7 @@ buf lint
 protoc -I proto/ptypes/ -I proto/vendor/protobuf/src/ proto/ptypes/common.proto --go_out=plugins=grpc:proto/ptypes/
 python3 -m grpc_tools.protoc -I proto/ptypes/ -I proto/vendor/protobuf/src/ --python_out=integration/rpc/controller --grpc_python_out=integration/rpc/controller proto/ptypes/common.proto
 python3 -m grpc_tools.protoc -I proto/ptypes/ -I proto/vendor/protobuf/src/ --python_out=integration/rpc/replica --grpc_python_out=integration/rpc/replica proto/ptypes/common.proto
+python3 -m grpc_tools.protoc -I proto/ptypes/ -I proto/vendor/protobuf/src/ --python_out=integration/rpc/sync --grpc_python_out=integration/rpc/sync proto/ptypes/common.proto
```
shuo-wu, line 20: One question: Do we need to switch to the format of the instance-manager grpc generate script for this repo?

integration/rpc/sync/sync_agent_client.py
shuo-wu: This looks weird. Not sure why this API implementation ignores that content of input `sync_file_info_tuples` and directly generates a hard-coded list for `self.stub.FilesSync()` invocation.

shuo-wu (commented):
In general LGTM

integration/rpc/sync/sync_agent_client.py
ejweber: Good catch. There was some weirdness around imports (only on ARM) that caused me to pass this info to sync_files as a tuple, but I intended to actually use the info (no hard code it)! Didn't notice because the whole point of my new test is to fail without syncing a file. Fixing.

app/cmd/sync_agent.go
ejweber: I struggled a bit here, as each server seems to me to be launched a bit differently. I ultimately decided to move this logic to `pkg/snyc/rpc/server.go/` because the others make the `grpc.NewServer` call and add the interceptor within their respective packages.

generate_grpc.sh
ejweber: It would be good to do so, though there isn't a conflict yet that would require it. I would prefer to do that work in a separate issue. If you agree, I can create it.

proto/ptypes/controller.proto
```diff
@@ -48,6 +48,7 @@ message Volume {
 
 message ReplicaAddress {
     string address = 1;
+    string instanceName = 2;
```
PhanLe1010, line 51: Tracing through the code, I see that the purpose of adding this field is to help with the case:
1. longhorn-manager want the engine to do something with one of the replica (for example, adding)
2. longhorn-manager would send this info to engine  
3. engine would use this one to make sure it is talking to the correct replica instance (among its multiple replicas' instances)

This is working well if the operation just involves 1 replica. However, for an operation that involves multiple replicas such as PrepareRebuildReplica or ReplicaVerifyRebuild this is maybe not complete as:
1. The operation requires the engine to talk to multiple replicas: the target replica and the source replica
2. For the target replica, the engine knows the instance name from the parameter ReplicaAddress.InstanceName so it is good
3. However, for the source replica, engine doesn't know the instance name thus it skips the validation mechanism by setting empty string https://github.com/longhorn/longhorn-engine/pull/902/files#diff-d4c1f609d8fdb5a2cd78e82cd073ba59ec0ad3779ff8d86b935293e08b422646R56

In order to completely verify all replicas, I am thinking if we should make the engine to remember the instance name of all if its replicas. AKA adding instanceName [here](https://github.dev/longhorn/longhorn-engine/blob/d1d6d308d37a5b465bc1c2824da2e0702830de7d/pkg/types/types.go#L130-L133). WDYT?

pkg/sync/sync.go
```diff
@@ -186,7 +188,8 @@ func (t *Task) PurgeSnapshots(skip bool) error {
 		go func(rep *types.ControllerReplicaInfo) {
 			defer wg.Done()
 
-			repClient, err := replicaClient.NewReplicaClient(rep.Address)
+			// We don't know the replica's instanceName, so create a client without it.
+			repClient, err := replicaClient.NewReplicaClient(rep.Address, t.volumeName, "")
```
PhanLe1010, line 192: If we do this https://github.com/longhorn/longhorn-engine/pull/902#discussion_r1266149496, we will have instance name for verification here and similar Task function in this file

version.go
```diff
@@ -37,7 +37,10 @@ func version(c *cli.Context) error {
 	v := VersionOutput{ClientVersion: &clientVersion}
 
 	if !c.Bool("client-only") {
-		controllerClient, err := client.NewControllerClient(c.GlobalString("url"))
+		url := c.GlobalString("url")
+		volumeName := c.GlobalString("volume-name")
+		engineInstanceName := c.GlobalString("engine-instance-name")
```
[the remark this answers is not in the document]
ejweber: `engine-instance-name` is a global flag which can be applied to every command and is used by almost all commands to refer the the engine-instance a particular command communicates with (almost all commands operate primarily by communicating with an engine instance).

Some commands also take a `replica-instance-name` flag to refer to a particular replica (e.g. the `add-replica` command). IMO, it is important to distinguish between the two and keep a consistent naming scheme, so my preference is to not make this change.

proto/ptypes/interceptor.go
```diff
@@ -0,0 +1,74 @@
...
+func WithIdentityValidationReplicaServerInterceptor(volumeName, instanceName string) grpc.ServerOption {
+	return grpc.UnaryInterceptor(identityValidationServerInterceptor(volumeName, instanceName, "replica"))
+}
+
+func identityValidationServerInterceptor(volumeName, instanceName, serverType string) grpc.UnaryServerInterceptor {
+	// Use a closure to remember the correct volumeName and/or instanceName.
+	return func(ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (any, error) {
+		md, ok := metadata.FromIncomingContext(ctx)
+		if ok {
+			incomingVolumeName, ok := md["volume-name"]
```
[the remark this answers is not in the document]
ejweber: Yes! Noticed this when messing around with your panic comments. Will address them as one.

proto/ptypes/interceptor.go
```diff
@@ -0,0 +1,74 @@
...
+			// Only refuse to serve if both client and server provide validation information.
+			if ok && volumeName != "" {
+				log := logrus.WithFields(logrus.Fields{"method": info.FullMethod,
+					"clientVolumeName": incomingVolumeName[0], "serverVolumeName": volumeName})
+				if incomingVolumeName[0] != volumeName {
```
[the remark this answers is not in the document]
ejweber: There is no practical way this can happen, as all "correct" methods of generating the metadata on the client side (including the one we use) would at least provide `incomingVolumeName[0] == ""`. However, I did experiment with generating the metadata "incorrectly" and was able to trigger a panic (https://go.dev/play/p/ybP0IwGjvrX), so I am fixing this along with your comment above.

proto/ptypes/interceptor.go
```diff
@@ -0,0 +1,74 @@
...
+				}
+				log.Trace("Valid gRPC metadata")
+			}
+
+			incomingInstanceName, ok := md["instance-name"]
+			// Only refuse to serve if both client and server provide validation information.
+			if ok && instanceName != "" {
+				log := logrus.WithFields(logrus.Fields{"method": info.FullMethod,
+					"clientInstanceName": incomingInstanceName[0], "serverInstanceName": instanceName})
+				if incomingInstanceName[0] != instanceName {
```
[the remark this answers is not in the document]
ejweber: Same as above.

proto/ptypes/controller.proto
ejweber: You are right in everything you have posted here. We often DO skip instance name validation because the engine does not know the instance name of its replicas, and it WOULD be better if it did. The v2 engine is being written to include replica instance name as a "first class" identifier that the engine keeps track of.

However, when I set out to complete this work, I made refactoring longhorn-engine in this way a non-goal (https://github.com/longhorn/longhorn/blob/master/enhancements/20230420-engine-identity-validation.md#non-goals) because of the added complexity. IMO, the work I have done so far will fix the all observed real-world issues. 

Because:

- Replicas of the same volume rarely exist within the same instance-manager (and thus, are not susceptible to the issue in which one might quickly take the port of another, and
- Communicating with a different replica of the same volume would not incorrectly expand one of the replicas, etc,

The problem-set we would be addressing by doing as you suggest would be small compared to the work required. IMO we should complete this work as-is and potentially add a future enhancement to take it further if there is support to do so.

pkg/sync/sync.go
ejweber: Please see my response. Let's follow up offline for further discussion as necessary.

pkg/controller/client/controller_client.go
```diff
@@ -306,13 +307,14 @@ func (c *ControllerClient) ReplicaUpdate(address string, mode types.Mode) (*type
 	return GetControllerReplicaInfo(cr), nil
 }
 
-func (c *ControllerClient) ReplicaPrepareRebuild(address string) ([]types.SyncFileInfo, error) {
+func (c *ControllerClient) ReplicaPrepareRebuild(address, instanceName string) ([]types.SyncFileInfo, error) {
```
PhanLe1010, line 310: Can we can also add instanceName to [ReplicaCreate](https://github.com/longhorn/longhorn-engine/blob/0504cf983e5224b116368117b74bc358257da678/pkg/controller/client/controller_client.go#L263) for verification?

proto/ptypes/controller.proto
PhanLe1010: Thank you for the explanation. Agree that it is not worth the effort

pkg/meta/version.go
```diff
@@ -2,7 +2,7 @@ package meta
 
 const (
 	// CLIAPIVersion used to communicate with user e.g. longhorn-manager
-	CLIAPIVersion    = 8
+	CLIAPIVersion    = 9
 	CLIAPIMinVersion = 3
 
 	// ControllerAPIVersion used to communicate with instance-manager
```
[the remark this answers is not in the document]
ejweber: I think you are right. It seems CLIAPIVersion relates to whether or not I have changed CLI flags and ControllerAPIVersion relates to whether or not I have changed gRPC fields? Since I have done both, both should be bumped? And since in all cases I have made the new flags/fields optional, I do not need to bump MinVersions?

I too am confused on exactly how ControllerAPIVersion is used. instance-manager does not reference it, but maybe we are thinking this way:

If there is already an engine successfully running in an instance-manager with ControllerAPIVersion = x and MinControllerAPIVersion = y, clearly instance-manager is able to handle those values (even though instance-manager itself is not checking them). As long as we don't decrease ControllerAPIVersion (say to x - 1) or increase MinControllerAPIVersion (say to y + 1), the same instance-manager can run the new engine?

pkg/controller/client/controller_client.go
ejweber: I think not without modifying the controller's in-memory structures (as discussed above) and doing some in-depth modification of gRPC messages. 

`ReplicaPrepareRebuild` directly creates a replica client (which can make use of the instanceName) and its gRPC call takes a ReplicaAddress (which was easily modified to include an instanceName field).

`ReplicaCreate` creates a `Backend` structure and then uses it to facilitate communication with the replica. While adding an additional instanceName field to the `Remote` type of `Backend` is fairly easy, the `ReplicaCreate`gRPC call itself takes an `Address` (string) instead of a `ReplicaAddress` (modified with an instanceName field), so we would likely want to modify this (and similar requests for consistency).

In short, I think this work crosses the line we discussed previously. If we feel it is necessary, I propose we do it in a followup series of PRs.

Note that ReplicaCreate is only called in two places. In both places, we make a direct call to a replica with the instanceName BEFORE calling ReplicaCreate, so I don't think there's actually a practical benefit to modifying it.

- https://github.com/longhorn/longhorn-engine/blob/bee235337246b50b5c3dd86e46cfebdef3fe3434/pkg/sync/sync.go#L416-L428
- https://github.com/longhorn/longhorn-engine/blob/bee235337246b50b5c3dd86e46cfebdef3fe3434/pkg/sync/sync.go#L362-L372

PhanLe1010:
It looks like the [replica dataServer](https://github.com/longhorn/longhorn-engine/blob/bee235337246b50b5c3dd86e46cfebdef3fe3434/app/cmd/replica.go#L144) doesn't have authentication built in so any engine controller can connect to it.

However, this seems fine because every time we connect to it, we did the r.info() check with replica grpc server first https://github.com/longhorn/longhorn-engine/blob/bee235337246b50b5c3dd86e46cfebdef3fe3434/pkg/backend/remote/remote.go#L307-L319

PhanLe1010 (approved):
LGTM. Thanks for the great work!

pkg/controller/client/controller_client.go
PhanLe1010: I see. It sounds good to me now. Let's keep the current implementation

pkg/meta/version.go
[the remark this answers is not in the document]
PhanLe1010: > It seems CLIAPIVersion relates to whether or not I have changed CLI flags and ControllerAPIVersion relates to whether or not I have changed gRPC fields? Since I have done both, both should be bumped? And since in all cases I have made the new flags/fields optional, I do not need to bump MinVersions?

This is what I expected as well. But got confused about how ControllerAPIVersion is used as discussed above.

> If there is already an engine successfully running in an instance-manager with ControllerAPIVersion = x and MinControllerAPIVersion = y, clearly instance-manager is able to handle those values (even though instance-manager itself is not checking them). As long as we don't decrease ControllerAPIVersion (say to x - 1) or increase MinControllerAPIVersion (say to y + 1), the same instance-manager can run the new engine?

But in practice, we did increase MinControllerAPIVersion several times. So I am expecting there is some check in instance-manager but couldn't find any. Maybe this field is not used at all and we should combine it into CLIAPIVersion?

ejweber:
There is a new commit that fixes cloning, which is not included in the core tests. My code mistakenly assumed that the volume name we cloned from and to would be the same. This caused us to try to communicate with a from replica using the to volume name (triggering an invalid gRPC error). Please rereview when you are able.

cc @PhanLe1010 @shuo-wu @james-munson

pkg/sync/sync.go
```diff
@@ -117,7 +118,8 @@ func NewTask(ctx context.Context, controllerAddress string) (*Task, error) {
 	}()
 
 	return &Task{
-		client: controllerClient,
+		client:     controllerClient,
+		volumeName: volumeName,
```
shuo-wu, line 122: Can we cache `volumeName` in `ControllerClient`?

shuo-wu (commented):
In general LGTM. The comments are about some nitty issues. The main noticeable one is, should we cache and output `replicaName` for the replica list of `Controller`?

app/cmd/controller.go
```diff
@@ -96,9 +96,10 @@ func startController(c *cli.Context) error {
 	if c.NArg() == 0 {
 		return errors.New("volume name is required")
 	}
-	name := c.Args()[0]
+	volumeName := c.Args()[0]
+	// The global "--volume-name" flag is ignored here. It is redundant with the above required positional argument.
```
shuo-wu, line 100: I am thinking if we can use the global "--volume-name" flag here as well

PhanLe1010 (dismissed):
LGTM

(reviewed the newly added snapshot clone commit)

pkg/controller/rebuild.go
```diff
@@ -53,12 +53,12 @@ func (c *Controller) VerifyRebuildReplica(address string) error {
...
 	}
 
-	fromDisks, _, err := GetReplicaDisksAndHead(rwReplica.Address)
+	fromDisks, _, err := GetReplicaDisksAndHead(rwReplica.Address, c.Name, "")
 	if err != nil {
 		return err
 	}
 
-	toDisks, _, err := GetReplicaDisksAndHead(address)
+	toDisks, _, err := GetReplicaDisksAndHead(address, c.Name, instanceName)
```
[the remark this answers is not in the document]
ejweber: You mentioned it in the instance-manager PR: https://github.com/longhorn/longhorn-instance-manager/pull/229#discussion_r1251101973.

This is something @PhanLe1010 and I have discussed quite a bit, including in https://github.com/longhorn/longhorn-engine/pull/902#discussion_r1267028573.

Adding instanceName to types.Replica would be one thing, but we would need to make quite a few changes to controller client methods (which would propagate up the Longhorn stack) in order to make it useful. Because of the added complexity, and the limited practical benefit, we have steered away from doing this. If we do proceed that direction, I suggest we do it in a separate set of PRs.

Changing to VolumeName in the latest commit.

pkg/sync/sync.go
ejweber: I think this is somewhat awkward because of the way we reference it (instead of grabbing it directly from the task we must access it as an exported field of the controller client every time we instantiate a replica client. However, I don't feel strongly about it. Made the change in the latest commit.

app/cmd/controller.go
ejweber: I think it is a bit awkward either way, but prefer to leave it as is because:

1. `arg[0]` is a required parameter, while generally `--volume-name` is not.
1. `arg[0]` has been a part of the CLI API from (at least close to) the beginning and the information it contains is still relevant/necessary.

WDYT?

pkg/controller/rebuild.go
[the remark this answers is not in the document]
shuo-wu: > Because of the added complexity, and the limited practical benefit, we have steered away from doing this. If we do proceed that direction, I suggest we do it in a separate set of PRs.

Sometimes this lacking of the replica name makes the log debugging of the engine more difficult. And now with this info, we can make the identity validation between engines and replicas more convenient. To be honest, I prefer to have this change. WDYT? cc @innobead

shuo-wu (approved):

pkg/controller/rebuild.go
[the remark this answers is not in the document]
shuo-wu: YES. After adding it, we should input `<replica name>=<replica address>` for the CLI and a map for gRPC API `VolumeStart`. It's more trivial and complicated.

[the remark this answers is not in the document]
PhanLe1010: Agree that we should handle it in a new improvement ticket since this PR already satisfy its goal

ejweber:

PhanLe1010:

PR #903, opened 2023-05-15
feat: Azure Blob Storage Support

mantissahz:
Introduce Azure blob storage support and update `vendor/github.com/longhorn/backupstore`

Ref: longhorn/longhorn#1309


innobead (approved):

innobead:
@mantissahz Please make sure other repos depending on backupstore lib  also update as well.

PR #910, opened 2023-06-12
vendor: update backupstore

derekbit:
Longhorn/longhorn#6100


innobead (approved):
LGMT. Probably can also update other repos as well, even though they are not directly impacted, like instance manager, and longhorn-manager.

innobead:
@mergify backport v1.5.x

PR #917, opened 2023-07-10
common: use http profiler for runtime profiling

Vicente-Cheng:
Related to: https://github.com/longhorn/longhorn/issues/6282


shuo-wu (commented):
Not sure if it's for a single engine/replica process, or for the whole instance manager...
- If it's for a single instance process, the port may be an issue since there will be multiple processes in one instance manager pod.
- If it's for an instance manager pod, we need to add it in `longhorn-instance-manager` repo instead.

innobead:
Reopening this.

derekbit:
> Not sure if it's for a single engine/replica process, or for the whole instance manager...
> 
> * If it's for a single instance process, the port may be an issue since there will be multiple processes in one instance manager pod.

Same concern. cc @Vicente-Cheng 

> * If it's for an instance manager pod, we need to add it in `longhorn-instance-manager` repo instead.

Vicente-Cheng:
Sorry, I missed this one.

Yap, as @shuo-wu mentioned.
The profiler of longhorn-engine I would like to have is a single process profiler.
So, the questions would be the only first process that could export the specific port for profiling.

I would like to make the following changes.
1. Open a different port for each process (randomly chosen from the range, e.g., 6060 - 8080)
2. Provide the way to use the query which volume engine and its profiling port

Then, the user could port-forward the related volume and get the profiling data.
WDYT? cc @innobead, @derekbit, @shuo-wu

derekbit:
> Open a different port for each process (randomly chosen from the range, e.g., 6060 - 8080)

Is it possible to make the creation of a profiling server on demand?

Vicente-Cheng:
> > Open a different port for each process (randomly chosen from the range, e.g., 6060 - 8080)
> 
> Is it possible to make the creation of a profiling server on demand?

Yap, I did some testing for it. 
```
instance-manager-e-2300e576f6cffd254ac517b2a63c8a0b:/ # ss -tnal
State                 Recv-Q                Send-Q                                 Local Address:Port                                  Peer Address:Port                Process
LISTEN                0                     4096                                         0.0.0.0:3260                                       0.0.0.0:*
LISTEN                0                     4096                                            [::]:3260                                          [::]:*
LISTEN                0                     4096                                               *:10000                                            *:*
LISTEN                0                     4096                                               *:10001                                            *:*
LISTEN                0                     4096                                               *:8500                                             *:*
LISTEN                0                     4096                                               *:8501                                             *:*
LISTEN                0                     4096                                               *:6774                                             *:*
LISTEN                0                     4096                                               *:6775                                             *:*
```
`6774`, `6775` are the profiling server port.
Because they are individual processes, the profiler started with the process.
So it looks like we could create profiling server on demand (means each volume engine will have its profiling server)

We could add the args to enable the profiling server and record it on the engine status.
It would be easy to do a port-forward if the user would like to check any profiling data with the specific volume engine.

WDYT?

derekbit:
> > > Open a different port for each process (randomly chosen from the range, e.g., 6060 - 8080)
> > 
> > 
> > Is it possible to make the creation of a profiling server on demand?
> 
> Yap, I did some testing for it.
> 
> ```
> instance-manager-e-2300e576f6cffd254ac517b2a63c8a0b:/ # ss -tnal
> State                 Recv-Q                Send-Q                                 Local Address:Port                                  Peer Address:Port                Process
> LISTEN                0                     4096                                         0.0.0.0:3260                                       0.0.0.0:*
> LISTEN                0                     4096                                            [::]:3260                                          [::]:*
> LISTEN                0                     4096                                               *:10000                                            *:*
> LISTEN                0                     4096                                               *:10001                                            *:*
> LISTEN                0                     4096                                               *:8500                                             *:*
> LISTEN                0                     4096                                               *:8501                                             *:*
> LISTEN                0                     4096                                               *:6774                                             *:*
> LISTEN                0                     4096                                               *:6775                                             *:*
> ```
> 
> `6774`, `6775` are the profiling server port. Because they are individual processes, the profiler started with the process. So it looks like we could create profiling server on demand (means each volume engine will have its profiling server)
> 
> We could add the args to enable the profiling server and record it on the engine status. It would be easy to do a port-forward if the user would like to check any profiling data with the specific volume engine.
> 
> WDYT?

Yes, the on-demand method is better than the previous implementation.
Can it profile replica process as well?

Vicente-Cheng:
> > > > Open a different port for each process (randomly chosen from the range, e.g., 6060 - 8080)
> > > 
> > > 
> > > Is it possible to make the creation of a profiling server on demand?
> > 
> > 
> > Yap, I did some testing for it.
> > ```
> > instance-manager-e-2300e576f6cffd254ac517b2a63c8a0b:/ # ss -tnal
> > State                 Recv-Q                Send-Q                                 Local Address:Port                                  Peer Address:Port                Process
> > LISTEN                0                     4096                                         0.0.0.0:3260                                       0.0.0.0:*
> > LISTEN                0                     4096                                            [::]:3260                                          [::]:*
> > LISTEN                0                     4096                                               *:10000                                            *:*
> > LISTEN                0                     4096                                               *:10001                                            *:*
> > LISTEN                0                     4096                                               *:8500                                             *:*
> > LISTEN                0                     4096                                               *:8501                                             *:*
> > LISTEN                0                     4096                                               *:6774                                             *:*
> > LISTEN                0                     4096                                               *:6775                                             *:*
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `6774`, `6775` are the profiling server port. Because they are individual processes, the profiler started with the process. So it looks like we could create profiling server on demand (means each volume engine will have its profiling server)
> > We could add the args to enable the profiling server and record it on the engine status. It would be easy to do a port-forward if the user would like to check any profiling data with the specific volume engine.
> > WDYT?
> 
> Yes, the on-demand method is better than the previous implementation. Can it profile replica process as well?

Sure, I also checked with the replica process. (on the im-r pod)
It works.

derekbit:
A document or help description for the configure and the usage is also appreciated.

Vicente-Cheng:
> A document or help description for the configure and the usage is also appreciated.

Agree, I will do that! Thanks for the suggestion.

shuo-wu:
Do you mean that we can enable or disable the below HTTP profiling service on demands for **each** **existing** engine/replica process? And when enabling it, the upper layer IM pod will allocate one port for this HTTP service.
```
	go func() {
		debugAddress := "127.0.0.1:6060"
		debugHandler := http.DefaultServeMux
		logrus.Infof("Profiling Server (longhorn-engine) listening on %s", debugAddress)
		if err := http.ListenAndServe(debugAddress, debugHandler); err != nil && err != http.ErrServerClosed {
			logrus.Errorf(fmt.Sprintf("ListenAndServe: %s", err))
		}
	}()
```

Vicente-Cheng:
> Do you mean that we can enable or disable the below HTTP profiling service on demands for **each** **existing** engine/replica process? And when enabling it, the upper layer IM pod will allocate one port for this HTTP service.
> 
> ```
> 	go func() {
> 		debugAddress := "127.0.0.1:6060"
> 		debugHandler := http.DefaultServeMux
> 		logrus.Infof("Profiling Server (longhorn-engine) listening on %s", debugAddress)
> 		if err := http.ListenAndServe(debugAddress, debugHandler); err != nil && err != http.ErrServerClosed {
> 			logrus.Errorf(fmt.Sprintf("ListenAndServe: %s", err))
> 		}
> 	}()
> ```

Not sure I misunderstood the on-demand meaning from @derekbit.

Let's check again here.
1. We could enable/disable profiler with each volume engine when creating.
2. We could enable/disable the profiler handler on-demand.

The first one means we could setup profiler individual, which means some volume engines are enabled, and some volume engines disabled. We could disable it by default, only enable it if we want to do the profiling. (need to re-launch the volume engine)

The second one means we always enable the HTTP server but do not add a handler. Then add handler dynamic.
(Not 100% confirm that would work, but see some discussion with it). The advantage is we could enable profiling w/o re-launch, but we need to do more implementation.

I thought we were talking about the first one, but @shuo-wu reminds me of another on-demand design.

WDYT for it? cc @shuo-wu, @derekbit, @innobead

shuo-wu:
In other words, the 1st one means that users need to decide if a volume engine should start the profiler during/before the attachment. This may help save some resources... But I am not sure if profilers are resource-consuming...

As for the 2nd idea, YES that's how I understand "**on-demand**". If we prefer this, we need to add a new engine API and modify the port allocation for the IM. With this, users may easily fetch the info from the problematic engine as long as the engine is still able to handle requests. 

Notice that in both ideas, we need to modify the port allocation logic for IM pods.

Vicente-Cheng:
> In other words, the 1st one means that users need to decide if a volume engine should start the profiler during/before the attachment. This may help save some resources... But I am not sure if profilers are resource-consuming...
> 
> As for the 2nd idea, YES that's how I understand "**on-demand**". If we prefer this, we need to add a new engine API and modify the port allocation for the IM. With this, users may easily fetch the info from the problematic engine as long as the engine is still able to handle requests.
> 
> Notice that in both ideas, we need to modify the port allocation logic for IM pods.

Yes, for both ideas we need to do some port allocation logic for IM pods (means we need to manage the ports, including engine create/delete, and replica create/delete)

IIRC, the performance impact mainly occurs when we start to fetch the profiling data because of the following source https://cs.opensource.google/go/go/+/refs/tags/go1.21.1:src/net/http/pprof/pprof.go;l=133-157

I could do some benchmarks to confirm the impact w/ the profiler.

derekbit:
> Not sure I misunderstood the on-demand meaning from @derekbit.
> 
> Let's check again here.
> 
> 1. We could enable/disable profiler with each volume engine when creating.
>
> 2. We could enable/disable the profiler handler on-demand.

I mean the second method.

Vicente-Cheng:
> > Not sure I misunderstood the on-demand meaning from @derekbit.
> > Let's check again here.
> > 
> > 1. We could enable/disable profiler with each volume engine when creating.
> 
> This is a bit weird, because we don't know if we need to profile it when creating a volume in most situations.

For the first method, we could allocate the port for the volume engine and enable it if needed. (but we need to re-launch engine). The most conv is we could not do the runtime profile for the problematic engine(replica). 

> 
> > 2. We could enable/disable the profiler handler on-demand.
> 
> I mean the second method.

OK, I could try a simple implementation and benchmark w/ profiler and w/o profiler.

So, the most significant concern we do not want to enable profiler default is performance impact. Do I understand correctly?

derekbit:
> > > Not sure I misunderstood the on-demand meaning from @derekbit.
> > > Let's check again here.
> > > 
> > > 1. We could enable/disable profiler with each volume engine when creating.
> > 
> > 
> > This is a bit weird, because we don't know if we need to profile it when creating a volume in most situations.
> 
> For the first method, we could allocate the port for the volume engine and enable it if needed. (but we need to re-launch engine). The most cons is we could not do the runtime profile for the problematic engine(replica).

If an engine is problematic, we might want to profile it without any restart. It seems less useful than the 2nd method.
 
> 
> > > 2. We could enable/disable the profiler handler on-demand.
> > 
> > 
> > I mean the second method.
> 
> OK, I could try a simple implementation and benchmark w/ profiler and w/o profiler.
> 
> So, the most significant concern we do not want to enable profiler default is performance impact. Do I understand correctly?

Vicente-Cheng:
Hi @derekbit, @shuo-wu

I implemented the command for the engine side.
The usage would look like the following.

```
$ longhorn --url localhost:10000 profiler show
Profiler is not enabled
$ longhorn --url localhost:10000 profiler enable
Profiler enabled at Addr localhost:16060
$ longhorn --url localhost:10000 profiler show
Profiler enabled at Addr localhost:16060
$ longhorn --url localhost:10000 profiler disable
Profiler disabled!
$ longhorn --url localhost:10000 profiler show
Profiler is not enabled
```

I use the gRPC port (because the gRPC port should be unique) as the base and add the 6060 as the profiler port.

I will complete the replica side if this implementation is fine. (That should be very similar to the engine).

Thanks!

pkg/controller/control.go
```diff
@@ -1156,6 +1167,77 @@ func (c *Controller) GetLatestMetics() *ptypes.Metrics {
 	return metrics
 }
 
+func (c *Controller) ShowProfiler() (string, bool, error) {
```
shuo-wu, line 1170: Can we include the profiler info in API `ControllerServer.VolumeGet` output?

pkg/controller/control.go
```diff
@@ -1156,6 +1167,77 @@ func (c *Controller) GetLatestMetics() *ptypes.Metrics {
...
+	if c.profiler.ProfilerEnabled {
+		return c.profiler.ProfilerServer.Addr, true, nil
+	}
+	// grpcProt is unique, we could use it as the base port of profiler
+	grpcPort, err := strconv.Atoi(strings.Split(c.GRPCAddress, ":")[1])
+	if err != nil {
+		logrus.Warnf("Failed to convert the gRPC port to integer.")
+		return "", false, err
+	}
+	profilerPort := profilerPortBase + grpcPort
```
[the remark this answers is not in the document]
Vicente-Cheng: hi @shuo-wu, thanks for the review.
Yes, the profiler port is the sum of the grpcPort and the base (6060).

We have the unique gPRC server on each im-e pod, so unless we have other processes running on the im-e pod and use this port. It should not conflict.  Or I could try to add --profilerPort to the profiler enable command. To give users a chance to define which port they wanted. WDYT?

If the port is already binding by others, ListenAndServe will return an error. We may need to check the log to know that.

pkg/controller/control.go
```diff
@@ -85,6 +92,10 @@ func NewController(name string, factory types.BackendFactory, frontend types.Fro
 		DataServerProtocol:        dataServerProtocol,
 
 		fileSyncHTTPClientTimeout: fileSyncHTTPClientTimeout,
+		profiler: &types.Profiler{
+			ProfilerServer:  nil,
+			ProfilerEnabled: false,
+		},
```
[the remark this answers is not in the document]
Vicente-Cheng: Nice suggestion, I will update with that!

proto/ptypes/controller.proto
```diff
@@ -29,6 +29,8 @@ service ControllerService {
     rpc VersionDetailGet(google.protobuf.Empty) returns(VersionDetailGetReply);
 
     rpc MetricsGet(google.protobuf.Empty) returns(MetricsGetReply);
+
+    rpc ProfilerOperations(ProfilerOperationsRequest) returns(ProfilerOperationsReply);
```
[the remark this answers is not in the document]
Vicente-Cheng: Yes, we need. I will also add the replica side if this implementation is ok.

pkg/controller/control.go
Vicente-Cheng: Sure, I could also add it to VolumeGet

Vicente-Cheng:
Update the latest changes.
Some examples the following
```
instance-manager-e-2300e576f6cffd254ac517b2a63c8a0b:/ # /engine-binaries/longhornio-longhorn-engine-v1.4.3/longhorn --url localhost:10000 profiler --engine show
Profiler is not enabled
instance-manager-e-2300e576f6cffd254ac517b2a63c8a0b:/ # /engine-binaries/longhornio-longhorn-engine-v1.4.3/longhorn --url localhost:10000 profiler --engine disable
Profiler is disabled!
instance-manager-e-2300e576f6cffd254ac517b2a63c8a0b:/ # /engine-binaries/longhornio-longhorn-engine-v1.4.3/longhorn --url localhost:10000 profiler --engine enable
Profiler enabled at Addr localhost:16060
instance-manager-e-2300e576f6cffd254ac517b2a63c8a0b:/ # /engine-binaries/longhornio-longhorn-engine-v1.4.3/longhorn --url localhost:10000 info
{
        "name": "pvc-f5bd3e43-11ef-4afd-99ce-c236091282d5",
        "size": 107374182400,
        "replicaCount": 3,
        "endpoint": "/dev/longhorn/pvc-f5bd3e43-11ef-4afd-99ce-c236091282d5",
        "frontend": "tgt-blockdev",
        "frontendState": "up",
        "isExpanding": false,
        "lastExpansionError": "",
        "lastExpansionFailedAt": "",
        "unmapMarkSnapChainRemoved": false,
        "profilerAddress": "localhost:16060"
}
```

We could use the profiler command to enable profiler server runtime.
for engine side we need `--engine` flag, for replica side we need `--replica` flag
Also `enable` command supports the flag `--port` to specific the target port.
If we do not use `--port` flag, the default profiler server port would be the rpc server port + 6060.
(For example, if the rpc server is running on port 10001, the profiler server would be 16061)

We will see the error when the port is already binding when we try to enable profiler.
At that time, we could use `--port` to specify the target port we wanted.
 
Also, I am unsure how to resolve the `codeFactor` failure. I do not know why trigger `complexity = 127` because I added a similar function to other replica commands. Would you like to give me some tips for resolving it?

cc @derekbit, @shuo-wu, @innobead Thanks for your opinions!

pkg/controller/control.go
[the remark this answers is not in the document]
shuo-wu: In v1 volume implementation, **each** engine or replica is running as a separate process and **using some ports** inside the IM pod. And by design, [port 10000~30000 is reserved for engines and replicas](https://github.com/longhorn/longhorn-instance-manager/blob/f80354ec9a09e986f378017410891b26dddecc00/app/cmd/start.go#L49-L56) in IM pods. Hence, an engine or replica processing may be using port 16060. 
I am thinking if we can use more smaller number (<10000) as the 1st profiler port instead. Or maybe you can directly use 6060.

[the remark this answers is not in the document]
Vicente-Cheng: Hi @shuo-wu, Thanks for the explanation. I do not find this reservation.

OK, that I could use the profiler port base `20001`. Then, the profiler server will listen on the engine/replica port + 20001.
That means we will use 30001 ~ 50001 for the profiler server.
WDYT?

pkg/controller/control.go
```diff
@@ -1170,6 +1179,82 @@ func (c *Controller) GetLatestMetics() *ptypes.Metrics {
...
+	return "", true, nil
+}
+
+func (c *Controller) EnableProfiler(portNumber int32) (string, bool, error) {
+	c.profilerLock.Lock()
+	defer c.profilerLock.Unlock()
+
+	logrus.Infof("Prepare to do EnableProfiler")
+	var err error
+	if c.profilerServer != nil {
```
shuo-wu, line 1199: Do we need to return an error or do anything if the input port number does not match the existing server port?

proto/ptypes/replica.proto
```diff
@@ -173,3 +176,22 @@ message PrepareRemoveAction {
...
 }
+
+message RepProfilerOpCommonRequest {
+    string op = 1;
+}
+
+message RepProfilerOpEnableRequest {
+    string addr = 1;
+    int32 port_number = 2;
+}
```
shuo-wu, line 187: The API definition looks a little bit weird. I prefer to use either 3 separate APIs (enable, disable, show) or one single API (ReplicaProfileRequest --op <enable/disable/show> --port-number <for enable op>; the output is the port number (enable/show) or empty(disable))

pkg/controller/control.go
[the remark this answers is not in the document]
shuo-wu: I am fine with using either start port 30001+ or 6060+.
BTW, there is one minor issue: Each replica process would take 10 port numbers. Then for replica IM pods, the replica profiler ports will be like: 30001, 30011, 30021...... It's a little bit of waste

pkg/controller/control.go
Vicente-Cheng: Hi @shuo-wu, I do not understand that.

The `--port` argument lets the user define the port the profiler server was running. (It doesn't require)
So, in general, we assume that the user who uses `--port` will use a different port with the default port number.

Could you explain more about it?

proto/ptypes/replica.proto
Vicente-Cheng: Thanks for the suggestion!
I combine them into one. `RepProfilerOp` and `RepProfilerOpResponse`.
Please help to recheck it! Thanks!

pkg/controller/control.go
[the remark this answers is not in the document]
Vicente-Cheng: Hi @shuo-wu 
Because I would like to let users know that the gRPC port + 20001 would be the profiler server port (if enabled), I would like to keep that mechanism. 
Yes, that would be weird about the profiler port for the replica side. (Means 30001, 30011, 30021). But if the user would like to define the profiler server port continuously, they could use `--port` argument.
What do you think of that? Thanks!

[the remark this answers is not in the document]
shuo-wu: If you prefer not to use the port allocator in the IM server, I am fine with the current implementation.

pkg/controller/control.go
shuo-wu: For example, the caller sends 2 profiler enabling request with different port numbers for an engine, then the server will return successes for both requests. But actually, the input port number of the 2nd request does not match the running profiler port

Vicente-Cheng: Ah.. I know your means.
Indeed, this case should have a clear output.
Thanks!!  I will update that later.

pkg/replica/server.go
```diff
@@ -343,3 +357,79 @@ func (s *Server) PingResponse() error {
...
+	defer s.profilerLock.Unlock()
+
+	logrus.Infof("Prepare to do ShowProfilerConfig")
+	if s.profilerServer != nil {
+		return s.profilerServer.Addr, true, nil
+	}
+	return "", true, nil
+}
+
+func (s *Server) EnableProfiler(addr string, portNumber int32) (string, bool, error) {
```
shuo-wu, line 372: The boolean output is unnecessary. It will be `false` when there is an error. And it's not used in the client side.

pkg/controller/control.go
```diff
@@ -58,12 +61,16 @@ type Controller struct {
...
+
+	profilerServer *http.Server
+	profilerLock   sync.Mutex
 }
 
 const (
 	freezeTimeout         = 60 * time.Minute // qemu uses 60 minute timeouts for freezing
 	syncTimeout           = 60 * time.Minute
 	lastModifyCheckPeriod = 5 * time.Second
+	profilerPortBase      = 20001
```
shuo-wu, line 73: NIT: add a comment to explain why

shuo-wu (commented):
Others LGTM

pkg/controller/control.go
Vicente-Cheng: I added a comment and moved to `pkg/types/types.go` to avoid duplicates. Thanks!

pkg/replica/server.go
Vicente-Cheng: Good catch!
I add this result at first, indeed it is not used anymore.
Updated!

pkg/controller/control.go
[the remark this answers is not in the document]
Vicente-Cheng: OK, thanks for it!
Because the profiler server is not enabled by default, I would like to make the mechanism of the profiler server port on the internal engine instead of by IM server.

In general cases, we might enable the profiler server for a while (for profiling) and then disable it.

Feel free to let me know if there is something I do not consider. I would like to refine this!

pkg/controller/control.go
```diff
@@ -1170,6 +1177,86 @@ func (c *Controller) GetLatestMetics() *ptypes.Metrics {
...
+	if profilerPort == 0 {
+		var grpcPort int
+		grpcPort, err = strconv.Atoi(strings.Split(c.GRPCAddress, ":")[1])
+		if err != nil {
+			logrus.Warnf("Failed to convert the gRPC port to integer.")
+			return "", err
+		}
+		profilerPort = types.ProfilerPortBase + grpcPort
+	}
+	profilerAddr := fmt.Sprintf("localhost:%d", profilerPort)
```
derekbit, line 1216: Why not use `:%d` or `0.0.0.0:%d`?

Vicente-Cheng: directly use `:d`, updated!

app/cmd/profiler.go
```diff
@@ -0,0 +1,250 @@
...
+		return false
+	}
+	return true
+}
+
+func showProfiler(c *cli.Context) error {
+	engine := c.GlobalBool("engine")
+	replica := c.GlobalBool("replica")
+	if !opValidate(engine, replica) {
+		return fmt.Errorf("only `engine` or `replica` can be specified")
```
innobead, line 95: Move the error as a returned var of `opValidate`.

app/cmd/profiler.go
```diff
@@ -0,0 +1,250 @@
...
+		Name: "disable",
+		Action: func(c *cli.Context) {
+			if err := disableProfiler(c); err != nil {
+				logrus.WithError(err).Fatalf("Error running profiler disable")
+			}
+		},
+	}
+}
+
+func opValidate(engine, replica bool) bool {
```
innobead, line 81: `opValidate ....) error {`

This is a check function, so suggest returning an error instead of a bool var to have more info about why the check would fail.

app/cmd/profiler.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+	return showReplicaProfiler(c)
+
+}
+
+func enableProfiler(c *cli.Context) error {
+	portNumber := int32(c.Int("port"))
+	if !validatePortNumber(portNumber) {
+		return fmt.Errorf("port number should be bigger than 30000. The reserved port number is [10000 - 30000]")
+	}
```
innobead, line 108: The same here, move the error as a returned variable of the validate function.

pkg/controller/control.go
```diff
@@ -1170,6 +1177,86 @@ func (c *Controller) GetLatestMetics() *ptypes.Metrics {
 	return metrics
 }
 
+func (c *Controller) ShowProfiler() (string, error) {
+	c.profilerLock.Lock()
+	defer c.profilerLock.Unlock()
+
+	logrus.Infof("Prepare to do ShowProfilerConfig")
```
innobead, line 1184: `Preparing ..`

pkg/controller/control.go
```diff
@@ -1170,6 +1177,86 @@ func (c *Controller) GetLatestMetics() *ptypes.Metrics {
...
+		return c.profilerServer.Addr, nil
+	}
+	return "", nil
+}
+
+func (c *Controller) EnableProfiler(portNumber int32) (string, error) {
+	c.profilerLock.Lock()
+	defer c.profilerLock.Unlock()
+
+	logrus.Infof("Prepare to do EnableProfiler")
```
innobead, line 1195: `Preparing ...`

Generally, please use `Verb+ing/ed...` for logs.

pkg/controller/control.go
```diff
@@ -1170,6 +1177,86 @@ func (c *Controller) GetLatestMetics() *ptypes.Metrics {
 	return metrics
 }
 
+func (c *Controller) ShowProfiler() (string, error) {
+	c.profilerLock.Lock()
```
innobead, line 1181: Could be read lock.

pkg/controller/control.go
```diff
@@ -1170,6 +1177,86 @@ func (c *Controller) GetLatestMetics() *ptypes.Metrics {
...
+		}
+		logrus.Infof("Profiler server (%v) is closed.", c.profilerServer.Addr)
+	}()
+	logrus.Infof("Waiting the profiler server(%v) to start.", c.profilerServer.Addr)
+	// Wait for the profiler server to start, and check the profiler server.
+	time.Sleep(3 * time.Second)
+	if c.profilerServer == nil {
+		return "", fmt.Errorf("failed to start profiler server(%v)", profilerAddr)
+	}
+	return c.profilerServer.Addr, nil
```
innobead, line 1236: This waiting seems best effort to start the profiler. 

Can we just update the error into `profilerServer` if it can't be started, so we can simply know if the profiler is really running asynchronously by the show method. If it's not running, the caller can re-enable it.

pkg/controller/control.go
```diff
@@ -1170,6 +1177,86 @@ func (c *Controller) GetLatestMetics() *ptypes.Metrics {
...
+
+	logrus.Infof("Prepare to do DisableProfiler")
+	if c.profilerServer == nil {
+		return "", nil
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	if err := c.profilerServer.Shutdown(ctx); err != nil {
+		logrus.Warnf("Get error when close profiler server(%v): %v", c.profilerServer.Addr, err)
```
innobead, line 1251: `Failed to shutdown the profiler...`

pkg/controller/rpc/server.go
```diff
@@ -78,6 +78,10 @@ func (cs *ControllerServer) syncFileInfoToControllerFormat(info types.SyncFileIn
 
 func (cs *ControllerServer) getVolume() *ptypes.Volume {
 	lastExpansionError, lastExpansionFailedAt := cs.c.GetExpansionErrorInfo()
+	profilerAddr, _ := cs.c.ShowProfiler()
+	if profilerAddr == "" {
+		profilerAddr = "Profiler is not enabled"
+	}
```
innobead, line 84: Keep it empty, it's self-explanatory.

pkg/controller/rpc/server.go
```diff
@@ -239,6 +244,27 @@ func (cs *ControllerServer) JournalList(ctx context.Context, req *ptypes.Journal
 	return &empty.Empty{}, nil
 }
 
+func (cs *ControllerServer) EngProfilerOp(ctx context.Context, req *ptypes.EngProfilerOpRequest) (*ptypes.EngProfilerOpReply, error) {
+	logrus.Infof("Profiler operation: %v, port (only used with enable): %v", req.Op, req.PortNumber)
```
innobead, line 248: What does `port (only used with enable)` mean here? Isn't it straightforward? 

Suggest the log as below.
`Operating profiler: %v, port: %v`

pkg/replica/client/client.go
```diff
@@ -787,3 +787,22 @@ func (c *ReplicaClient) SnapshotHashLockState() (bool, error) {
 
 	return resp.IsLocked, nil
 }
+
+func (c *ReplicaClient) ReplicaProfilerOp(op, addr string, portNumber int32) (string, error) {
+	replicaServiceClient, err := c.getReplicaServiceClient()
```
innobead, line 792: Add a similar log as `EngineProfilerOp`

innobead (changes requested):
In general, got your intention. However, in the current PR, the implementation seems having several reduant parts. Therefore, I would recommend making an improvement to it and even make it reusable.

I suggest implementing this functionality in a common library (https://github.com/longhorn/go-common-libs), allowing us to use it across all primary running components, such as longhorn-manager, instance manager, share manager, backing image manager, and backing image manager.

We can create a gRPC service implementation in that lib, which can later be registered in the gRPC server of each component. WDYT?

innobead:
FYI. Some convention including logging, https://github.com/longhorn/longhorn/wiki/coding-convention.

Vicente-Cheng:
Hi @innobead,
I created https://github.com/longhorn/go-common-libs/pull/6 to create a general profiler service.
Also, resolve some comments on the above PR. I will keep the others for later usage on the engine side as a caller.

Thanks for the review!

innobead:
@Vicente-Cheng see if we can make this into 1.6.0 as well. Also, we need to backport this to the actively supported 1.4 and 1.5.

Vicente-Cheng:
> @Vicente-Cheng see if we can make this into 1.6.0 as well. Also, we need to backport this to the actively supported 1.4 and 1.5.

Sure, I will handle this today!

app/cmd/profiler.go
Vicente-Cheng: updated with the above suggestions.

Vicente-Cheng:
Hi folks,
I use the general grpc profiler service for this.
Please help to recheck it. Thanks!

main.go
```diff
@@ -145,6 +145,7 @@ func longhornCli() {
 		cmd.InfoCmd(),
 		cmd.FrontendCmd(),
 		cmd.SystemBackupCmd(),
+		cmd.ProfileCmd(),
```
innobead, line 148: `cmd.ProfilerCmd`.

app/cmd/profiler.go
```diff
@@ -0,0 +1,161 @@
...
+
+const (
+	opShow    = "SHOW"
+	opEnable  = "ENABLE"
+	opDisable = "DISABLE"
+
+	// ProfilerPortBase is the base number that we used to generate the profiler server port.
+	// We reserve the port 10000 - 30000 for engine v1/v2.
+	// We need to avoid to use them, so 20001 is big enough of the above interval.
+	profilerPortBase = 20001
```
innobead, line 24: `30001` instead?

innobead (changes requested):

The is pretty clean, thanks!

app/cmd/profiler.go
innobead: @Vicente-Cheng explained. 
> This is because we use port 10000 ~ 30000.
So if we do not define port number, we will use the grpc server port (means 10000 ~ 30000)
Then the grpc server port + profilerPortBase would be the final port -> means 30001 ~ 500001. It should be unique on each im pod.
So I think we can just keep this base on the 20001?

main.go
Vicente-Cheng: Updated!

innobead (approved):

PR #919, opened 2023-08-03
feat: talos support

c3y1huang:
https://github.com/longhorn/longhorn/issues/3161


go.mod
```diff
@@ -2,24 +2,28 @@ module github.com/longhorn/longhorn-engine
 
 go 1.20
 
+replace github.com/longhorn/go-iscsi-helper => github.com/c3y1huang/go-iscsi-helper v0.0.0-20230818102652-d94028f9f0e8
+
+replace github.com/longhorn/go-common-libs => github.com/c3y1huang/go-common-libs v0.0.0-20230818102433-680b5db81371
```
c3y1huang, line 7: TODO: update after https://github.com/longhorn/go-common-libs/pull/1

go.mod
```diff
@@ -2,24 +2,28 @@ module github.com/longhorn/longhorn-engine
 
 go 1.20
 
+replace github.com/longhorn/go-iscsi-helper => github.com/c3y1huang/go-iscsi-helper v0.0.0-20230818102652-d94028f9f0e8
```
c3y1huang, line 5: TODO: update after https://github.com/longhorn/go-iscsi-helper/pull/69

app/cmd/restore_to_file.go
```diff
@@ -198,7 +200,8 @@ func CheckBackingFileFormat(backingFilePath string) error {
 }
 
 func CopyFile(backingFilepath, outputFile string) error {
-	if _, err := iutil.ExecuteWithoutTimeout("cp", []string{backingFilepath, outputFile}); err != nil {
+	exec := lhutils.NewExecutor().Execute
+	if _, err := exec(nil, "cp", []string{backingFilepath, outputFile}, lhtypes.ExecuteNoTimeout); err != nil {
```
shuo-wu, line 204: NIT: Do we need this function. It seems that we can use `lhns.CopyFiles` instead

shuo-wu (commented):
Notice that some functions like `os.Remove` are not replaced by the functions in `go-common-libs/namespace/file.go` yet. Then the question is, when will we need to use the new ns util functions? Or you just forget to update that...?

app/cmd/restore_to_file.go
c3y1huang: Why do we need the namespace switching for this? IIRC it's just a normal command execution.

c3y1huang:
> Notice that some functions like `os.Remove` are not replaced by the functions in `go-common-libs/namespace/file.go` yet. Then the question is, when will we need to use the new ns util functions? Or you just forget to update that...?

IIRC the `os.Remove` calls within the existing codebase are intended to be executed within the current namespace. I believe I've already replaced all instances of nsenter commands with suitable implementations for switching namespaces. I don't think I've missed any, could you point out where?

shuo-wu:
Last time we talked about the joiner part, you mentioned that some util functions like file handling should use the joiner. Or in other words, when will we need to switch to a different namespace via a joiner?

c3y1huang:
> Last time we talked about the joiner part, you mentioned that some util functions like file handling should use the joiner. Or in other words, when will we need to switch to a different namespace via a joiner?

The namespace joiner is mainly used to replace the need for executing GNU core utilities such as  [sync](https://github.com/longhorn/longhorn-engine/pull/919/files#diff-c1efcd0538e5be63994d3d895fcbef838dbcd577e1262fb88d680d1672163477L192), [echo](https://github.com/longhorn/go-iscsi-helper/pull/69/files#diff-6ab38bea9b1590ef136807af5ef619338e844848ec653957bd07bc0093eaba6fL47), rm, cp, etc within a different namespace using nsenter. When a binary is executed without the need for nsenter wrapping, it runs within the current namespace. Therefore, there is no need for namespace switches using namespace joiner or namespace executor.

innobead (approved):
LGTM. Merge until the deps get updated.

PR #922, opened 2023-08-17
feat: engine version enforcement

c3y1huang:
https://github.com/longhorn/longhorn/issues/5842

### DIFF
--- a/pkg/meta/version.go
+++ b/pkg/meta/version.go
@@ -3,11 +3,11 @@ package meta
 const (
 	// CLIAPIVersion used to communicate with user e.g. longhorn-manager
 	CLIAPIVersion    = 9
-	CLIAPIMinVersion = 3
+	CLIAPIMinVersion = 8
 
 	// ControllerAPIVersion used to communicate with instance-manager
 	ControllerAPIVersion    = 5
-	ControllerAPIMinVersion = 3
+	ControllerAPIMinVersion = 4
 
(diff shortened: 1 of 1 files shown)


pkg/meta/version.go
```diff
@@ -3,11 +3,11 @@ package meta
 const (
 	// CLIAPIVersion used to communicate with user e.g. longhorn-manager
 	CLIAPIVersion    = 9
-	CLIAPIMinVersion = 3
+	CLIAPIMinVersion = 8
```
innobead, line 6: 8 was introduced in v1.5, so that means we only allow at least 1.5 engine cli being used in v1.6.

pkg/meta/version.go
```diff
@@ -3,11 +3,11 @@ package meta
...
 
 	// ControllerAPIVersion used to communicate with instance-manager
 	ControllerAPIVersion    = 5
-	ControllerAPIMinVersion = 3
+	ControllerAPIMinVersion = 4
```
innobead, line 10: 4 was introduced in v1.5, so that means we only allow at least 1.5 IM being used in v1.6.

innobead (approved):
LGTM.

We need test cases to test different versions of enging image and instance manager pod to ensure the expected behavior. Doc is needed as well.

innobead:
Also, we allow users to manage engine images with/ different versions, so we can keep doing that, but should not allow users to update an unsupported/incompatible version to any volumes.

PR #923, opened 2023-08-17
Add replica-instance-name argument to verify-rebuild-replica command

ejweber:
longhorn/longhorn#6522

https://github.com/longhorn/longhorn-manager/pull/2108 cause `verify-replica-rebuild` to be called with `--replica-instance-name`, but I missed adding the flag in longhorn/longhorn#5845.


PhanLe1010 (approved):
LGTM

PR #924, opened 2023-08-18
Fix 6481 - unbounded growth in BackupList from failure retries.

james-munson:
[BUG] Longhorn Instance Manager Memory leak #6481 

Three changes:  
 - Put in a log to catch runaway growth.
 - Ensure only one list entry per customer (backup ID).
 - Do the prune logic for "error" as well as "completed" status.

### DIFF
--- a/pkg/sync/rpc/list.go
+++ b/pkg/sync/rpc/list.go
@@ -4,14 +4,24 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/sirupsen/logrus"
+
 	"github.com/longhorn/longhorn-engine/pkg/replica"
 )
 
 const (
-	MaxBackupSize = 5
+	BackupListHighWaterMarkQuantum = 10
 
 	MaxSnapshotHashJobSize = 10
 )
+
+// Count of "completed" records to keep in BackupList when pruning.
+var retainBackupStateCounts = map[replica.ProgressState]int{
+	replica.ProgressStateComplete: 5,
+	replica.ProgressStateError:    10,
+}
+
+var backupListHighWaterMark = 0
 
 type BackupList struct {
 	sync.RWMutex
@@ -33,7 +43,7 @@ type SnapshotHashInfo struct {
 	job          *replica.SnapshotHashJob
 }
 
-// The APIs BackupAdd, BackupGet, Refresh, BackupDelete implement the CRUD interface for the backup object
+// The APIs BackupAdd, BackupGet, BackupDelete implement the CRUD interface for the backup object
 // The slice Backup.backupList is implemented similar to a FIFO queue.
 
 // BackupAdd creates a new backupList object and appends to the end of the list maintained by backup object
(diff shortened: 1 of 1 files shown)


pkg/sync/rpc/list.go
```diff
@@ -42,13 +52,24 @@ func (b *BackupList) BackupAdd(backupID string, backup *replica.BackupStatus) er
...
 	b.Lock()
 	b.infos = append(b.infos, &BackupInfo{
 		backupID:     backupID,
 		backupStatus: backup,
 	})
 	b.Unlock()
 
+	if backupListHighWaterMark < len(b.infos) {
+		backupListHighWaterMark = len(b.infos)
+		if 0 == backupListHighWaterMark%BackupListHighWaterMarkQuantum {
```
ejweber, line 68: I'm not sure I understand this. This only returns `true` if `backupListHighWaterMark` is a multiple of 10, right? Are we trying to limit how often we log (only once every ten increases)?

james-munson: Yep.  If we do have 1000 backups (possible) we don't really want to log 1000 times, but we don want to know if we've got a runaway list.  So this is just a simple-minded throttle.

ejweber: Got it. Is there any known way to have 1000 backups after these changes? Shouldn't we delete any more than 10 error and 5 completed in `refresh` on 73?

(Not saying I have a problem with the throttle. Just making sure I understand our expectations.)

pkg/sync/rpc/list.go
```diff
@@ -90,32 +111,34 @@ func (*BackupList) remove(b []*BackupInfo, index int) ([]*BackupInfo, error) {
...
 // Refresh deletes all the old completed backups from the front. Old backups are the completed backups
-// that are created before MaxBackupSize completed backups
+// that are created before the number we choose to retain.
 func (b *BackupList) refresh() error {
 	b.Lock()
 	defer b.Unlock()
 
-	var index, completed int
+	for state, limit := range retainBackupStateCounts {
+		var index, completed int
```
PhanLe1010, line 120: The variable `completed` is a bit confusing. It used to count both `error` and `completed` backups. Should we change it to `count` instead?

PhanLe1010 (approved):
LGTM except for a small NIT

PhanLe1010:
Could we run a e2e backup related tests for this one before merging? It is even better if run the full e2e test

pkg/sync/rpc/list.go
james-munson: It is certainly possible to have 1000 backups in progress - not yet either complete or error.  The refresh function only prunes those backups that have reached a conclusion.

ejweber (approved):
LGTM, question was answered.

I agree with @PhanLe1010 that we should link in a successful run of the end-to-end backup tests before merging.

derekbit:
@james-munson 
Shall we remove the `backup` from the list if failed to`DoBackupCreate`?
https://github.com/longhorn/longhorn-engine/blob/master/pkg/sync/rpc/server.go#L721

PhanLe1010:
Looks like this is closed in favor of https://github.com/longhorn/longhorn-engine/pull/927

PR #927, opened 2023-08-25
Fix 6481 - unbounded growth in BackupList from failure retries.

james-munson:
This is the same as the previous PR #924.  I made the mistake of renaming the branch, which closed that one.

[BUG] Longhorn Instance Manager Memory leak #6481

Three changes:

    Put in a log to catch runaway growth.
    Ensure only one list entry per customer (backup ID).
    Do the prune logic for "error" as well as "completed" status.

### DIFF
--- a/pkg/sync/rpc/list.go
+++ b/pkg/sync/rpc/list.go
@@ -111,23 +111,23 @@ func (*BackupList) remove(b []*BackupInfo, index int) ([]*BackupInfo, error) {
 }
 
 // Refresh deletes all the old completed backups from the front. Old backups are the completed backups
-// that are created before the number we choose to retain.
+// (finished or error) that were created before the number we choose to retain.
 func (b *BackupList) refresh() error {
 	b.Lock()
 	defer b.Unlock()
 
 	for state, limit := range retainBackupStateCounts {
-		var index, completed int
+		var index, count int
 
 		for index = len(b.infos) - 1; index >= 0; index-- {
 			if b.infos[index].backupStatus.State == state {
-				if completed == limit {
+				if count == limit {
 					break
 				}
-				completed++
+				count++
 			}
 		}
-		if completed == limit {
+		if count == limit {
 			// Remove all the older completed backups in the range backupList[0:index]
 			for ; index >= 0; index-- {
 				if b.infos[index].backupStatus.State == state {
--- a/pkg/sync/rpc/list.go
+++ b/pkg/sync/rpc/list.go
@@ -4,14 +4,24 @@ import (
 	"fmt"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


james-munson:
In #924 @derekbit asked:
"Shall we remove the backup from the list if failed to`DoBackupCreate`?  "

We could do so, but I don't think it is necessary.  I thought the whole point of the delayed pruning was to keep a short history of the most recent attempts - even failures.  As long as the status is "error" and not "0% done" it should be pruned eventually.  

But it would be an easy enough change.  WDYT?

PhanLe1010 (approved):
LGTM.

Looking forward to the e2e test result

derekbit:
> In #924 @derekbit asked: "Shall we remove the backup from the list if failed to`DoBackupCreate`? "
> 
> We could do so, but I don't think it is necessary. I thought the whole point of the delayed pruning was to keep a short history of the most recent attempts - even failures. As long as the status is "error" and not "0% done" it should be pruned eventually.
> 
> But it would be an easy enough change. WDYT?

Got it. We can keep it as it is. Thank you.

pkg/sync/rpc/list.go
```diff
@@ -42,13 +52,24 @@ func (b *BackupList) BackupAdd(backupID string, backup *replica.BackupStatus) er
...
+
 	b.Lock()
 	b.infos = append(b.infos, &BackupInfo{
 		backupID:     backupID,
 		backupStatus: backup,
 	})
 	b.Unlock()
 
+	if backupListHighWaterMark < len(b.infos) {
+		backupListHighWaterMark = len(b.infos)
```
shuo-wu, line 67: NIT: Should `len(b.infos)` be protected by the lock?

shuo-wu (approved):
In general LGTM

pkg/sync/rpc/list.go
james-munson: I was thinking not, since the HWM is just a rough indicator of whether there is a runaway memory allocation.  We don't use the exact value for any critical decisions.  If the length changes beneath us, it will be updated the next time around.

james-munson:
Full e2e tests passed, except for these four, which passed on individual re-run:
  + tests.test_backing_image.test_snapshot_with_backing_image
  + tests.test_basic.test_dr_volume_activated_with_failed_replica[s3]
  + tests.test_recurring_job.test_recurring_jobs_allow_detached_volume[nfs]
  + tests.test_recurring_job.test_recurring_jobs_when_volume_detached_unexpectedly[s3]

james-munson:

pkg/sync/rpc/list.go
innobead: Sounds good.

PR #933, opened 2023-09-08
feat: add snapshot max count and size

FrankYang0529:
ref: https://github.com/longhorn/longhorn/issues/6563


innobead (changes requested):
In general, LGTM. Some comments need to resolve first.

Also, for `250`, pleas use a constant for it instead if could, as it has been used in many places.

PhanLe1010:
Sorry. I miss this PR review. I will review this one tomorrow

pkg/controller/control.go
```diff
@@ -248,7 +252,7 @@ func (c *Controller) Expand(size int64) error {
 		// Should block R/W during the expansion.
 		c.Lock()
 		defer c.Unlock()
-		if _, err := c.backend.RemainSnapshots(); err != nil {
+		if err := c.backend.CanDoSnapshot(); err != nil {
```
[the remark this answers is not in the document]
FrankYang0529: When we do `expand`, we create a disk:

https://github.com/longhorn/longhorn-engine/blob/7ea91cba9f25e5d290d73a739c2e0fce02cdcacb/pkg/replica/replica.go#L1192-L1216

pkg/controller/control.go
```diff
@@ -661,6 +665,78 @@ func (c *Controller) checkUnmapMarkSnapChainRemoved() error {
...
+		// The related backend is nil if the mode is ERR
+		if r.Mode == types.ERR {
+			continue
+		}
+		err := c.backend.SetMaximumSnapshotCount(r.Address, count)
+		if err != nil {
+			logrus.Errorf("Failed to set flag MaximumSnapshotCount to %d in replica %s. Mark as ERR", count, r.Address)
+			c.setReplicaModeNoLock(r.Address, types.ERR)
+		} else {
+			allFailed = false
```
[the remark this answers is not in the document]
FrankYang0529: Yeah, we may have a case like some backend are updated, but some are not. The best practice is to revert previous change here.

However, from the perspective of whole system, I think it's okay. The request is from engine controller. If there is error, we will reconcile it next time.

ref: https://github.com/longhorn/longhorn-manager/blob/dda6a2973514af9663e9fde357c5fc5d27da8927/controller/engine_controller.go#L993-L996

pkg/controller/replicator.go
```diff
@@ -383,6 +383,39 @@ func (r *replicator) RemainSnapshots() (int, error) {
...
+	if len(r.backends) == 0 {
+		return nil
+	}
+
+	retCount := math.MaxInt32
+	for _, backend := range r.backends {
+		if backend.mode == types.ERR {
+			continue
+		}
+		// ignore error and try next one. We can deal with all error situation later
```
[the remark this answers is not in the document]
FrankYang0529: I followed the original implementation from `RemainSnapshots`. If we return errors from `backend` here, it means we only do snapshot when all backends are ready. Do you think it's better, or we want to keep current implementation? Thanks.

https://github.com/longhorn/longhorn-engine/blob/7ea91cba9f25e5d290d73a739c2e0fce02cdcacb/pkg/controller/replicator.go#L368-L374

pkg/controller/replicator.go
```diff
@@ -383,6 +383,39 @@ func (r *replicator) RemainSnapshots() (int, error) {
 	return ret, nil
 }
 
+func (r *replicator) CanDoSnapshot() error {
+	// addReplica may call here even without any backend
+	if len(r.backends) == 0 {
+		return nil
+	}
+
+	retCount := math.MaxInt32
```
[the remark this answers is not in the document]
FrankYang0529: We want to find a minimal remaining count from all backends. If we don't use `MaxInt32` as default value, we cannot distinguish whether `0` result is from backend or default value.

pkg/replica/client/client.go
```diff
@@ -165,6 +165,7 @@ func GetReplicaInfo(r *ptypes.Replica) *types.ReplicaInfo {
 		HeadFileSize:              r.HeadFileSize,
 		RevisionCounterDisabled:   r.RevisionCounterDisabled,
 		UnmapMarkDiskChainRemoved: r.UnmapMarkDiskChainRemoved,
+		RemainSnapshotSize:        r.RemainSnapshotSize,
```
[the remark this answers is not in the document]
FrankYang0529: We already have `RemainSnapshots` here.

https://github.com/longhorn/longhorn-engine/blob/7ea91cba9f25e5d290d73a739c2e0fce02cdcacb/pkg/replica/client/client.go#L163

app/cmd/replica.go
```diff
@@ -73,6 +73,15 @@ func ReplicaCmd() cli.Command {
...
 				Usage: "Name of the replica instance (for validation purposes)",
 			},
+			cli.IntFlag{
+				Name:  "snapshot-max-count",
+				Value: 250,
+				Usage: "Maximum number of snapshots to keep",
+			},
+			cli.StringFlag{
+				Name:  "snapshot-max-size",
+				Usage: "Maximum total snapshot size in bytes or human readable 42kb, 42mb, 42gb",
```
PhanLe1010, line 83: Does it mean that when:
1. User set the snapshot-max-size to 1GB
1. Write 2GB to the volume

Then cannot take the snapshot at all? Looks like a bit rough behavior. Could you provide the use-case for this feature? Thank you

FrankYang0529: The use case is that we want to provide a way for user to control snapshot space. We offer `SnapshotMaxCount` and `SnapshotMaxSize` to manage it. For `SnapshotMaxCount`, the minimal value is `2`, because it means `2` disks in longhorn-engine. One is snapshot and another is volume head.

For `SnapshotMaxSize`:
* If `SnapshotMaxSize` is 0, it means there is no snapshot size limit for a Volume.
* If `SnapshotMaxSize` is not 0. The minimum value for `SnapshotMaxSize` is same as `Size * 2` in a Volume, because a volume can have at least two snapshots.


LEP: https://github.com/longhorn/longhorn/pull/6634

PhanLe1010: Oh, so `SnapshotMaxSize` is the max value of sum(all snapshots' size) but not the max size of an individual snapshot. Got it!

pkg/controller/control.go
```diff
@@ -656,6 +660,78 @@ func (c *Controller) checkUnmapMarkSnapChainRemoved() error {
 	return nil
 }
 
+func (c *Controller) SetMaximumSnapshotCount(count int) error {
+	c.Lock()
+	defer c.Unlock()
+
+	c.maximumSnapshotCount = count
```
PhanLe1010, line 667: If we blindly set the c.maximumSnapshotCount to whatever the passed-in value `count` here, will it encounter this corner case:
1. c.maximumSnapshotCount is currently 10 and the current number of snapshot is 5
1. User set volume.Spec.SnapshotMaxCount to 2, we blindly set c.maximumSnapshotCount to 2
1. Now, the volume has more snapshots than c.maximumSnapshotCount which breaks the convention

IMO, we should failed to set the  c.maximumSnapshotCount to 2 while the current number of snapshots is already 5. WDYT?

pkg/controller/control.go
```diff
@@ -656,6 +660,78 @@ func (c *Controller) checkUnmapMarkSnapChainRemoved() error {
...
+	allFailed := true
+	for _, r := range c.replicas {
+		// The related backend is nil if the mode is ERR
+		if r.Mode == types.ERR {
+			continue
+		}
+		err := c.backend.SetMaximumSnapshotCount(r.Address, count)
+		if err != nil {
+			logrus.Errorf("Failed to set flag MaximumSnapshotCount to %d in replica %s. Mark as ERR", count, r.Address)
+			c.setReplicaModeNoLock(r.Address, types.ERR)
```
PhanLe1010, line 678: Will setting the replica to ERR a bit aggressive? This will trigger a rebuild which is a costly operation. Is it ok to just print out the log and return ERR so that engine can retry next time? 

cc @innobead  @shuo-wu @derekbit

pkg/controller/control.go
FrankYang0529: Yeah, we have related check in webhook. Do you think we also need to add it in longhorn-engine? Thanks.

https://github.com/longhorn/longhorn-manager/blob/4b5dbeca6ca731668d899492a8b11e878be6b47f/webhook/resources/volume/validator.go#L510-L524

PhanLe1010: I think enforcing at the engine level will make it 100% sure. The webhook get data from CRs which might be out of date (we have been dealing with stale cache data issue for a while now in webhook)

pkg/controller/control.go
innobead: Agreed.

I believe Poan just used the same implementation from other operations to apply here; however, for others, they are intended to fail the replica.

Here, I would suggest doing a 2-phase commit. Check all downstream are OK to set. If yes, the second commit will be applied. The same for MaxSnapshotSize.

FrankYang0529: For now, I think 2-phase commit is a little overengineering. We need to acquire lock for each backend in the first commit and apply the change in the second commit. Our request it to send an RPC call. If there is any error, we need to have a way to release the lock. There're too many cases need to handle. I would prefer to return an error and reconcile it next time.

innobead: Sounds good to me.

pkg/backend/remote/remote.go
```diff
@@ -211,16 +211,16 @@ func (r *Remote) SectorSize() (int64, error) {
 	return replicaInfo.SectorSize, nil
 }
 
-func (r *Remote) RemainSnapshots() (int, error) {
+func (r *Remote) GetSnapshotCountAndSizeUsage() (int, int64, error) {
```
innobead, line 214: nit: for multiple returned values, suggest having named variables.

pkg/controller/replicator.go
```diff
@@ -354,33 +353,58 @@ type backendWrapper struct {
...
+		if backend.mode == types.ERR {
+			continue
+		}
+		// ignore error and try next one
+		size, err := backend.backend.GetHeadFileSize()
+		if err != nil {
+			logrus.Errorf("Failed to get head file size usage: %v", err)
+			continue
+		}
+		hasResult = true
```
innobead, line 398: nit: ideally, should skip the loop if found the number.

pkg/controller/replicator.go
```diff
@@ -354,33 +353,58 @@ type backendWrapper struct {
...
-			if remain < ret {
-				ret = remain
-			}
+		// ignore error and try next one
+		count, size, err := backend.backend.GetSnapshotCountAndSizeUsage()
+		if err != nil {
+			logrus.Errorf("Failed to get snapshot count and size usage: %v", err)
+			continue
+		}
+		hasResult = true
```
innobead, line 370: nit: ideally, should skip the loop if found the number.

pkg/replica/replica.go
```diff
@@ -62,6 +60,9 @@ type Replica struct {
 	revisionCounterDisabled bool
 
 	unmapMarkDiskChainRemoved bool
+
+	snapshotMaxCount int
+	SnapshotMaxSize  int64
```
innobead, line 65: Why exported?

pkg/replica/server.go
```diff
@@ -19,15 +19,19 @@ type Server struct {
 	backing                   *backingfile.BackingFile
 	revisionCounterDisabled   bool
 	unmapMarkDiskChainRemoved bool
+	snapshotMaxCount          int
+	SnapshotMaxSize           int64
```
innobead, line 23: Why exported?

app/cmd/controller.go
```diff
@@ -83,6 +83,15 @@ func ControllerCmd() cli.Command {
...
 				Usage:    "HTTP client timeout for replica file sync server",
 			},
+			cli.IntFlag{
+				Name:  "snapshot-max-count",
+				Value: 250,
+				Usage: "Maximum number of snapshots to keep",
+			},
+			cli.StringFlag{
+				Name:  "snapshot-max-size",
+				Usage: "Maximum total snapshot size in bytes or human readable 42kb, 42mb, 42gb",
```
innobead, line 93: What values this exactly handle? KB, KiB or case-insensitive?

innobead (changes requested):
Just a few comments.

pkg/controller/replicator.go
FrankYang0529: Ideally, all replicas have same data. We can just get a result from one of them and return. However, the data may not be synchronized in some cases, so I prefer to check all replicas and return biggest value.

pkg/controller/replicator.go
FrankYang0529: Ideally, all replicas have same data. We can just get a result from one of them and return. However, the data may not be synchronized in some cases, so I prefer to check all replicas and return biggest value.

pkg/controller/replicator.go
innobead: Knew your intention. Sounds good.

app/cmd/controller.go
FrankYang0529: We use docker/go-units to parse the string. All formats are supported.

https://github.com/docker/go-units/blob/47565b4f722fb6ceae66b95f853feed578a4a51c/size.go#L34

pkg/backend/remote/remote.go
FrankYang0529: I updated the one in `pkg/controller/replicator.go`. We don't have calculation in remote.go. We just return result from another function, so I don't update it.

pkg/replica/replica.go
FrankYang0529: I updated naming and didn't notice that. Thank you.

pkg/replica/server.go
FrankYang0529: I updated naming and didn't notice that. Thank you.

innobead (approved):

PR #935, opened 2023-09-18
Upgrade packages version

FrankYang0529:
* protoc: 24.3
* protoc-gen-go: 1.5.3
* python grpcio: 1.58.0
* python protobuf: 4.24.3
* python libqcow: 20210419
* python tox: 4.11.3
* python directio: 1.3
* google.golang.org/grpc: v1.58.1
* empty.proto: 24.3
* replace github.com/golang/protobuf with google.golang.org/protobuf

ref: https://github.com/longhorn/longhorn/issues/6666


integration/common/frontend.py
```diff
@@ -10,7 +10,7 @@
 
 
 def readat_direct(dev, offset, length):
-    pg = offset / PAGE_SIZE
+    pg = offset // PAGE_SIZE
```
innobead, line 13: Why changed this?

integration/core/test_identity.py
```diff
@@ -103,7 +103,7 @@ def test_validation_fails_with_client(engine_manager_client,
     with pytest.raises(grpc.RpcError) as e:
         e_client.replica_create('tcp://' +
                                 get_process_address(replica_different))
-    assert e.value.code() == grpc.StatusCode.UNKNOWN
+    assert e.value.code() == grpc.StatusCode.FAILED_PRECONDITION
```
innobead, line 106: Why did the returned status change? Does this imply that there have been recent changes to the returned status of these methods?

The same questions to other status assertion change.

innobead (changes requested):
LGTM, except for two comments. Need to resolve them first.

integration/core/test_identity.py
FrankYang0529: The `FAILED_PRECONDITION` and `incorrect volume name` error message (in line 107) were introduced by https://github.com/longhorn/longhorn-engine/pull/902. The behavior change maybe cause of updating python grpcio from `1.25.0` to `1.58.0`. @ejweber, could you help to check whether status code `FAILED_PRECONDITION` is reasonable here? Thank you.

integration/common/frontend.py
FrankYang0529: I think we want to get page count here, so it's better to get an integer. Another line in `writeat_direct` also uses `//`.

https://github.com/longhorn/longhorn-engine/blob/c672c3a30435e50d3922b739ac274e15c7e19da1/integration/common/frontend.py#L33-L34

integration/core/test_identity.py
innobead: I see, but why did these cases not fail recently in PR checks before this change?

integration/common/frontend.py
innobead: cc @shuo-wu

integration/core/test_identity.py
FrankYang0529: I can't find any logic change in our code. The different part may be `serialized_options` in generated files:

* Old file: https://github.com/longhorn/longhorn-engine/blob/8469537f32a03e1f8b1fdfbb18c8a374d2785743/integration/rpc/controller/controller_pb2.py#L222
* PR change: https://github.com/FrankYang0529/longhorn-engine/blob/075afeeb813ab8a9798b4455ad099b63dc440a18/integration/rpc/controller/github/com/longhorn/longhorn_engine/proto/ptypes/controller_pb2.py#L28

ejweber: I'm not able to understand this yet.

The replica service returns FAILED_PRECONDITION to the controller service, but IIRC, the controller service doesn't automatically return the same code to the controller client. Since it doesn't recognize the error in particular, it just returns an UNKNOWN. That is why the original test checked for UNKNOWN and also the specific text on line 107.

Maybe some change now allows the FAILED_PRECONDITION to propagate from replica service -> controller service -> controller client?

innobead: Will these cases fail w/o these changes? If yes, we need to figure out the reasons.

integration/common/frontend.py
innobead: As discussed, please help create a ticket to review these read/write direct functions later.

integration/core/test_identity.py
innobead: After checking the code, the assertion change is unnecessary here. What you encountered should be somehow the call did not take effect due to service problems. One example in the current test cases is below where we hit the failed precondition error when the client is unable to reach out to the server. 

[/integration/core/test_identity.py#L80-L88](https://github.com/longhorn/longhorn-engine/blob/2e09ee964d970ab6cbd26b40d942e7b1e9d9a9f6/integration/core/test_identity.py#L80-L88)
```py
    # CASE 6:
    # Our sync agent client is created with a different instance name than the
    # sync agent.
    # We cannot communicate with the sync agent.
    s_client = SyncAgentClient(get_sync_agent_service_address(replica),
                               VOLUME_NAME, 'wrong')
    with pytest.raises(grpc.RpcError) as e:
        s_client.replica_rebuild_status()
    assert e.value.code() == grpc.StatusCode.FAILED_PRECONDITION
```

Just checked the current the engine gRPC implementation, which did not handle any specific gRPC status error when returning an error from a method call. In this case, gRPC will just return an unknown status code.

https://github.com/grpc/grpc-go/blob/v1.58.2/status/status.go#L96-L127

Suggest rerunning the test cases to figure out why the tests failed instead.

integration/common/frontend.py
FrankYang0529: We got error here because we upgrade Python version. In new version, `os.lseek` needs input type as integer.

```python
(function) def lseek(
    __fd: int,
    __position: int,
    __whence: int,
    /
) -> int
```

```
    def readat_direct(dev, offset, length):
        pg = offset / PAGE_SIZE
        in_page_offset = offset % PAGE_SIZE
        # either read less than a page, or whole pages
        if in_page_offset != 0:
            assert pg == (offset + length) / PAGE_SIZE
            to_read = PAGE_SIZE
        else:
            assert length % PAGE_SIZE == 0
            to_read = length
        pg_offset = pg * PAGE_SIZE
    
        f = os.open(dev, os.O_DIRECT | os.O_RDONLY)
        try:
>           os.lseek(f, pg_offset, os.SEEK_SET)
E           TypeError: 'float' object cannot be interpreted as an integer
```

innobead: The change is OK, as it's looking for the start page to write and read.

integration/core/test_identity.py
innobead: My fault. The change is necessary, as the current engine/replica gRPC server has been implemented with an interceptor which will issue `FAILED_PRECONDITION` if the engine can't communicate with the replica. 

[/proto/ptypes/interceptor.go#L54-L58](https://github.com/longhorn/longhorn-engine/blob/5464ea87c09fb0dd2aed575127b1d7c3756cfd12/proto/ptypes/interceptor.go#L54-L58)
```go
				if incomingInstanceName != instanceName {
					log.Error("Invalid gRPC metadata")
					return nil, status.Errorf(codes.FailedPrecondition, "incorrect instance name %s; check %s address",
						incomingInstanceName, serverType)
				}
```

innobead: cc @ejweber

innobead:
@FrankYang0529 It's good to merge after the PR check passes.

innobead (approved):

innobead:
@FrankYang0529 @derekbit merged

PR #938, opened 2023-10-13
fix: rebuilding in volume.meta blocks engine start

c3y1huang:
https://github.com/longhorn/longhorn/issues/6626

### DIFF
--- a/pkg/backend/remote/remote.go
+++ b/pkg/backend/remote/remote.go
@@ -258,7 +258,7 @@ func (r *Remote) SetUnmapMarkSnapChainRemoved(enabled bool) error {
 	conn, err := grpc.Dial(r.replicaServiceURL, grpc.WithInsecure(),
 		ptypes.WithIdentityValidationClientInterceptor(r.volumeName, ""))
 	if err != nil {
-		return fmt.Errorf("cannot connect to ReplicaService %v: %v", r.replicaServiceURL, err)
+		return errors.Wrapf(err, "failed connecting to ReplicaService %v", r.replicaServiceURL)
 	}
 	defer conn.Close()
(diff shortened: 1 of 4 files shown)


pkg/backend/remote/remote.go
```diff
@@ -266,6 +275,40 @@ func (r *Remote) SetUnmapMarkSnapChainRemoved(enabled bool) error {
...
+	if !isRebuilding {
+		return nil
+	}
+
+	logrus.Warnf("Resetting %v rebuild", r.name)
+
+	conn, err := grpc.Dial(r.replicaServiceURL, grpc.WithInsecure(),
+		ptypes.WithIdentityValidationClientInterceptor(r.volumeName, ""))
+	if err != nil {
+		return fmt.Errorf("cannot connect to ReplicaService %v: %v", r.replicaServiceURL, err)
```
derekbit, line 293: errors.Wrapf(err, "....")

pkg/controller/control.go
```diff
@@ -737,6 +737,13 @@ func (c *Controller) Start(volumeSize, volumeCurrentSize int64, addresses ...str
 			continue
 		}
 
+		if len(addresses) == 1 {
+			err = newBackend.ResetRebuild()
```
derekbit, line 741: Can we add a comment explaining the purpose of the reset of rebuilding flag for single replica volume?

derekbit (commented):
In general,LGTM.

innobead (approved):
LGTM, since this is for one replica case. However, this is not for fixing the root cause, but rather an resilience solution to recover from the failure (which could be caused by some concourrent operations between replica rebuilding)

derekbit (approved):

innobead (approved):

innobead:

PR #942, opened 2023-10-15
ci: pause integration testing for s390x

innobead:
https://drone-publish.longhorn.io/longhorn/longhorn-engine/718/1/3

`iscsiadm` not found from the s390x runner. After fixing this, the integration testing will be re-enabled in s390x pipeline. 

```
E           grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with:
E           	status = StatusCode.UNKNOWN
E           	details = "failed to init frontend: device test-a11a5832-volume: failed to stop iSCSI device: failed to logout target: failed to execute: nsenter [--mount=/host/proc/263111/ns/mnt --net=/host/proc/263111/ns/net iscsiadm --version], output , stderr nsenter: failed to execute iscsiadm: No such file or directory
```

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -234,7 +234,7 @@ steps:
   pull: default
   image: rancher/dapper:v0.5.8
   commands:
-  - dapper ci
+  - dapper ci-without-it
   privileged: true
   volumes:
   - name: socket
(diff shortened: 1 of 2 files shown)


innobead:

PR #946, opened 2023-10-18
poc: backport talos support to v1.4.x

c3y1huang:
https://github.com/longhorn/go-iscsi-helper/pull/69#issuecomment-1742350938


innobead:
don't think we need this. closing first.

PR #954, opened 2023-10-26
build: improve the gPRC generator script

Vicente-Cheng:
related to: https://github.com/longhorn/longhorn/issues/6973

Example for the usage.
```
# After we add/modify gRPC call
$ make generate-rpc
... some build logs
integration/rpc/controller/controller_pb2.py | 200 ++++++++++---- integration/rpc/controller/controller_pb2_grpc.py | 17 ++ integration/rpc/replica/re
plica_pb2.py | 114 +++++++- integration/rpc/replica/replica_pb2_grpc.py | 17 ++ proto/ptypes/controller.pb.go | 317 ++++++++++++++++------ proto/ptype
s/replica.pb.go | 307 +++++++++++++++------
Please update GRPC generated code.

$ git status
On branch tmp-gen-rpc
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   integration/rpc/controller/controller_pb2.py
        modified:   integration/rpc/controller/controller_pb2_grpc.py
        modified:   integration/rpc/replica/replica_pb2.py
        modified:   integration/rpc/replica/replica_pb2_grpc.py
        modified:   proto/ptypes/controller.pb.go
        modified:   proto/ptypes/replica.pb.go

no changes added to commit (use "git add" and/or "git commit -a")
```

Also, if we use `make` or `dapper`, the bulder_flag would be set. Then, we still keep the same behavior as before.

### DIFF
--- a/scripts/generate-grpc
+++ b/scripts/generate-grpc
@@ -2,6 +2,9 @@
 
 set -e
 
+# if BUILDER_FLAG is set, then the script will exit with the error code
+: ${BUILDER_FLAG:=0}
+
 cd $(dirname $0)/..
 
 ./generate_grpc.sh
@@ -12,5 +15,5 @@ if [ $? -eq 0 ]
 then
 	echo $output
 	echo GRPC generated code is not up to date
-	exit 1
+	exit $BUILDER_FLAG
 fi
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -141,5 +141,6 @@ RUN cd /go/src/github.com/longhorn && \
 
 VOLUME /tmp
 ENV TMPDIR /tmp
+ENV DAPPER_OUTPUT ./integration/rpc/controller ./integration/rpc/replica ./proto
 ENTRYPOINT ["./scripts/entry"]
 CMD ["ci"]
--- a/scripts/ci
+++ b/scripts/ci
@@ -3,7 +3,7 @@ set -e
 
 cd $(dirname $0)
 
-./prebuild
+BUILDER_FLAG=1 ./generate-grpc
 ./build
 ./validate
 ./test
--- a/scripts/ci-without-it
+++ b/scripts/ci-without-it
@@ -3,7 +3,7 @@ set -e
 
 cd $(dirname $0)
 
-./prebuild
+BUILDER_FLAG=1 ./generate-grpc
 ./build
 ./validate
 ./test


scripts/generate-grpc
```diff
@@ -2,15 +2,18 @@
 
 set -e
 
+# if BUILDER_FLAG is set, then the script will exit with the error code
```
innobead, line 5: Why we need to do this? I assume we should always fail with a consistent error code, so the original code looks good to me.

When we need to customize it?

innobead (changes requested):
Just one question needs to clarify .

scripts/generate-grpc
Vicente-Cheng: Yes, we need to `exit 1` to break this script.
So, in the build flow, we need to take it down if the gRPC check fails.

But to generate the gRPC code and keep the output.
We need to make this script exit normally (which means `exit 0`), so I made the exit variable on-demand.

innobead (changes requested):
LGTM. Please help remove prebuild, which is duplicated and meaningless. Then, we can merge this PR.

scripts/generate-grpc
innobead: Clear. Sorry for the late reply.

ref:
[/file/file.go#L135-L152](https://github.com/rancher/dapper/blob/c1e5b08924c6430528141534b704546c755cb8f8/file/file.go#L135-L152)
```go
	source := d.env.Source()
	output := d.env.Output()
	if !d.IsBind() && !d.NoOut {
		for _, i := range output {
			p := i
			if !strings.HasPrefix(p, "/") {
				p = path.Join(source, i)
			}
			targetDir := path.Dir(i)
			if err := os.MkdirAll(targetDir, 0755); err != nil {
				return err
			}
			logrus.Infof("docker cp %s %s", p, targetDir)
			if err := d.exec("cp", name+":"+p, targetDir); err != nil {
				logrus.Debugf("Error copying back '%s': %s", i, err)
			}
		}
	}
```

innobead:
ref: https://github.com/longhorn/longhorn/issues/7320

cc @FrankYang0529

FrankYang0529 (approved):
LGTM. Thanks for the PR.

innobead (approved):
Just noticed the prebuilt has been renamed 🙂

PR #955, opened 2023-11-13
build: reduce the image size by introducing multi stage build

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/5911

reduce from **756MB** to **392MB**

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.suse.com/bci/bci-base:15.5
+FROM registry.suse.com/bci/bci-base:15.5 AS builder
 
 ARG ARCH=amd64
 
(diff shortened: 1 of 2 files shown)


ChanYiLin:
Wait for e2e tests

ChanYiLin:
e2e tests **PASSED**

package/Dockerfile
```diff
@@ -1,40 +1,62 @@
...
+RUN zypper -n addrepo --refresh https://download.opensuse.org/repositories/system:/snappy/SLE_15/system:snappy.repo && \
+    zypper -n addrepo --refresh https://download.opensuse.org/repositories/network:/utilities/SLE_15_SP5/network:utilities.repo && \
+    zypper --gpg-auto-import-keys ref
+
+RUN zypper -n install kmod curl fuse wget \
+    nfs-client nfs4-acl-tools cifs-utils \
+    libibverbs librdmacm1 rdma-core-devel perl-Config-General libaio1 sg3_utils \
+    iputils telnet iproute2 qemu-tools \
+    e2fsprogs xsltproc && \
+    rm -rf /var/cache/zypp/*
```
derekbit, line 44: Should we add `zypper clean`?

ChanYiLin: I think `zypper clean` is equivalent to remove the `/var/cache/zypp/*` ?

derekbit: I think yes.

package/Dockerfile
```diff
@@ -1,40 +1,62 @@
...
+
+# Install grpc_health_probe
+RUN wget https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/v0.4.21/grpc_health_probe-linux-${ARCH} -O /usr/local/bin/grpc_health_probe && \
+    chmod +x /usr/local/bin/grpc_health_probe
```
derekbit, line 48: How about moving this line to `builder` and copying `grpc_health_probe` to `release`?
Then, we don't need to install `wget`.

package/Dockerfile
```diff
@@ -1,40 +1,62 @@
...
+ARG ARCH=amd64
+
+RUN zypper -n addrepo --refresh https://download.opensuse.org/repositories/system:/snappy/SLE_15/system:snappy.repo && \
+    zypper -n addrepo --refresh https://download.opensuse.org/repositories/network:/utilities/SLE_15_SP5/network:utilities.repo && \
+    zypper --gpg-auto-import-keys ref
+
+RUN zypper -n install kmod curl fuse wget \
+    nfs-client nfs4-acl-tools cifs-utils \
+    libibverbs librdmacm1 rdma-core-devel perl-Config-General libaio1 sg3_utils \
+    iputils telnet iproute2 qemu-tools \
```
derekbit, line 42: I'm curious about the purpose of some packages such as fuse, telnet, rdma-core-devel and so on. Do we need them?

ChanYiLin: I am also curios about them, might need to trial and error to see if we can remove them

package/Dockerfile
ChanYiLin: sounds good

package/Dockerfile
ChanYiLin: I think we still need to leave curl/iputils/iproute2 for user to debug network connection issue, WDYT?

derekbit: Sure!

derekbit (approved):

package/Dockerfile
```diff
@@ -1,40 +1,60 @@
...
+
+ARG ARCH=amd64
+
+RUN zypper -n addrepo --refresh https://download.opensuse.org/repositories/system:/snappy/SLE_15/system:snappy.repo && \
+    zypper -n addrepo --refresh https://download.opensuse.org/repositories/network:/utilities/SLE_15_SP5/network:utilities.repo && \
+    zypper --gpg-auto-import-keys ref
+
+RUN zypper -n install kmod curl \
+    nfs-client nfs4-acl-tools cifs-utils \
+    libibverbs librdmacm1 perl-Config-General libaio1 sg3_utils \
```
derekbit, line 44: kmod: operate kernel space
curl
libibverbs: for RDMA
librdmacm1: manage RDMA connection.
perl-Config-General: parsing tool?

The packages might be not required.

ChanYiLin: Let me remove them and try to run coretest to check

ChanYiLin: e2e tests PASSES after remove these packages
these are safe to removed
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5226/

innobead (approved):

package/Dockerfile
innobead: I guess that should be for debugging purposes, but users should be able to install them at runtime. (of course, it could be an airgap env). We can remove them first and see if there is any specific need in the future.

ChanYiLin:
@mergify backport v1.5.x

PR #969, opened 2023-12-05
ci(drone): remove privileged in build stage

yangchiu:
ci(drone): remove privileged in build stage

Signed-off-by: Yang Chiu [yang.chiu@suse.com](mailto:yang.chiu@suse.com)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -25,7 +25,6 @@ steps:
   image: rancher/dapper:v0.5.3
   commands:
   - dapper ci
-  privileged: true
   volumes:
   - name: socket
     path: /var/run/docker.sock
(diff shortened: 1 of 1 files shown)


innobead:
@shuo-wu @derekbit  I assume the engine integration testing would use the iSCSI kernel module. If yes, why after disabling privilege mode, the tests ran successfully?

innobead (approved):

innobead:
> @shuo-wu @derekbit I assume the engine integration testing would use the iSCSI kernel module. If yes, why after disabling privilege mode, the tests ran successfully?

The answer from @yangchiu . 
>  There is an env variable DAPPER_RUN_ARGS --privileged in Dockerfile.dapper, so the one in .drone.yaml could be not important.

However, the privilege is set beyond the drone, so removing it still works, which is quite weird because the privilege allowance should be supercharged by the upper layer, the drone. To clarify this, there will be a need to check the drone implementation, but currently, it's not worth it.

PR #971, opened 2023-12-12
Suppress repeated warning logs.

james-munson:
longhorn/longhorn#6406

I wrote the issue and this PR assuming that the logging was for an odd but not dangerous situation - a race between replicas with slightly changed contents so that one could not duplicate the exact unmap pattern of another in the same volume.  It's not recorded as an error (a question that was left pending in the original PR review.  https://github.com/longhorn/longhorn-engine/pull/742#discussion_r1026391759)

I don't know of an issue where it is seen to lead to data corruption.  Since it has been seen in practice so often, it seems reasonable to remove the log flooding as just a nuisance.

### DIFF
--- a/pkg/controller/multi_unmapper_at.go
+++ b/pkg/controller/multi_unmapper_at.go
@@ -7,6 +7,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"github.com/longhorn/longhorn-engine/pkg/types"
+	diskutil "github.com/longhorn/longhorn-engine/pkg/util/disk"
 )
 
 type MultiUnmapperAt struct {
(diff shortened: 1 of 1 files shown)


pkg/controller/multi_unmapper_at.go
```diff
@@ -57,7 +58,11 @@ func (m *MultiUnmapperAt) UnmapAt(length uint32, off int64) (int, error) {
 			if unmappedSize == 0 {
 				unmappedSize = n
 			} else if unmappedSize != n {
-				logrus.Warnf("One of the MultiUnmapper get different size %v from others %v", n, unmappedSize)
+				// Could log at debug level here, but a difference of 4096 is commonplace.  (Why is it?  TBD).
+				// Log anything bigger than that, which is expected to be uncommon.
+				if n > unmappedSize+diskutil.VolumeSectorSize || unmappedSize > n+diskutil.VolumeSectorSize {
+					logrus.Warnf("One of the MultiUnmapper %v got a very different size %v from others %v", u, n, unmappedSize)
+				}
```
PhanLe1010, line 65: Since we are not 100% sure if this error is unharmful, I am a bit hesitant to silence it. Can we aggregate the error and only print out the error after 1, 10, 100, 1000 times the same error occurs? I am thinking we can do this by adding more data structure into the MultiUnmapperAt struct and doing something similar to backoff mechanism

innobead:
@james-munson where we are?

james-munson:
I'm going to close this request and update the issue.

PR #986, opened 2024-01-17
fix: return the read sectors length when it reads the expansion part

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/6899

I think we should still return `len(read sectors)` when reading out of the boundary (expansion part of the specific image)
The reason is that, from the caller's perspective, it is reading from the `"Volume"`
```
func (r *Replica) ReadAt(buf []byte, offset int64) (int, error) {
	r.RLock()
	c, err := r.volume.ReadAt(buf, offset) // which will call diffDisk.fullReadAt()
	r.RUnlock()
	return c, err
}
```
So it has no knowledge about the snapshot chain underneath.

If it reads outside of the boundary of the `"Volume"`, we do return `io.ErrUnexpectedEOF` error [here](https://github.com/longhorn/longhorn-engine/blob/master/pkg/replica/diff_disk.go#L235)
But if it only reads outside of the boundary of `"specific snapshot image"`, and we allow it to do so, then we should consider it as reading a bunch of "0", and that is also true. The caller does get a bunch of "0". 
The caller should not take care of the situation. 
It might get confused that it doesn't get EOF error but how come the data length is not correct.

### DIFF
--- a/pkg/replica/diff_disk.go
+++ b/pkg/replica/diff_disk.go
@@ -220,32 +220,42 @@ func (d *diffDisk) fullReadAt(buf []byte, offset int64) (int, error) {
 	return count, nil
 }
 
-func (d *diffDisk) read(target byte, buf []byte, startOffset int64, startSector int64, sectors int64) (int, error) {
+// read called by fullReadAt reads the data from the target disk into buf
+// offset is the offset of the whole volume
+// startSector indicates the sector in the buffer from which the data should be read from the target disk.
+// sectors indicates how many sectors we are going to read
+func (d *diffDisk) read(target byte, buf []byte, offset int64, startSector int64, sectors int64) (int, error) {
 	bufStart := startSector * d.sectorSize
 	bufLength := sectors * d.sectorSize
-	offset := startOffset + bufStart
-	size, err := d.files[target].Size()
+	diskReadStartOffset := offset + bufStart
+	diskReadEndOffset := diskReadStartOffset + bufLength
+	diskSize, err := d.files[target].Size()
 	if err != nil {
 		return 0, err
 	}
+	volumeSize := d.size
 
-	// Reading the out-of-bound part is not allowed
-	if bufLength > d.size-offset {
-		logrus.Warn("Trying to read the out-of-bound part")
-		return 0, io.ErrUnexpectedEOF
+	count := 0
+
+	if diskReadStartOffset < diskSize {
+		newBuf := buf[bufStart : bufStart+min(bufLength, diskSize-diskReadStartOffset)]
+		n, err := d.files[target].ReadAt(newBuf, diskReadStartOffset)
+		if err != nil {
+			return n, err
+		}
+		count += n
+	}
+	if diskReadEndOffset <= diskSize {
+		return count, nil
 	}
 
-	// May read the expanded part
-	if offset >= size {
-		return 0, nil
+	// Read out of disk boundary but still in Volume boundary, we should remain the valid range with 0 and return the correct count
+	if diskReadEndOffset <= volumeSize {
+		return count + int(diskReadEndOffset-max(diskSize, diskReadStartOffset)), nil
 	}
-	var newBuf []byte
-	if bufLength > size-offset {
-		newBuf = buf[bufStart : bufStart+size-offset]
-	} else {
-		newBuf = buf[bufStart : bufStart+bufLength]
-	}
-	return d.files[target].ReadAt(newBuf, offset)
+
+	// Read out of Volume boundary, we should remain the valid range with 0 and return the correct count with EOF error
+	return count + int(volumeSize-max(diskSize, diskReadStartOffset)), io.ErrUnexpectedEOF
 }
 
 func (d *diffDisk) lookup(sector int64) (byte, error) {
--- a/pkg/replica/replica_test.go
+++ b/pkg/replica/replica_test.go
@@ -3,6 +3,7 @@ package replica
 import (
 	"crypto/md5"
 	"fmt"
+	"io"
 	"os"
 	"path"
 	"strconv"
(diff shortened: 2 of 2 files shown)


ChanYiLin:
## Test is now running
`-m coretest`
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6007/console

shuo-wu (commented):
The read count of the below `d.files[target].ReadAt(newBuf, offset)` can be smaller than `bufLength`. 

For example:
```
buf         [xxxxxxxx]
newBuf      [xxxx]
disk        [xxxx]
```
`diffDisk.read` will return `len(newBuf) = 4` rather than `len(buf) = 8`. Then the upper caller final result won't be the length the application wants to read

ChanYiLin:
Hi @shuo-wu This change only applies to when read start offset is larger than target disk size

```
func (d *diffDisk) read(target byte, buf []byte, startOffset int64, startSector int64, sectors int64) (int, error) {
	bufStart := startSector * d.sectorSize
	bufLength := sectors * d.sectorSize  // Jack: bufLength here is the read sectors for the "given target disk"
	offset := startOffset + bufStart
	size, err := d.files[target].Size()
	if err != nil {
		return 0, err
	}
	// Reading the out-of-bound part is not allowed
        // Jack: bufLength here is the read sectors for the "given target disk"
	if bufLength > d.size-offset {
		logrus.Warn("Trying to read the out-of-bound part")
		return 0, io.ErrUnexpectedEOF
	}

	// May read the expanded part
        // Jack: the change only applies when the read start offset is larger than the "target disk size"
        // directly return with all 0 and bufLength count
	if offset >= size {
		return int(bufLength), nil
	}

        // Jack: your comment above will fall into this part of code
        // where if the bufLength is larger than the remain size of the "target disk"
	var newBuf []byte
	if bufLength > size-offset {
		newBuf = buf[bufStart : bufStart+size-offset]
	} else {
		newBuf = buf[bufStart : bufStart+bufLength]
	}
	return d.files[target].ReadAt(newBuf, offset)
}
```

PhanLe1010:
> The read count of the below `d.files[target].ReadAt(newBuf, offset)` can be smaller than `bufLength`.
> 
> For example:
> 
> ```
> buf         [xxxxxxxx]
> newBuf      [xxxx]
> disk        [xxxx]
> ```
> 
> `diffDisk.read` will return `len(newBuf) = 4` rather than `len(buf) = 8`. Then the upper caller final result won't be the length the application wants to read

Good catch. So maybe we should always return `bufLength` like:
```go
func (d *diffDisk) read(target byte, buf []byte, startOffset int64, startSector int64, sectors int64) (int, error) {
	bufStart := startSector * d.sectorSize
	bufLength := sectors * d.sectorSize
	offset := startOffset + bufStart
	size, err := d.files[target].Size()
	if err != nil {
		return 0, err
	}

	// Reading the out-of-bound part is not allowed
	if bufLength > d.size-offset {
		logrus.Warn("Trying to read the out-of-bound part")
		return 0, io.ErrUnexpectedEOF
	}

	// May read the expanded part
	if offset >= size {
		return 0, nil
	}
	var newBuf []byte
	if bufLength > size-offset {
		newBuf = buf[bufStart : bufStart+size-offset]
	} else {
		newBuf = buf[bufStart : bufStart+bufLength]
	}
	count, err := d.files[target].ReadAt(newBuf, offset)
        if err != nil {
               return count, err
        }
        return int(bufLength), nil
}

```

ChanYiLin:
I have summarize all the case for `read()`

I think the point here is that when reading from the `"specific target disk"`
1. We read the data inside disk boundary => get count and check error
2. If `diskReadEndOffset` is inside the disk boundary => **Return the count with nil error**

But if the `diskReadEndOffset` is outside the disk boundary, then it will break into 2 cases
    - If `diskReadEndOffset` is also out of the volume boundary, then we fill in 0 to [volumeSize - max(dSize, diskReadStartOffset)] range and add count. **Return count and EOF**
    - if `diskReadEndOffset` is inside the volume boundary, then we fill in 0 to [diskReadEndOffset - max(dSize, diskReadStartOffset)] range and add count. **Return count with nil error**


```
Index   [0, 1, 2, 3, 4, 5, 6]
Disk    [_, x, x]
Volume  [_, _, _, _, _]

Read       [x, x]               => count = 2
Read          [x, 0]            => count = 2    `diskReadEndOffset` is inside the volume boundary
Read             [0, 0]         => count = 2    `diskReadEndOffset` is inside the volume boundary
Read                [0, _]      => count = 1, EOF    `diskReadEndOffset` is out of the volume boundary,
Read                   [_, _]   => count = 0, EOF    `diskReadEndOffset` is out of the volume boundary,
Read          [x, 0, 0, _]      => count = 3, EOF    `diskReadEndOffset` is out of the volume boundary,
Read             [0, 0, _]      => count = 2, EOF    `diskReadEndOffset` is out of the volume boundary,
```

In this case the uppercaller `diffDisk.fullReadAt` won't need to do any change, the count return from the `read()` has already consider the boundary issue, and the count should be correct no matter
- disk is smaller than volume
- read is out of disk boundary 
- read is out of volume boundary

and even Volume EOF, the count would be correct and the buf will still have valid data for valid range.

ChanYiLin:
## Manual Test - PASSED

test with the original [harvester test case](https://github.com/longhorn/longhorn/issues/6899#issuecomment-1880453217)
- cksum are the same
- filesystem is not corrupted

PhanLe1010 (commented):
We need some unit test for this modification

pkg/replica/diff_disk.go
```diff
@@ -222,30 +222,42 @@ func (d *diffDisk) fullReadAt(buf []byte, offset int64) (int, error) {
 
 func (d *diffDisk) read(target byte, buf []byte, startOffset int64, startSector int64, sectors int64) (int, error) {
 	bufStart := startSector * d.sectorSize
+
 	bufLength := sectors * d.sectorSize
-	offset := startOffset + bufStart
-	size, err := d.files[target].Size()
+	diskReadStartOffset := startOffset + bufStart
+	diskReadEndOffset := diskReadStartOffset + bufLength
+	dSize, err := d.files[target].Size()
```
derekbit, line 229: nit: `dSize` and `vSize` are confusing when reading the codes. Probably can replace them with a better names.

pkg/replica/diff_disk.go
```diff
@@ -222,30 +222,42 @@ func (d *diffDisk) fullReadAt(buf []byte, offset int64) (int, error) {
 
 func (d *diffDisk) read(target byte, buf []byte, startOffset int64, startSector int64, sectors int64) (int, error) {
 	bufStart := startSector * d.sectorSize
+
```
derekbit, line 225: Redundant line?

pkg/replica/diff_disk.go
```diff
@@ -222,30 +222,42 @@ func (d *diffDisk) fullReadAt(buf []byte, offset int64) (int, error) {
 
 func (d *diffDisk) read(target byte, buf []byte, startOffset int64, startSector int64, sectors int64) (int, error) {
```
derekbit, line 223: Can you help add a doc for explaining the parameters?

pkg/replica/diff_disk.go
```diff
@@ -222,30 +222,42 @@ func (d *diffDisk) fullReadAt(buf []byte, offset int64) (int, error) {
...
+		if bufLength > dSize-diskReadStartOffset {
+			newBuf = buf[bufStart : bufStart+dSize-diskReadStartOffset]
+		} else {
+			newBuf = buf[bufStart : bufStart+bufLength]
+		}
+		n, err := d.files[target].ReadAt(newBuf, diskReadStartOffset)
+		if err != nil {
+			return n, err
+		}
+		count += n
```
derekbit, line 248:
```suggestion
            var newBuf []byte
            newBuf = buf[bufStart : bufStart+min(bufLength, dSize-diskReadStartOffset)]
            n, err := d.files[target].ReadAt(newBuf, diskReadStartOffset)
            if err != nil {
                return n, err
            }
```

pkg/replica/diff_disk.go
```diff
@@ -222,30 +222,42 @@ func (d *diffDisk) fullReadAt(buf []byte, offset int64) (int, error) {
...
 	}
-	return d.files[target].ReadAt(newBuf, offset)
+
+	// Read out of disk boundary but still in Volume boundary, we should remain the valid range with 0 and return the correct count
+	if diskReadEndOffset <= vSize {
+		return count + int(diskReadEndOffset-max(dSize, diskReadStartOffset)), nil
+	}
+
+	// Read out of Volume boundary, we should remain the valid range with 0 and return the correct count with EOF error
+	return count + int(vSize-max(dSize, diskReadStartOffset)), io.ErrUnexpectedEOF
```
derekbit, line 260: Will you create a e2e test for the EOF case?

pkg/replica/diff_disk.go
ChanYiLin: Added, thanks

pkg/replica/diff_disk.go
ChanYiLin: thanks for the advice

pkg/replica/diff_disk.go
ChanYiLin: I will add a unit test for the function
how to handle EOF is the responsible of the upper application
so I don't think we need a full e2e test for EOF as long as we make sure we return EOF in correct cases

shuo-wu (dismissed):
LGTM

ChanYiLin:
Hi @PhanLe1010 @derekbit 
I have added unit tests, thanks

pkg/replica/diff_disk.go
derekbit: One question, will a generic block device such a loop device or disk return the EOF in the same situation?
Just for making sure Longhorn disk can respond as same as a generic block device.

derekbit (commented):
LGTM and just one remaining question.

pkg/replica/diff_disk.go
ChanYiLin: From the golang io package
I think yes, it will return available data with count < len(buf) and EOF like what we do here
```
// ReaderAt is the interface that wraps the basic ReadAt method.
//
// ReadAt reads len(p) bytes into p starting at offset off in the
// underlying input source. It returns the number of bytes
// read (0 <= n <= len(p)) and any error encountered.
//
// When ReadAt returns n < len(p), it returns a non-nil error
// explaining why more bytes were not returned. In this respect,
// ReadAt is stricter than Read.
//
// Even if ReadAt returns n < len(p), it may use all of p as scratch
// space during the call. If some data is available but not len(p) bytes,
// ReadAt blocks until either all the data is available or an error occurs.
// In this respect ReadAt is different from Read.
//
// If the n = len(p) bytes returned by ReadAt are at the end of the
// input source, ReadAt may return either err == EOF or err == nil.
//
// If ReadAt is reading from an input source with a seek offset,
// ReadAt should not affect nor be affected by the underlying
// seek offset.
//
// Clients of ReadAt can execute parallel ReadAt calls on the
// same input source.
//
// Implementations must not retain p.
type ReaderAt interface {
	ReadAt(p []byte, off int64) (n int, err error)
}
```

ChanYiLin:
```
// ReaderAt is the interface that wraps the basic ReadAt method.
//
// ReadAt reads len(p) bytes into p starting at offset off in the
// underlying input source. It returns the number of bytes
// read (0 <= n <= len(p)) and any error encountered.
//
// When ReadAt returns n < len(p), it returns a non-nil error
// explaining why more bytes were not returned. In this respect,
// ReadAt is stricter than Read.
//
// Even if ReadAt returns n < len(p), it may use all of p as scratch
// space during the call. If some data is available but not len(p) bytes,
// ReadAt blocks until either all the data is available or an error occurs.
// In this respect ReadAt is different from Read.
//
// If the n = len(p) bytes returned by ReadAt are at the end of the
// input source, ReadAt may return either err == EOF or err == nil.
//
// If ReadAt is reading from an input source with a seek offset,
// ReadAt should not affect nor be affected by the underlying
// seek offset.
//
// Clients of ReadAt can execute parallel ReadAt calls on the
// same input source.
//
// Implementations must not retain p.
type ReaderAt interface {
	ReadAt(p []byte, off int64) (n int, err error)
}
```

Both Reader and ReadAt in golang all will return partial data with count <= len(buf) with non err

derekbit: Thanks for the confirmation. LGTM.

derekbit (approved):

pkg/replica/diff_disk.go
ChanYiLin: In the end we use golang io ReadAt to read the file, so the way we handle EOF in our wrapper function follows the same semantics here
```
n, err := d.files[target].ReadAt(newBuf, diskReadStartOffset)
```

innobead (approved):

innobead:
@mergify backport v1.6.x v1.5.x v1.4.x

PR #1003, opened 2024-02-01
feat(auto-balance): disk in pressure

c3y1huang:

Issue longhorn/longhorn#4105

Introduce file local sync to copy the file when there is another running replica on the same node so the data transfer doesn't need to go through TCP.

`None`

`None`


app/cmd/add_replica.go
```diff
@@ -35,6 +35,11 @@ func AddReplicaCmd() cli.Command {
 				Required: false,
 				Usage:    "Enable fast file synchronization using change time and checksum",
 			},
+			cli.BoolFlag{
+				Name:     "sync-local",
+				Required: false,
+				Usage:    "sync local replica",
+			},
```
shuo-wu, line 42: NIT: Do we need this if we won't implement local sync for this debug cmdline?

proto/ptypes/syncagent.proto
```diff
@@ -65,6 +65,12 @@ message FilesSyncRequest {
   repeated SyncFileInfo sync_file_info_list = 3;
   bool fast_sync = 4;
   int32 file_sync_http_client_timeout = 5;
+  FileLocalSync local_sync = 6;
```
shuo-wu, line 68: NIT: I am thinking we can reuse the existing fields `from_address` and `to_host` for local sync. But a boolean field is required instead. I am fine with the current implementation anyway.

shuo-wu (approved):

innobead:
Need to wait https://github.com/longhorn/longhorn-engine/pull/1015. cc @FrankYang0529

pkg/types/types.go
```diff
@@ -203,3 +203,8 @@ func GRPCReplicaModeToReplicaMode(replicaMode enginerpc.ReplicaMode) Mode {
 	}
 	return ERR
 }
+
+type FileLocalSync struct {
+	Source string
+	Target string
```
derekbit, line 209: nit: suggest using `SourcePath` and `TargetPath`

pkg/sync/rpc/server.go
```diff
@@ -445,16 +447,77 @@ func (s *SyncAgentServer) FilesSync(ctx context.Context, req *enginerpc.FilesSyn
...
+		sourcePath := filepath.Join("/host", req.LocalSync.Source, info.FromFileName)
+		targetPath := filepath.Join("/host", req.LocalSync.Target, info.ToFileName)
+		targetPaths = append(targetPaths, targetPath)
+
+		log.Tracef("Copying file %v to %v", sourcePath, targetPath)
+
+		err = lhio.CopyFile(sourcePath, targetPath, true)
+		if err != nil {
+			return err
+		}
```
derekbit, line 498: `lhio.CopyFile` is based on `io.Copy` ([codes](https://github.com/longhorn/go-common-libs/blob/d78642cff5e1e810d919ba3db2698a4acff59a99/io/file.go#L137-L140)). `io.Copy` doesn't support sparse file copy.

pkg/sync/rpc/server.go
```diff
@@ -445,16 +447,77 @@ func (s *SyncAgentServer) FilesSync(ctx context.Context, req *enginerpc.FilesSyn
...
+
+		err := lhio.CheckIsFileSizeSame(sourcePath, targetPath)
```
derekbit, line 500: `CheckIsFileSizeSame` is checking the apparent sizes of the source and target files. We also need to check the actual file sizes (actual spaces occupied on the disk) as well.

derekbit (commented):
The flow is LGTM. However, we need to implement the sparse file copy rather than use `io.Copy`. The [doc](https://stackoverflow.com/a/43040498) might be helpful.

pkg/sync/rpc/server.go
c3y1huang: https://github.com/longhorn/go-common-libs/pull/49

pkg/sync/rpc/server.go
c3y1huang: https://github.com/longhorn/sparse-tools/pull/104

derekbit:
@c3y1huang  Is it ready?

c3y1huang:
@derekbit , this is ready for review.

pkg/sync/rpc/server.go
```diff
@@ -445,16 +447,77 @@ func (s *SyncAgentServer) FilesSync(ctx context.Context, req *enginerpc.FilesSyn
...
+		log.Tracef("Copying file %v to %v", sourcePath, targetPath)
+
+		err = sparse.SyncLocalFile(sourcePath, targetPath)
+		if err != nil {
+			return err
+		}
+
+		err := lhio.CheckIsFileSizeSame(sourcePath, targetPath)
+		if err != nil {
+			log.WithError(err).Tracef("Failed to check file size for file %v", sourcePath)
```
derekbit, line 502: Why not use Errorf?
Should it error out if the sizes are mismatched?

c3y1huang: I've updated to use Warnf as it seems more appropriate since we are only logging the error without returning it. What do you think? I am also fine changing it to Errorf if you think that is more appropriate.

pkg/sync/rpc/server.go
```diff
@@ -445,16 +447,77 @@ func (s *SyncAgentServer) FilesSync(ctx context.Context, req *enginerpc.FilesSyn
...
+
+		err = sparse.SyncLocalFile(sourcePath, targetPath)
+		if err != nil {
+			return err
+		}
+
+		err := lhio.CheckIsFileSizeSame(sourcePath, targetPath)
+		if err != nil {
+			log.WithError(err).Warnf("Failed to check file size for file %v", sourcePath)
+		}
```
derekbit, line 503: Should we return error here?

c3y1huang: Yes, it's a good idea to return the error and allow it to fallback to remote sync. Updated, thank you.

derekbit (approved):
LGTM

PR #1043, opened 2024-03-07
Automatically add Longhorn device to the group id 6 (disk group) by default

PhanLe1010:
More explanation is at https://github.com/longhorn/longhorn/issues/8088#issuecomment-1982300242

longhorn/longhorn#8088


ejweber:
Looks like this was handled in https://github.com/longhorn/longhorn-engine/pull/1017/ instead. Feel free to reopen if you disagree.

PR #1050, opened 2024-03-11
chore(vendor): update go-common-libs/go-iscsi-utils and resolve breaking changes

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7960

Update dependencis and resolve the breaking changes.
- go-iscsi-helper
- go-common-libs
The PR is mainly based on this PR: https://github.com/longhorn/longhorn-engine/pull/919/files#diff-0fa449830b5dc2b274b7c95a58b5071edc23f1e44af84a99c1fa9e600767829f


ChanYiLin:
**CORE TEST SUCCEEDED**
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6623/console

app/cmd/restore_to_file.go
```diff
@@ -93,7 +94,7 @@ func restore(url string, concurrentLimit int) error {
 	bar := pb.StartNew(100)
 	periodicChecker := time.NewTicker(PeriodicRefreshIntervalInSeconds * time.Second)
 
-	progressKeyName := requestedBackupName + util.RandomID(8)
+	progressKeyName := requestedBackupName + util.RandomID()
```
derekbit, line 97: Move `RandomID` function to go-common-libs.
nit: I'm fine with the parameter, randomIDLenth` which makes the function more generalized.

pkg/util/util.go
```diff
@@ -316,3 +287,11 @@ func GetAddresses(volumeName, address string, dataServerProtocol types.DataServe
...
 	}
 }
+
+func UUID() string {
+	return uuid.New().String()
+}
+
+func RandomID() string {
+	return UUID()[:randomIDLenth]
+}
```
derekbit, line 297: Move to go-common-libs.

derekbit (commented):
In general, LGTM.

app/cmd/restore_to_file.go
ChanYiLin: Fixed, thanks!

pkg/util/util.go
ChanYiLin: Fixed, thanks

derekbit (dismissed):
LGTM

app/cmd/restore_to_file.go
```diff
@@ -13,7 +13,9 @@ import (
 	"github.com/urfave/cli"
 	"gopkg.in/cheggaaa/pb.v2"
 
-	iutil "github.com/longhorn/go-iscsi-helper/util"
+	lhexec "github.com/longhorn/go-common-libs/exec"
+	lhtypes "github.com/longhorn/go-common-libs/types"
+	commonUtils "github.com/longhorn/go-common-libs/utils"
```
innobead, line 18: nit: suggest using the same prefix for the package alias. for example, here we can just use `commonExec`, `commonTypes` and `commonUtils`.

This minor consistent alignment can be improved later.

app/cmd/shutdown.go
```diff
@@ -20,7 +20,7 @@ func addShutdown(f func() (err error)) {
 	}
 
 	hooks = append(hooks, f)
-	logrus.Debugf("Added shutdown func %v", util.GetFunctionName(f))
+	logrus.Debugf("Added shutdown func %v", lhutils.GetFunctionPath(f))
```
innobead, line 23: Why not still use `lhutils.GetFunctionName`?

pkg/util/util.go
```diff
@@ -8,35 +8,30 @@ import (
...
 
-	HostProc = "/host/proc"
-
 	unixDomainSocketDirectoryInContainer = "/host/var/lib/longhorn/unix-domain-socket/"
 )
 
 const (
 	BlockSizeLinux = 512
+
+	randomIDLenth = 8
```
innobead, line 34: `randomIDLenth` still used?

innobead (changes requested):

pkg/util/util.go
ChanYiLin: Fixed.

app/cmd/restore_to_file.go
ChanYiLin: Fixed.

innobead:
Discussed with @ChanYiLin, all changes here are just from the master. If anything requires changes, let's fix them in the mater branch.

innobead (approved):

PR #1053, opened 2024-03-13
dataconn: Fix error log

shuo-wu:

Issue #

When the error is triggered, the log should print out the wrong `msg.MagicVersion` rather than its address


innobead (approved):

innobead:
@shuo-wu backport this if needed.

shuo-wu:
@mergify backport v1.5.x v1.6.x

PR #1064, opened 2024-03-21
Prevent the unmap length of a negative number

Vicente-Cheng:
Issue: https://github.com/longhorn/longhorn/issues/8235

Prevent the unmap length of a negative number

Also, add unittest for unmap operations

### DIFF
--- a/pkg/replica/diff_disk.go
+++ b/pkg/replica/diff_disk.go
@@ -290,10 +290,11 @@ func (d *diffDisk) lookup(sector int64) (byte, error) {
 }
 
 func (d *diffDisk) UnmapAt(unmappableDisks []string, length uint32, offset int64) (int, error) {
-	if length == 0 {
+	if length < uint32(d.sectorSize) {
 		return 0, nil
 	}
 
+	origLength := length
 	startSectorOffset := offset % d.sectorSize
 	endSectorOffset := (int64(length) + offset) % d.sectorSize
 
@@ -304,8 +305,15 @@ func (d *diffDisk) UnmapAt(unmappableDisks []string, length uint32, offset int64
 	if endSectorOffset != 0 {
 		length -= uint32(endSectorOffset)
 	}
-	if length <= 0 {
+	if length == 0 {
 		return 0, nil
+	}
+
+	// The the final length must be smaller than original length. The following case should not happen.
+	// The only case is the length is calculated to a negative number and then overflows a large number.
+	// This protection mechanism avoids unmapping the abnormal length and returns errors.
+	if origLength < length {
+		return 0, fmt.Errorf("final unmap length(%v) should not be larger than original length(%v)", length, origLength)
 	}
 
 	var unmappedSizeErr error
--- a/pkg/replica/replica_test.go
+++ b/pkg/replica/replica_test.go
@@ -1119,3 +1119,65 @@ func (s *TestSuite) TestUnmapMarkDiskRemoved(c *C) {
 	c.Assert(r.activeDiskData[1].Removed, Equals, false)
 	c.Assert(r.activeDiskData[1].Parent, Equals, "")
 }
+
+func (s *TestSuite) TestUnmap(c *C) {
+	dir, err := os.MkdirTemp("", "replica")
+	c.Assert(err, IsNil)
+	defer os.RemoveAll(dir)
+
+	r, err := New(8*b, b, dir, nil, false, false, 250, 0)
+	c.Assert(err, IsNil)
+	defer r.Close()
+
+	buf := make([]byte, 8*b)
+	fill(buf, 1)
+	_, err = r.WriteAt(buf, 0)
+	c.Assert(err, IsNil)
+
+	// default sector size is 4096, we follow this assumption in the test
+	// offset = 0, unmap length = 0: 0 sectors can be unmapped
+	ret, err := r.UnmapAt(0, 0)
+	c.Assert(err, IsNil)
+	c.Assert(ret, Equals, 0)
+
+	// offset = 0, unmap length = 512: 0 sectors can be unmapped
+	ret, err = r.UnmapAt(512, 0)
+	c.Assert(err, IsNil)
+	c.Assert(ret, Equals, 0)
+
+	// offset = 0, unmap length = 4096: 1 sector can be unmapped
+	ret, err = r.UnmapAt(b, 0)
+	c.Assert(err, IsNil)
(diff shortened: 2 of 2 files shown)


pkg/replica/diff_disk.go
```diff
@@ -304,10 +308,15 @@ func (d *diffDisk) UnmapAt(unmappableDisks []string, length uint32, offset int64
 	if endSectorOffset != 0 {
 		length -= uint32(endSectorOffset)
 	}
-	if length <= 0 {
+	if length == 0 {
```
derekbit, line 311: Is it possible that length is smaller than 0?

pkg/replica/diff_disk.go
```diff
@@ -304,10 +308,15 @@ func (d *diffDisk) UnmapAt(unmappableDisks []string, length uint32, offset int64
...
 		return 0, nil
 	}
 
+	// final check
+	if origLength < length {
+		return 0, fmt.Errorf("FATAL, final unmap length(%v) should not be larger than original length(%v)", length, origLength)
```
derekbit, line 317:
```suggestion
		return 0, fmt.Errorf("final unmap length(%v) should not be larger than original length(%v)", length, origLength)
```

pkg/replica/diff_disk.go
```diff
@@ -304,10 +308,15 @@ func (d *diffDisk) UnmapAt(unmappableDisks []string, length uint32, offset int64
 	if endSectorOffset != 0 {
 		length -= uint32(endSectorOffset)
 	}
-	if length <= 0 {
+	if length == 0 {
 		return 0, nil
 	}
 
+	// final check
+	if origLength < length {
```
derekbit, line 316: Is it possible?

pkg/replica/diff_disk.go
```diff
@@ -293,7 +293,11 @@ func (d *diffDisk) UnmapAt(unmappableDisks []string, length uint32, offset int64
 	if length == 0 {
 		return 0, nil
 	}
+	if length < uint32(d.sectorSize) {
+		return 0, nil
+	}
```
derekbit, line 298:
```suggestion
        if length == 0 || length < uint32(d.sectorSize) {
	        return 0, nil
        }
```

pkg/replica/diff_disk.go
shuo-wu: Why cannot this help handle the negative `length` case?

Vicente-Cheng: Because the length is uint32, it overflows when the result is negative.
The original check was fine, but it could not capture the negative case, so I changed it to `==` instead of `<=`

pkg/replica/diff_disk.go
Vicente-Cheng: Agree, thanks for the suggestions!

pkg/replica/diff_disk.go
Vicente-Cheng: No, I thought it was impossible after this patch.
I would like to have this because I am afraid I missed some cases.
We could avoid the unexpected punch hole length and get the error.

WDYT?

derekbit: I'm fine with it. Can you add a comment for the check to avoid any confusion?

Vicente-Cheng: Sure, updated the comment!

pkg/replica/diff_disk.go
derekbit: OK. Using `<=` rather than `==` is due to that the data type of `length` is unsigned.

derekbit (approved):

derekbit:
```
Running: golangci-lint run --new-from-patch=1064.patch
Running: go fmt
pkg/replica/diff_disk.go
time="2024-03-22T06:55:58Z" level=fatal msg="exit status 1"
```

@Vicente-Cheng Can you check the faiulre?

Vicente-Cheng:
> ```
> Running: golangci-lint run --new-from-patch=1064.patch
> Running: go fmt
> pkg/replica/diff_disk.go
> time="2024-03-22T06:55:58Z" level=fatal msg="exit status 1"
> ```
> 
> @Vicente-Cheng Can you check the failure?

Fixed... some indent not align.

Also, remove the `length == 0` because it is covered by `length < uint32(d.sectorSize)` with your suggestions.
(Sorry for the wrong rebase, updated.)

Please help to recheck, thanks!

derekbit (approved):
LGTM

pkg/replica/diff_disk.go
shuo-wu: the unexpected punch hole length? I was thinking it handles `length` being overflowed to a large number

Vicente-Cheng: The length being overflowed to a large number means it might be a large number for the punch hole length.
That's why I say the unexpected punch hole length.

This checks here because I want to create some punch hole protection.
The final length is bigger than the original length, which means the extra sector will be punch hole. It caused data corruption.

I update the comment again. Hope that might be more clear.
Thanks!

derekbit:
One more question, what's the size of the length that `UnampAt received in the corrupted filesystem?

Vicente-Cheng:
> One more question, what's the size of the length that `UnampAt received in the corrupted filesystem?
 I thought the 512 bytes

shuo-wu (dismissed):

pkg/replica/replica_test.go
```diff
@@ -1119,3 +1119,65 @@ func (s *TestSuite) TestUnmapMarkDiskRemoved(c *C) {
...
+	ret, err = r.UnmapAt(512, 1)
+	c.Assert(err, IsNil)
+	c.Assert(ret, Equals, 0)
+
+	// offset = 1, unmap length = 4096: 0 sector can be unmapped
+	ret, err = r.UnmapAt(b, 1)
+	c.Assert(err, IsNil)
+	c.Assert(ret, Equals, 0)
+
+	// offset = 2048, unmap length = 4096: 1 sector can be unmapped
```
innobead, line 1172: // offset = 2048, unmap length = 8192: 1 sector can be unmapped

innobead (changes requested):
LGTM, just one comment.

innobead:

pkg/replica/replica_test.go
Vicente-Cheng: nice catch! Updated!

innobead (approved):

innobead:
cc @DamiaSan fyi

PR #1082, opened 2024-04-22
[V1.6.x] fix: missing rpc from longhorn-instance-manager

FrankYang0529:

https://github.com/longhorn/longhorn/issues/6744

https://github.com/longhorn/longhorn-engine/pull/1081#issuecomment-2068079458


derekbit (approved):
LGTM.

PR #1083, opened 2024-04-24
Add fsfreeze functionality to snapshot

ejweber:

longhorn/longhorn#2187

- Attempt to bind mount the file system on the root partition of a Longhorn volume and freeze it with fsfreeze if requested by snapshot command line argument or snapshot gRPC command. Otherwise, just sync like we have always done.
- Attempt to unfreeze the file system on the root partition during frontend shutdown in case we are in the middle of a snapshot.
- Attempt to unfreeze the file system on the root partion of a Longhorn volume during frontend startup in case its engine previously crashed before an unfreeze was issued.

Depends on https://github.com/longhorn/types/pull/8.

We need to roll back the debugging commit and fix the imports before merging.


ejweber:
I found a pretty convenient way to modify the integration test that was written for this feature's original implementation to work with the current implementation. It tests basic functionality (e.g., when `Snapshot` is called with `shouldFreeze == true`, we can successfully detect the mounted file system, do the bind mound, and attempt a freeze).

Otherwise, I manually tested the following scenarios using this modified engine and the modified instance manager. In the interest of time, I write up an exhaustive description of how the tests were executed, but here are the basic commands involved:

- Run a simple Longhorn volume in a Docker container.
  `docker run --privileged -v /:/host --rm --net host --name test longhornio/longhorn-engine:2149c73d launch-simple-longhorn test 10g tgt-blockdev`
- Make a file system on the volume and mount it.
  `mkfs.ext4 /dev/longhorn/test`
  `mount /dev/longhorn/test /mnt/test`
- Constantly write data to the file system while snapshotting, etc.
  `while true; do dd of=/mnt/test/file if=/dev/urandom bs=16M; done`
- Watch the dirty page cache while freeze is ongoing.
  `while true; do cat /proc/meminfo | grep -i dirty; sleep 0.5; done`
- Check inside the Docker container for mounted volumes.
  `docker exec test bash -c " while true; do mount | grep test; sleep 0.5; done"`
- Create a snapshot with or without a request to freeze.
  `bin/longhorn -url localhost:10015 snapshot create --freeze-fs`
  `bin/longhorn -url localhost:10015 snapshot create`
- Check for interesting processes and determine if they are stuck and why.
  `while true; do ps -eo pid,ppid,pgid,user,stat,pcpu,comm,wchan:32 | grep -e freeze -e dmsetup -e " dd" -e laun; sleep 1; done`
- Delete and recreate Longhorn processes using instance-manager.
  `docker exec test longhorn-instance-manager process delete --name test-e`
  `docker exec test longhorn-instance-manager process delete --name test-r`
  `docker exec test longhorn-instance-manager process create --name test-r --binary /usr/local/bin/longhorn --port-count 15 --port-args --listen,localhost: -- replica /volume/ --size 10g`
  `docker exec test longhorn-instance-manager process create --name test-e --binary /usr/local/bin/longhorn --port-count 1 --port-args --listen,localhost: -- controller test --frontend tgt-blockdev --size 10g --current-size 10g --replica tcp://localhost:10000`
- Kill the whole Longhorn volume immediately.
  `kill -9 -<process_group of the launch process>`

Here are the test cases. All tests were executed on a known "good" kernel (Rocky 9.3 with an upstream `6.8.7-1.el9.elrepo.x86_64` kernel installed).

---

Take a snapshot without --freeze-fs.

Check:
  Logs clearly indicate system sync.
  Dirty pages trend toward zero before the snapshot is finished (but don't necessarily reach it).

---

Take a snapshot with --freeze-fs.
  
Check:
  Logs clearly indicate freeze and unfreeze.
  Logs do not indicate system sync.
  Loop running mount command clearly shows mount being created and destroyed.
  Loop running cat command clearly show cache being flushed.

---

Take a snapshot with --freeze-fs. Ctrl+C the instance manager container while the freeze is ongoing.

Check:
  Logs clearly indicate freeze.
  Logs indicate a timeout unfreezing. (This is because we automatically try to unfreeze during shutdown, but unfreeze is blocked by freeze. We return anyway.)
  Logs may indicate a later attempt to unfreeze. (This is because our freeze attempt finally fails when I/O errors return and then we attempt to undo it. It's really not particularly helpful or interesting, but it is expected.)
  The snapshot fails with an indication that freeze failed.
  The dd command gets unstuck (once I/O errors hit after the iSCSI timeout).
  The file system can be unmounted.

---

Take a snapshot with --freeze-fs. kill -9 the instance manager container while the freeze is ongoing.

Check:
  Logs clearly indicate freeze.
  The snapshot fails with an error that includes an EOF.
  The dd command gets unstuck (once I/O errors hit after the iSCSI timeout).
  The file system can be unmounted.

---

Take a snapshot with --freeze-fs. kill -9 the instance manager container BETWEEN FREEZE AND LOCK (need code modification).

Check:
  Logs clearly indicate freeze.
  The snapshot fails with an error that includes an EOF.
  After instance-manager is restarted:
    Logs clearly indicate unfreeze (from the instance-manager startup logic).
	The dd command gets unstuck.
	The file system can be unmounted.

---

Take a snapshot with --freeze-fs. Stop the engine using a call to instance-manager BETWEEN FREEZE AND LOCK (need code modification).

Check:
  Logs clearly indicate freeze.
  The snapshot fails with an error that includes an EOF.
  Logs clearly indicate unfreeze (from the shutdown frontend logic).
  The dd command gets unstuck.
  The file system can be unmounted.

---

Take a snapshot with --freeze-fs. kill -9 the engine process (but not instance-manager) BETWEEN FREEZE AND LOCK (need code modification).
  Logs clearly indicate freeze.
  The snapshot fails with an error that includes an EOF.
  To recover:
    Use a call to instance-manager to delete the engine and replica processes.
	Use a call to instance-manager to start the engine and replica processes.
  After this:
    Logs clearly indicate unfreeze (from the startup frontend logic).
    The dd command gets unstuck.
    The file system can be unmounted.  

  One concern here. Should we do this earlier? If not, we can never issue an unfreeze for an engine that can't get at least one backend running somewhere.

---

Take a snapshot with --freeze-fs. Then, quickly take another snapshot.
    Logs clearly indicate freeze and unfreeze for first snapshot.
	The snapshot fails with an indication that the file system is already being frozen

pkg/controller/control.go
```diff
@@ -1285,3 +1343,79 @@ func (c *Controller) GetLatestMetics() *enginerpc.Metrics {
...
+		// We must verify it is still save to freeze the filesystem. If the source mount was unmounted by someone else
+		// before we bind mounted, the bind mount could refer to the root filesystem.
+		var device string
+		device, _, err = mount.GetDeviceNameFromMount(mounter, freezePoint)
+		if err != nil || device != devicePath {
+			// This would likely only happen if an upper layer unmounted the volume. If that is the case, we are
+			// probably being torn down. Trying other sourcePoints will likely lead to a pointless race.
+			err = errors.Wrapf(err, "cannot verify mount point %v refers to volume", freezePoint)
+			return
+		}
```
PhanLe1010, line 1382: Should we directly find the filesystem on the root partition of the `devicePath` and mount it to `freezePoint` so we don't have to find and bind mount the existing mount points? And if there is no filesystem on the root partition, we can do no-op and fallback to sync.

By doing this, we don't have to worry that the upper layer is unmounting while we are doing bind mount and attempt to freeze

pkg/frontend/tgt/frontend.go
```diff
@@ -138,3 +150,28 @@ func (t *Tgt) Expand(size int64) error {
...
+		mounter := mount.New("")
+		mountPoints, err := mounter.List()
+		if err != nil {
+			logrus.WithError(err).Warn("Failed to list mount points while deciding whether or not unfreeze")
+		}
+		endpoint := t.Endpoint()
+		for _, mountPoint := range mountPoints {
+			if mountPoint.Device == endpoint {
+				util.AttemptUnfreezeFilesystem(mountPoint.Path, lhexec.NewExecutor(), false, logrus.New())
+				break
```
PhanLe1010, line 173: Should we prioritize the mount point we created first (the one at `freezePointDirectory  = "/var/lib/longhorn/freeze"`)?

pkg/frontend/tgt/frontend.go
```diff
@@ -70,10 +74,18 @@ func (t *Tgt) Startup(rwu types.ReaderWriterUnmapperAt) error {
...
+	// attempt to unfreeze it.
+	t.attemptUnfreezeFilesystem()
+
 	return nil
 }
 
 func (t *Tgt) Shutdown() error {
+	// If the engine is shutting down during a snapshot (in the preparation phase, before the snapshot operation obtains
+	// a lock) we may have left a frozen filesystem. This is a good opportunity to unfreeze it.
+	t.attemptUnfreezeFilesystem()
```
PhanLe1010, line 87: Should we put this logic into [func (d *LonghornDevice) Shutdown()](https://github.com/longhorn/longhorn-engine/blob/a807f0fd6bfd9c4700f2c19808038e87a2ab814e/vendor/github.com/longhorn/go-iscsi-helper/longhorndev/dev.go#L177) instead?

pkg/frontend/tgt/frontend.go
```diff
@@ -70,10 +74,18 @@ func (t *Tgt) Startup(rwu types.ReaderWriterUnmapperAt) error {
 
 	t.isUp = true
 
+	// If the engine failed during a snapshot, we may have left a frozen filesystem. This is a good opportunity to
+	// attempt to unfreeze it.
+	t.attemptUnfreezeFilesystem()
```
PhanLe1010, line 79: Should we put this logic into [func (d *LonghornDevice) startScsiDevice(](https://github.com/longhorn/longhorn-engine/blob/a807f0fd6bfd9c4700f2c19808038e87a2ab814e/vendor/github.com/longhorn/go-iscsi-helper/longhorndev/dev.go#L122)

pkg/controller/control.go
ejweber: I thought about this, and I don't feel too strongly either way. If a different implementation did as you suggested, it would probably be fine to me. Here are a couple reasons the bind mount approach might be better.

- If the file system is not mounted, freezing it is unnecessary. The bind mount approach only applies to a mounted file system.
- I'm not 100% sure of the behavior when directly mounting a file system that is already mounted. I tried it for ext4 and XFS volumes in my Rocky environment and it seemed to have the same effect as a bind mount. However, CSI drivers (including ours) always bind mount the staging path to the target path instead of doing a direct mount to the target path. It seems safer to me to follow suit.

ejweber:
Reworking some things in response to @PhanLe1010's feedback. Will reopen again when I am back to fairly confident.

pkg/frontend/tgt/frontend.go
ejweber: It probably wouldn't hurt. I'll add this functionality in an upcoming pass.

pkg/frontend/tgt/frontend.go
ejweber: While moving the logic into `go-iscsi-helper`, I realized that we should almost certainly execute it BEFORE `startFrontend`. If there are no available backends, we will never reach `startFrontend`. That's probably fine for the existing cleanup `startFrontend` might do (it'll strand a block device on the node), but it's NOT fine for the unfreeze case. We need to make sure the filesystem gets unfrozen, regardless of whether or not the engine can run.

As a result, I'm keeping the logic in `longhorn-engine`, but moving it to the `util` package so it can be executed earlier.

pkg/frontend/tgt/frontend.go
ejweber: See my reply to your similar comment about `startScsiDevice`. I will keep the logic in `longhorn-engine` because of the new location we will execute it from.

pkg/controller/control.go
```diff
@@ -185,25 +190,78 @@ func (c *Controller) addReplica(address string, snapshotRequired bool, mode type
...
+	}()
 
-	if name == "" {
-		name = lhutils.UUID()
+	// We create a bind mount of one of the discovered mount points (it does not matter which one) inside the container
+	// mount namespace (assuming longhorn-engine is running in a container). The mount point cannot be unmounted except
+	// by us, so we can be confident it is safe to freeze. This approach is preferable to locking some source mount
+	// point with an open file descriptor because it cannot be circumvented by a lazy unmount at the source.
+	if shouldFreeze && endpoint != "" {
+		if !c.snapshotFreezeLock.TryLock() {
```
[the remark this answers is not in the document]
ejweber: We discussed this offline. While there are some pros and cons to both behaviors, for now we will leave it as a `TryLock`. It is mainly because `Lock` may be stuck forever in some unknown scenario, but the `Snapshot` client will give up waiting. In that case, the user will not know why the snapshot failed and the goroutine will exist indefinitely. It may be better to just tell the user the snapshot failed immediately and allow them to retry. 

FWIW, the argument for `Lock` is quite valid as well. We can almost certainly get the lock if we are patient, so we don't really need to return failure. If we notice that the chosen behavior is causing problems, we can reevaluate.

ejweber:
Moved back to ready for review @PhanLe1010. Sorry for the delay. I responded to all of your comments and made changes with respect to some of them.

I also fixed an issue Codefactor had with my integration test and ensured we actually cleaned up the directories within the instance-manager container we bind mounted to (in case a long running instance-manager used too many!).

PhanLe1010:
> ensured we actually cleaned up the directories within the instance-manager container we bind mounted to (in case a long running instance-manager used too many!)

Could you point me to this? Thank you

ejweber:
Sorry for the confusion. [This commit](https://github.com/longhorn/longhorn-engine/pull/1083/commits/d57877e1f0252cd07fd10ecc276b26ce3944b0cc) uses CleanupMountPoint in two places:

- One where we previously just used Unmount (so did not clean up the directory).
- One where we previously did not Unmount at all. (This was an oversight that could have led to mounts for shut down engines accumulating over time.)

PhanLe1010 (dismissed):
LGTM

Thank you for the great implementation!

Dockerfile.dapper
```diff
@@ -12,7 +12,10 @@ ENV PATH /go/bin:$PATH
 ENV DAPPER_DOCKER_SOCKET true
 ENV DAPPER_ENV TAG REPO DRONE_REPO DRONE_PULL_REQUEST DRONE_COMMIT_REF SKIP_TASKS
 ENV DAPPER_OUTPUT bin coverage.out
-ENV DAPPER_RUN_ARGS --privileged --tmpfs /go/src/github.com/longhorn/longhorn-engine/integration/.venv:exec --tmpfs /go/src/github.com/longhorn/longhorn-engine/integration/.tox:exec -v /dev:/host/dev -v /proc:/host/proc
+# For filesystem freeze tests, our container must be able to see the filesystem mounted in the host mount namespace.
+# Usually, instance-manager runs with the equivalent of "-v /:/host". For integration testing, use "-v /tmp:host/tmp"
```
james-munson, line 16: Should be "-v /tmp:/host/tmp" in the comment (note missing "/").

pkg/controller/control.go
```diff
@@ -1287,3 +1357,69 @@ func (c *Controller) GetLatestMetics() *enginerpc.Metrics {
...
+		}
+		log.Debugf("Mounting filesystem to %v", freezePoint)
+		if err = attemptBindMountFilesystem(sourcePoint.Path, freezePoint, mounter); err != nil {
+			// There's no particular reason to think we will be successful with a different sourcePoint.
+			err = errors.Wrapf(err, "failed to bind mount filesystem from %v to %v", sourcePoint, freezePoint)
+			return
+		}
+		mounted = true
+
+		// We must verify it is still save to freeze the filesystem. If the source mount was unmounted by someone else
```
james-munson, line 1387: "save" -> "safe"

james-munson (approved):
Just a couple of nits in comments, no issues.  LGTM

ejweber:
Since https://github.com/longhorn/types/pull/8 merged and I have two approving reviews, I will remove my debug commits and rebase in preparation for merge soon.

Dockerfile.dapper
```diff
@@ -115,7 +115,7 @@ RUN cd integration && \
 
 # Build longhorn-instance-manager for integration testing
 RUN cd /go/src/github.com/longhorn && \
-    git clone https://github.com/longhorn/longhorn-instance-manager.git && \
+    git clone https://github.com/ejweber/longhorn-instance-manager.git -b 2187-fsfreeze && \
```
shuo-wu, line 121: Remember to remove this commit when the PR is ready to go

pkg/frontend/tgt/frontend.go
```diff
@@ -75,16 +75,21 @@ func (t *Tgt) Startup(rwu types.ReaderWriterUnmapperAt) error {
...
 
 	return nil
 }
 
 func (t *Tgt) Shutdown() error {
 	// If the engine is shutting down during a snapshot (in the preparation phase, before the snapshot operation obtains
 	// a lock) we may have left a frozen filesystem. This is a good opportunity to unfreeze it.
-	t.attemptUnfreezeFilesystem()
+	if err := t.attemptUnfreezeFilesystem(); err != nil {
+		logrus.WithError(err).Errorf("Failed to clean up frozen file system during shutdown")
```
shuo-wu, line 81: `"..., will continue the tgt shutdown"`

pkg/util/fsfreeze.go
```diff
@@ -23,14 +28,25 @@ const (
...
+// GetFreezePointFromVolumeName returns the absolute path to the canonical location we will try to mount a filesystem to
+// before freezeing it.
+func GetFreezePointFromVolumeName(volumeName string) string {
+	return filepath.Join(freezePointDirectory, volumeName)
+}
+
+// GetDevicePathFromVolumeName mirrors longhorndev.getDev. It returns the device path that go-iscsi-helper will use.
+func GetDevicePathFromVolumeName(volumeName string) string {
+	return filepath.Join(longhorndev.DevPath, volumeName)
+}
```
shuo-wu: Should these functions return a non-empty value (`/var/lib/longhorn/freeze/`) when the input is empty?

pkg/controller/control.go
```diff
@@ -1287,3 +1329,79 @@ func (c *Controller) GetLatestMetics() *enginerpc.Metrics {
...
+		var device string
+		device, _, err = mount.GetDeviceNameFromMount(mounter, freezePoint)
+		if err != nil || device != devicePath {
+			// This would likely only happen if an upper layer unmounted the volume. If that is the case, we are
+			// probably being torn down. Trying other sourcePoints will likely lead to a pointless race.
+			err = errors.Wrapf(err, "cannot verify mount point %v refers to volume", freezePoint)
+			return
+		}
+
+		log.Infof("Freezing file system mounted at %v", freezePoint)
```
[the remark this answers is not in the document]
ejweber: I'm assuming we agree that this does not matter (since unmounting the source mount will not unmount the bind mount). I'm adding a comment to this effect.

pkg/controller/control.go
```diff
@@ -1287,3 +1329,79 @@ func (c *Controller) GetLatestMetics() *enginerpc.Metrics {
...
+	}
+
+	return mounter.Mount(sourcePoint, mountPoint, "", []string{"bind"})
+}
+
+// attemptUnmountFileSystem attempts to unmount the file system mounted at mountPoint. Returns success or failure.
+func attemptUnmountFileSystem(mountPoint string, mounter mount.Interface, log logrus.FieldLogger) bool {
+	log.Debugf("Unmounting file system mounted at %v", mountPoint)
+	if err := mounter.Unmount(mountPoint); err != nil {
+		log.WithError(err).Warnf("Failed to unmount file system mounted at %v", mountPoint)
```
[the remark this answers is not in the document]
ejweber: It looks like this comment is on outdated code from a previous commit. We already moved to using `mount.CleanUpMountPoint` instead of `mounter.Unmount` at all locations.

pkg/util/fsfreeze.go
ejweber: If I understand you correctly, you are saying it is safer if we return an empty value instead of whatever the base of the join is? Makes sense to me.

ejweber:
Added https://github.com/longhorn/longhorn-engine/pull/1083/commits/66fa2e6855196c400b831e36c0da8174b9e86181 for @shuo-wu's comments, removed debugging helpers, and pointed back to upstream types. Rebase is a mess given the number of commits / time in development and the dependency changes since I started, so we can use squash instead. Ready to go pending @shuo-wu's approval.

shuo-wu (approved):

shuo-wu:
After resolving the rebase and conflicts, it's good to go

PR #1085, opened 2024-04-25
Fix some v1 data path bottlenecks

PhanLe1010:
1. Improve revision counter logic
2. Increase the connection from engine to replica

Please see the improvement test result in the issue description at longhorn/longhorn#8436

longhorn/longhorn#8436

TODO:
- [ ] Do we need to make the number of connections from engine to replica customizable so user can adjust it for their particular env?

### DIFF
--- a/pkg/backend/remote/remote.go
+++ b/pkg/backend/remote/remote.go
@@ -22,7 +22,8 @@ import (
 )
 
 const (
-	PingInterval = 2 * time.Second
+	PingInterval        = 2 * time.Second
+	NumberOfConnections = 2
 )
 
 func New() types.BackendFactory {
@@ -406,12 +407,16 @@ func (rf *Factory) Create(volumeName, address string, dataServerProtocol types.D
 		return nil, fmt.Errorf("replica must be closed, cannot add in state: %s", replica.State)
 	}
 
-	conn, err := connect(dataServerProtocol, dataAddress)
-	if err != nil {
-		return nil, err
+	var conns []net.Conn
+	for i := 0; i < NumberOfConnections; i++ {
+		conn, err := connect(dataServerProtocol, dataAddress)
+		if err != nil {
+			return nil, err
+		}
+		conns = append(conns, conn)
 	}
 
-	dataConnClient := dataconn.NewClient(conn, engineToReplicaTimeout)
+	dataConnClient := dataconn.NewClient(conns, engineToReplicaTimeout)
 	r.ReaderWriterUnmapperAt = dataConnClient
 
 	if err := r.open(); err != nil {
--- a/pkg/replica/revision_counter.go
+++ b/pkg/replica/revision_counter.go
@@ -36,14 +36,21 @@ func (r *Replica) readRevisionCounter() (int64, error) {
 	return counter, nil
 }
 
+var revisionCounterBuf = sparse.AllocateAligned(revisionBlockSize)
+
 func (r *Replica) writeRevisionCounter(counter int64) error {
 	if r.revisionFile == nil {
 		return fmt.Errorf("BUG: revision file wasn't initialized")
 	}
 
-	buf := make([]byte, revisionBlockSize)
-	copy(buf, []byte(strconv.FormatInt(counter, 10)))
-	_, err := r.revisionFile.WriteAt(buf, 0)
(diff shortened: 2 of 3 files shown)


PhanLe1010:
> Do we need to make the number of connections from engine to replica customizable so user can adjust it for their particular env?

I think I agree with @ejweber that we should keep it simple and hardcode it to 2 for now. Will monitor and adjust later when needed.

pkg/replica/revision_counter.go
```diff
@@ -36,14 +36,21 @@ func (r *Replica) readRevisionCounter() (int64, error) {
...
 		return fmt.Errorf("BUG: revision file wasn't initialized")
 	}
 
-	buf := make([]byte, revisionBlockSize)
-	copy(buf, []byte(strconv.FormatInt(counter, 10)))
-	_, err := r.revisionFile.WriteAt(buf, 0)
+	counterBytes := []byte(strconv.FormatInt(counter, 10))
+	copy(revisionCounterBuf, counterBytes)
+	// Need to clear the part of the slice left over from the previous function call.
+	// The maximum length of the string representation of an int64 is 20 bytes so only need to clear upto 20 bytes
```
ejweber, line 49: We discussed this yesterday. I was of the opinion that we should maintain two versions of this function:

- One that allocates its own buffer to be called generically by anyone. (This is to ensure safety in case a generic caller might decrement the revisionCounter.)
- One that uses the pre-allocated buffer to be called only by increaseRevisionCounter. (I think the clearing logic here is not necessary for that one.)

When we were discussing it, @Phan mentioned that the code is just as fast with this clearing logic as it is without, so it seems a reasonable safety measure to leave it in for all callers.

ejweber (commented):
LGTM in general. One question I want to make sure we know the answer to before merging.

pkg/replica/revision_counter.go
PhanLe1010: Thank you for the recall! Yes, it doesn't seem to slow down so I think it is fine to have the cleaning logic in all callers unless we have additional concerns.

pkg/backend/remote/remote.go
```diff
@@ -406,12 +407,16 @@ func (rf *Factory) Create(volumeName, address string, dataServerProtocol types.D
 		return nil, fmt.Errorf("replica must be closed, cannot add in state: %s", replica.State)
 	}
 
-	conn, err := connect(dataServerProtocol, dataAddress)
-	if err != nil {
-		return nil, err
+	var conns []net.Conn
+	for i := 0; i < NumberOfConnections; i++ {
+		conn, err := connect(dataServerProtocol, dataAddress)
```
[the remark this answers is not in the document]
PhanLe1010: It seems working fine. 

When using the new implementation, I see that Longhorn opens multiple unix socket connections:
```
instance-manager-1df2fc71f1f32eb7b92b25f4a29f709a:/ # ss -x
Netid         State          Recv-Q          Send-Q                                                             Local Address:Port                       Peer Address:Port             Process         
u_str         ESTAB          0               0                                                                              * 38022507                              * 38027347                         
u_str         ESTAB          0               0                        /host/var/lib/longhorn/unix-domain-socket/testvol4.sock 38027347                              * 38022507                         
u_str         ESTAB          0               0                        /host/var/lib/longhorn/unix-domain-socket/testvol4.sock 38027346                              * 38022506            
```

Performance with master-head + strict local volume
```
IOPS (Read/Write)
        Random:          52,871 / 56,442
    Sequential:          36,167 / 76,660

Bandwidth in KiB/sec (Read/Write)
        Random:        326,076 / 427,580
    Sequential:        456,475 / 447,457
                                        

Latency in ns (Read/Write)
        Random:        312,905 / 170,790
    Sequential:        212,590 / 171,073

```

Performance with new implementation + strict local volume
```
IOPS (Read/Write)
        Random:          54,161 / 56,759
    Sequential:          37,150 / 78,013

Bandwidth in KiB/sec (Read/Write)
        Random:        319,321 / 440,061
    Sequential:        451,429 / 428,603
                                        

Latency in ns (Read/Write)
        Random:        308,744 / 170,879
    Sequential:        210,754 / 165,663
```

shuo-wu (approved):

PhanLe1010:

PR #1097, opened 2024-05-03
Improve replica revision counter

shuo-wu:

https://github.com/longhorn/longhorn/issues/8436

1. Remove unused replica HTTP service
2. Replace revision counter mutex lock with a go channel
3. Use a separate goroutine to increase the revision counter during write

This PR implements idea 4 and 5 mentioned in [this comment](https://github.com/longhorn/longhorn/issues/6600#issuecomment-2092210119)


pkg/replica/revision_counter.go
```diff
@@ -92,9 +90,30 @@ func (r *Replica) initRevisionCounter() error {
 	}
 	// Don't use r.revisionCache directly
 	// r.revisionCache is an internal cache, to avoid read from disk
-	// every time when counter needs to be updated.
-	// And it's protected by revisionLock
-	r.revisionCache = counter
+	atomic.SwapInt64(&r.revisionCache, counter)
```
ejweber, line 93:
```suggestion
	r.revisionCache.Swap(counter) // Or similar.
```

The Golang docs suggest replacing `SwapInt64` and similar functions with `Int64`.Swap.

https://pkg.go.dev/sync/atomic#SwapInt64

This means we need to change the type of r.revisionCounter from `int64` to `atomic.Int64`, which may have other ramifications (that are hopefully positive).

The idea seems to be that the new atomic wrappers prevent developers from accidentally performing non-atomic operations on the code elsewhere.

https://medium.com/@deckarep/the-go-1-19-atomic-wrappers-and-why-to-use-them-ae14c1177ad8

IMO, in a "post Go v1.19" world, if we are switching from mutex protection to atomic protection, we should follow this convention.

ejweber: If we make this change, other similar functions will need to change in this PR. I won't comment individually on them.

app/cmd/replica.go
ejweber, line 1: Can you help clarify the need for this new context to help my understanding? I guess if we don't have it, the new revision counter goroutine will not return when a fatal issue occurs elsewhere. Does that stop the replica from terminating?

pkg/replica/revision_counter.go
ejweber, line 93: Nit. We lost the following line from the comment, but it is still relevant.

```
// every time when counter needs to be updated.
```

ejweber (commented):
LGTM in principle. One significant suggestion, one nit, and one request for clarification.

app/cmd/replica.go
shuo-wu: I think it's better to have a graceful way to stop such a for loop in the goroutine. Yes, theoretically, the fatal error exit should stop everything for the engine process, including this goroutine.

ejweber (approved):
The changes LGTM. Thanks!

PhanLe1010 (approved):
LGTM in gerneral. Quick question: we are removing the `/pkg/replica/rest`, are we no longer need it in any testing tool?

shuo-wu:
> Quick question: we are removing the /pkg/replica/rest, are we no longer need it in any testing tool?

No, we don't need it anymore.

PR #1125, opened 2024-06-11
Make long gRPC timeout configurable

PhanLe1010:
longhorn/longhorn#2765

We need to update go.mod to have this change before merging https://github.com/longhorn/types/pull/14


PhanLe1010:
Note: we need to update go.mod to have this change before merging https://github.com/longhorn/types/pull/14

app/cmd/add_replica.go
```diff
@@ -92,11 +92,12 @@ func addReplica(c *cli.Context) error {
 
 	fastSync := c.Bool("fast-sync")
 	fileSyncHTTPClientTimeout := c.Int("file-sync-http-client-timeout")
+	grpcTimeoutSeconds := c.Int64("grpc-timeout-seconds")
```
innobead, line 95: Need to add this arg to the CLI common flags.

app/cmd/snapshot.go
```diff
@@ -469,7 +469,7 @@ func cloneSnapshot(c *cli.Context) error {
 	defer fromControllerClient.Close()
 
 	if err := sync.CloneSnapshot(controllerClient, fromControllerClient, volumeName, fromVolumeName,
-		snapshotName, exportBackingImageIfExist, fileSyncHTTPClientTimeout); err != nil {
+		snapshotName, exportBackingImageIfExist, fileSyncHTTPClientTimeout, 0); err != nil {
```
innobead, line 472: Why not get the arg from `grpc-timeout-seconds` which is also needed to add to the command flag.

innobead (changes requested):

app/cmd/add_replica.go
PhanLe1010: Thank you. Fixed

app/cmd/snapshot.go
PhanLe1010: Thank you. Fixed

PhanLe1010: Btw, for the context, I think the CLI is only used to debugging now since we will use gRPC sorely for communication between longhorn-manager -> instance-manager -> engine -> replica

innobead: yeah, exactly.

innobead (approved):

app/cmd/add_replica.go
```diff
@@ -41,6 +41,12 @@ func AddReplicaCmd() cli.Command {
 				Value:    5,
 				Usage:    "HTTP client timeout for replica file sync server",
 			},
+			cli.IntFlag{
+				Name:     "grpc-timeout-seconds",
```
derekbit, line 45: Should we make the name more specific? For example, `rebuilding-grpc-timeout-seconds`?

ejweber (approved):
LGTM.

app/cmd/add_replica.go
PhanLe1010: This is inside the AddReplicaCmd so I think it is only for rebuilding. WDYT?

Btw, we also use it inside snapshotClone command so I feel that rebuilding-grpc-timeout-seconds doesn't seem cover that case. WDYT?

derekbit: Got it. Then, using `grpc-timeout-seconds` is fine.

derekbit (approved):

PR #1132, opened 2024-06-17
ci: use golang bci instead of manual download

innobead:

Issue #


innobead:

PR #1137, opened 2024-06-25
ci: build and publish image with sbom attestation

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/5099

build and publish image with sbom attestation


innobead (approved):

yangchiu:
@mergify backport v1.6.x

PR #1174, opened 2024-08-06
fix(replica): best-effort to delete old head disk file

derekbit:

Issue Longhorn/longhorn#9168

The deletion of old head disk file is a best-effort task, and the failure should not hinder the following tasks.

The regression is caused by the [commit](https://github.com/longhorn/longhorn-engine/commit/737bf77b6e0f01e68dbb340e4b3421bb0c7417b6#diff-b54af9ffed94a317525bc7609415261b623020d2aa410c124e9fe3b871d2d2bdR918).

### DIFF
--- a/pkg/replica/replica.go
+++ b/pkg/replica/replica.go
@@ -920,8 +920,9 @@ func (r *Replica) createDisk(name string, userCreated bool, created string, labe
 	rollbackFuncList := []func() error{}
 	defer func() {
 		if err == nil {
-			err = r.rmDisk(oldHead)
-			logrus.WithError(err).Errorf("Failed to remove old head %v", oldHead)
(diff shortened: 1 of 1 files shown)


derekbit:
@mergify backport v1.7.x

ChanYiLin (approved):
LGTM!

innobead:
Does this mean there would be chances to have orphaned old head snapshot img?

derekbit:
> Does this mean there would be chances to have orphaned old head snapshot img?

Yes, need a manual intervention for the deletion.

derekbit:
@mergify backport v1.6.x

derekbit:
> @mergify backport v1.6.x

Backporting to v1.6.x is necessary, because the bug is also in v1.6.x,

PR #1211, opened 2024-09-04
fix: skip updating the status if the backup has already completed or failed

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/9168

To prevent race conditions where some goroutines might update the status to 'In Progress' after the backup has already completed or failed, we should skip updating the status in such cases.


derekbit (commented):
LGTM

derekbit (approved):

PR #1272, opened 2024-10-28
fix(logging): demote some Info logging to Debug

james-munson:

Part of https://github.com/longhorn/longhorn/issues/6702

Demote some noisy Info-level logging to Debug.  Routine snapshotting can flood logs with Info-level output.

### DIFF
--- a/pkg/backend/remote/remote.go
+++ b/pkg/backend/remote/remote.go
@@ -96,7 +96,7 @@ func (r *Remote) open() error {
 }
 
 func (r *Remote) Snapshot(name string, userCreated bool, created string, labels map[string]string) error {
-	logrus.Infof("Starting to snapshot: %s %s UserCreated %v Created at %v, Labels %v",
+	logrus.Debugf("Starting to snapshot: %s %s UserCreated %v Created at %v, Labels %v",
 		r.name, name, userCreated, created, labels)
 	conn, err := grpc.NewClient(r.replicaServiceURL, grpc.WithTransportCredentials(insecure.NewCredentials()),
 		interceptor.WithIdentityValidationClientInterceptor(r.volumeName, ""))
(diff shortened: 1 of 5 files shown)


PhanLe1010 (approved):
LGTM

Same comment as https://github.com/longhorn/longhorn-instance-manager/pull/700#pullrequestreview-2402786862

derekbit:
These log messages only print once for each operation and won't result in flooding the logs, so I think we should keep them at the INFO level.

derekbit:
> These log messages only print once for each operation and won't result in flooding the logs, so I think we should keep them at the INFO level.

WDYT? @PhanLe1010 @innobead @james-munson

james-munson:
As noted in the related https://github.com/longhorn/longhorn-instance-manager/pull/700, these logs were picked because they repeated more than 1000 times each over a two-week period in a recent escalation support bundle.  Granted, that was in an environment doing a good amount of snapshots.

derekbit:
> As noted in the related https://github.com/longhorn/longhorn-instance-manager/pull/700, these logs were picked because they repeated more than 1000 times each over a two-week period in a recent escalation support bundle. Granted, that was in an environment doing a good amount of snapshots.

I think it is not a good reason to demote the log level of this one-time and important log messages.

Do you know which pod log is the user complaining about? I remember longhorn-manager pods rather than longhorn-instance-manager pods?

james-munson:
> Do you know which pod log is the user complaining about? I remember longhorn-manager pods rather than longhorn-instance-manager pods?

There were 138 MB of longhorn-manager logs over 4 nodes, and 108 MB of instance-manager logs, so 246 MB in total.  The breakdown is in https://github.com/longhorn/longhorn/issues/6702#issuecomment-2436512562

If the consensus is that these are too likely to be needed, I will close this PR as well, and focus on propagating the global log level setting.

PR #1344, opened 2024-12-16
chore(dockerfile): clone and build correct instance-manager codes for v1.6.x

derekbit:

Issue #


innobead (approved):

innobead:
or `git clone -b` inline?

derekbit:
> or `git clone -b` inline?

Updated.

innobead (approved):

PR #1403, opened 2025-02-06
feat: new makefile target for secure image build

COLDTURNIP:

Supports security build.

Prepare for potential secure release

The secrets provided by EIO MUST NOT present in any visible output.

Guidance on achieving SLSA Level 3

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,48 @@ on:
     - v*
   pull_request:
 jobs:
+  build_info:
+    name: Collect build info
+    runs-on: ubuntu-latest
+    outputs:
+      version_major: ${{ steps.build_info.outputs.version_major }}
+      version_minor: ${{ steps.build_info.outputs.version_minor }}
+      version_patch: ${{ steps.build_info.outputs.version_patch }}
+      image_tag: ${{ steps.build_info.outputs.image_tag }}
+
+    steps:
+    - id: build_info
+      name: Declare build info
+      run: |
+        version_major=''
+        version_minor=''
+        version_patch=''
+        image_tag=''
+
+        branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,10 +1,11 @@
-FROM registry.suse.com/bci/golang:1.23
+FROM registry.suse.com/bci/golang:1.23 AS builder_base
```
COLDTURNIP, line 1: The `builder_base` stage is the former `Dockerfile.dapper`, and the rest stages are the content of `package/Dockerfile` originally.

- `builder_base` stage is the target stage of dapper for local build
- `builder_base` stage is also the base of `engine_builder` for self-contained building
- Due to dapper issue, we need to move Dockerfile from package to repo root: [FATA[0000] pattern contains path separator](https://github.com/rancher/dapper/issues/92)

Dockerfile
derekbit, line 1: Why need to rename the file?

package/Dockerfile
```diff
@@ -1,75 +0,0 @@
-FROM registry.suse.com/bci/bci-base:15.6 AS builder
```
derekbit, line 1: Dockerfile is deleted?

Dockerfile
COLDTURNIP: The `builder_base` stage is the former `Dockerfile.dapper`, and the rest stages are the content of `package/Dockerfile` originally.

- `builder_base` stage is the target stage of dapper for local build
- `builder_base` stage is also the base of `engine_builder` for self-contained building
- Due to dapper issue, we need to move Dockerfile from package to repo root: [FATA[0000] pattern contains path separator](https://github.com/rancher/dapper/issues/92)

COLDTURNIP: Per discussion, will revert back and extract build env setup as a shared script for dapper and package/Dockerfile.

derekbit (approved):

PR #1404, opened 2025-02-07
fix: not allow more than 250 snapshots including all kind of snapshots

derekbit:

Issue Longhorn/longhorn#10308

Not allow creation of more than 250 snapshots including all kind of snapshots (volume-head, backing image, removed snapshot, ...) because we use 8-bit map for snapshot indexing.

### DIFF
--- a/pkg/replica/replica.go
+++ b/pkg/replica/replica.go
@@ -1407,17 +1407,11 @@ func (r *Replica) GetSnapshotCountUsage() int {
 
 func (r *Replica) getSnapshotCountUsage() int {
 	var (
-		backingDiskName string
-		snapshotCount   int
+		snapshotCount int
 	)
-	// index 0 is nil or backing file
-	if r.activeDiskData[0] != nil {
-		backingDiskName = r.activeDiskData[0].Name
-	}
 
 	for _, disk := range r.diskData {
-		// exclude volume head, backing disk, and removed disks
-		if disk == nil || disk.Removed || disk.Name == backingDiskName || disk.Name == r.info.Head {
+		if disk == nil {
 			continue
 		}
 		snapshotCount++


derekbit:
@mergify backport v1.8.x v1.7.x v1.6.x

innobead (approved):

innobead:
Need to review if the documentation needs to be updated accordingly.

innobead (approved):

PhanLe1010 (changes requested):
Reviewing. Putting a change request here to hold the merge as there is a concern about behavior change

PhanLe1010:
Proposing an alternative solution https://github.com/longhorn/longhorn-engine/pull/1405 because this PR has behevior change which might cause confusion/regression in the following case:
1. User create a volume with backing image
1. User set maximum snapshot count for the volume to be 2
1. Longhorn would not allow user to take any snapshot because there are already 2 (backing image + volume-head). This might be a behavior change, or confusion, or regression

cc @derekbit @innobead

derekbit:
> Proposing an alternative solution #1405 because this PR has behavior change which might cause confusion/regression in the following case:
> 
> 1. User create a volume with backing image
> 2. User set maximum snapshot count for the volume to be 2
> 3. Longhorn would not allow user to take any snapshot because there are already 2 (backing image + volume-head). This might be a behavior change, or confusion, or regression
> 
> cc @derekbit @innobead

@PhanLe1010 
The difference of the solutions is how we define "snapshot" for users: can a backing image or a volume-head be regarded as a snapshot for user?  So, volume-head and backing-image should not count in https://github.com/longhorn/longhorn-engine/pull/1405 before `countTotal` (len(r.diskData)) >= types.MaximumTotalSnapshotCount. Is it correct?

PR #1432, opened 2025-02-21
fix: build failed

PhanLe1010:
When `./bin/longhorn-instance-manager` file does exist, `[[ ! -x ./bin/longhorn-instance-manager ]]` evaluates to `false` and return exit code 1. Since `set -e` is enabled, the script immediately exits on this failure.

Always copy and overwrite the `./bin/longhorn-instance-manager` if it exists


james-munson (approved):
LGTM.  Also tested it out with success.

derekbit (approved):
LGTM

derekbit:
@mergify backport v1.8.x

PR #1435, opened 2025-02-25
fix: rebuilding gets stuck if verification fails and longhorn-manager crashes

shuo-wu:

Issue Longhorn/longhorn#8589

https://github.com/longhorn/longhorn/issues/8589#issuecomment-2679915554


derekbit (approved):
LGTM

ChanYiLin (approved):
LGTM
I feel like I have also encountered this kind of error "not" handling issue for other bugs before.
This can be kind of the debug idea we should come up and check in the future.

derekbit:
@mergify backport v1.8.x

derekbit:
@mergify backport v1.7.x

PR #1438, opened 2025-02-26
feat(ci): enable CI validation for arm64 builds

COLDTURNIP:
Enable whole CI steps while building binary for arm64, include:

- validate
- sync-grpc-py
- test
- integration-test


COLDTURNIP:
Drop this since there's env issue for integration test on arm64.

PR #1533, opened 2025-07-18
feat(backup): configurable bkup block size

COLDTURNIP:

Issue longhorn/longhorn#5215

Implement configurable backup block size.

1. While creating a new backup, the block size is configured in the backup parameter. If the block size is not set, it fallback to default 2MB by default.
2. The backup block size will be recorded into the backup configuration file.
3. While restoring a backup, the block size is read from the backup configuration file.
4. For incremental backups, the block size must be identical between backups.

As the implementation of interface `backupstore.DeltaBlockBackupOperations`, the `backup` should be able to consume the block size from caller.

This PR depends on the following PRs

- longhorn/go-common-libs#122
- longhorn/backupstore#250

The dependency commits must be resolved before merging.


app/cmd/backup.go
```diff
@@ -286,6 +295,13 @@ func createBackup(c *cli.Context) error {
 	url := c.GlobalString("url")
 	volumeName := c.GlobalString("volume-name")
 	engineInstanceName := c.GlobalString("engine-instance-name")
+
+	if quantity, parseErr := resource.ParseQuantity(c.String("backup-block-size")); parseErr != nil {
+		return fmt.Errorf("cannot parse backup block size: %w", parseErr)
```
derekbit:
```suggestion
		return errors.Wrap(err, "failed to parse backup block size")
```

app/cmd/backup.go
```diff
@@ -80,6 +84,11 @@ func BackupCreateCmd() cli.Command {
 				Name:  "storage-class-name",
 				Usage: "Storage class name of the pv binding with the volume",
 			},
+			cli.StringFlag{
+				Name:  "backup-block-size",
+				Usage: "Backup block size quantity",
+				Value: backupstore.DEFAULT_BLOCK_SIZE_QUANTITY,
```
derekbit: Can we rename it to `DEFAULT_BLOCK_SIZE_IN_MB`?

```suggestion
				Value: backupstore.DEFAULT_BLOCK_SIZE_QUANTITY,
```

COLDTURNIP: Do you mean to accept `2` (MB) instead of `2Mi` as the option value here?

derekbit: Yes, might be more clear?

COLDTURNIP: Will change for this, thanks.

COLDTURNIP: Refactor the backupstore needed here: longhorn/backupstore/pull/251

app/cmd/backup.go
```diff
@@ -286,6 +295,15 @@ func createBackup(c *cli.Context) error {
 	url := c.GlobalString("url")
 	volumeName := c.GlobalString("volume-name")
 	engineInstanceName := c.GlobalString("engine-instance-name")
+
+	backupBlockSizeMB := c.Int64("backup-block-size")
+	if backupBlockSizeMB <= 0 {
+		return errors.New(fmt.Sprintf("invalid backup block size: %v", backupBlockSizeMB))
+	} else {
+		quantity := resource.NewQuantity(backupBlockSizeMB*1024*1024, resource.BinarySI)
+		parameters[lhbackup.LonghornBackupParameterBackupBlockSize] = strconv.FormatInt(quantity.Value(), 10)
```
derekbit, line 304: Just curious. Do we need the quantity here?

COLDTURNIP: Personally prefer the quantity because of readability. The parameters will be saved into the backup's metadata JSON file, and could be checked without any tool.

pkg/replica/backup.go
```diff
@@ -202,11 +202,11 @@ func (rb *BackupStatus) CompareSnapshot(snapID, compareSnapID, volumeID string)
 		if val <= byte(from) && val > byte(to) {
 			offset := int64(i) * rb.replica.volume.sectorSize
 			// align
-			offset -= (offset % backupBlockSize)
+			offset -= offset % blockSize
```
derekbit: nit:
I'm fine with `offset -= (offset % backupBlockSize)` because it has better readibility although the `%` operator has higher precedence than `-=`.

pkg/sync/rpc/server.go
```diff
@@ -155,6 +157,7 @@ func NewSyncAgentServer(startPort, endPort int, replicaAddress, volumeName, inst
 	server := grpc.NewServer(interceptor.WithIdentityValidationReplicaServerInterceptor(volumeName, instanceName))
 	enginerpc.RegisterSyncAgentServiceServer(server, sas)
 	reflection.Register(server)
+	profilerrpc.RegisterProfilerServer(server, profiler.NewServer(volumeName+"-sync-agent"))
```
derekbit: What's the purpose of this?

COLDTURNIP: Exposing the RPC interface. to make the sync agent profilable without re-compile the engine. The profiling is disabled by default, and user can enable/disable pprof by:

```
longhorn --url http://localhost:<sync_agent_port> profiler enable --port <profiler_port>
```

This would be useful to anyone interested about the runtime resource consumption for backup syncing.

derekbit: Can you put it in another PR or another commit? Also add this to https://longhorn.io/docs/1.9.1/troubleshoot/

derekbit: Sound good! Can you create another IMPROVEMENT ticket and PR for this improvement? Also add this to https://longhorn.io/docs/1.9.1/troubleshoot/. Thank you.

COLDTURNIP: Will remove this here and send another PR.

COLDTURNIP: Moved this part to another ticket: longhorn/longhorn#11386

derekbit (approved):
LGTM

COLDTURNIP:
@mergify backport v1.8.x v1.9.x

PR #1537, opened 2025-07-28
feat(backup): expose profiler interface for sync agent

COLDTURNIP:

Issue longhorn/longhorn#11386

Similar to [the engine & replica](https://github.com/longhorn/longhorn-engine/blob/master/pkg/controller/rpc/server.go#L55), expose the profiling control interface, and user can enable/disable pprof server by:

```bash
longhorn --url http://localhost:<sync_agent_port> profiler enable --port <profiler_port>
```

### DIFF
--- a/pkg/sync/rpc/server.go
+++ b/pkg/sync/rpc/server.go
@@ -14,9 +14,7 @@ import (
 
 	"github.com/gofrs/flock"
 	"github.com/longhorn/backupstore"
-	butil "github.com/longhorn/backupstore/util"
 	"github.com/longhorn/sparse-tools/sparse"
-	sparserest "github.com/longhorn/sparse-tools/sparse/rest"
 	"github.com/longhorn/types/pkg/generated/enginerpc"
 	"github.com/moby/moby/pkg/reexec"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


pkg/sync/rpc/server.go
```diff
@@ -28,6 +28,8 @@ import (
 	"google.golang.org/protobuf/types/known/emptypb"
 
 	lhio "github.com/longhorn/go-common-libs/io"
+	"github.com/longhorn/go-common-libs/profiler"
```
davidcheng0922, line 31: maybe need to be reorganized to follow the import order?
https://github.com/longhorn/longhorn/wiki/coding-convention#organize-package-imports

pkg/sync/rpc/server.go
```diff
@@ -155,6 +160,7 @@ func NewSyncAgentServer(startPort, endPort int, replicaAddress, volumeName, inst
 	server := grpc.NewServer(interceptor.WithIdentityValidationReplicaServerInterceptor(volumeName, instanceName))
 	enginerpc.RegisterSyncAgentServiceServer(server, sas)
 	reflection.Register(server)
+	profilerrpc.RegisterProfilerServer(server, profiler.NewServer(volumeName+"-sync-agent"))
```
derekbit, line 165: Should we add the suffix `-sync-agent`?

COLDTURNIP: In fact, [the name of the server is not used in the implementation](https://github.com/longhorn/go-common-libs/blob/main/profiler/profiler.go#L45-L53). Adding the `-sync-agent` suffix here to avoid potential confusion with the replica's profiler server.

derekbit: Got it. Can you add a comment?

derekbit:
After addressing the comment, we can merge the PR

PR #1560, opened 2025-08-28
feat: increase the timeout of tgt's timeout_handler

derekbit:

Issue Longhorn/longhorn#8572

To reduce the chance of hitting the read-only issue, the key is to properly tune the timeout setting for the tgt’s timeout_handler. Because the SCSI layer limits the number of outstanding write requests, this ensures that in-flight requests have sufficient time to complete rather than being prematurely terminated.


derekbit:
@mergify backport v1.8.x v1.9.x

PhanLe1010 (approved):
LGTM

PR #1582, opened 2025-09-05
fix(replica): validate size with backing image virtual size

COLDTURNIP:

Issue longhorn/longhorn#11362

To reject backed replica creation if the size is less than the backing image virtual size.

This PR depends on longhorn/go-common-libs/pull/132

### DIFF
--- a/pkg/backingfile/backingfile.go
+++ b/pkg/backingfile/backingfile.go
@@ -1,15 +1,12 @@
 package backingfile
 
 import (
-	"bufio"
-	"bytes"
 	"fmt"
 	"os"
-	"os/exec"
-	"strings"
-	"syscall"
 
 	"github.com/longhorn/sparse-tools/sparse"
+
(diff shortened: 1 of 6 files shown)


pkg/backingfile/backingfile.go
```diff
@@ -1,72 +1,24 @@
...
-	"strings"
-	"syscall"
 
 	"github.com/longhorn/sparse-tools/sparse"
 
+	"github.com/longhorn/go-common-libs/backingimage"
 	"github.com/longhorn/longhorn-engine/pkg/qcow"
 	"github.com/longhorn/longhorn-engine/pkg/types"
 	"github.com/longhorn/longhorn-engine/pkg/util"
 	diskutil "github.com/longhorn/longhorn-engine/pkg/util/disk"
```
derekbit: nit: reorg the imports

derekbit:
@mergify backport v1.10.x v1.9.x v1.8.x

PR #1680, opened 2026-01-08
fix: reduce retries for file sync

shuo-wu:

Issue longhorn/longhorn#12436


COLDTURNIP (approved):

derekbit (approved):
Verified. We don't need to update the test case.

PR #1750, opened 2026-04-12
chore(deps): updated vendored sparse-tools with retry on packetlost

mschneider82:
Issue longhorn/longhorn#8745

Updated vendored sparse-tools to include retry logic on packet loss during replica rebuilding. When network packets are lost during sparse file syncing, the operation could previously fail without retry, causing rebuild failures. The updated sparse-tools now retries on transient network errors using avast/retry-go.

This is a vendor-only change. The retry logic is implemented upstream in longhorn/sparse-tools (see linked PR below).

Upstream sparse-tools PR: https://github.com/longhorn/sparse-tools/pull/120


mschneider82:
@derekbit done

derekbit (approved):

PR #1752, opened 2026-04-17
fix: expand 16M space when starting replicas for encrypted volumes

mantissahz:

Issue # longhorn/longhorn#9205


pkg/replica/server.go
```diff
@@ -57,27 +62,28 @@ func (s *Server) Create(size int64) error {
...
 	}
 
-	_, info := s.Status()
+	state, info := s.Status()
 	sectorSize := s.getSectorSize()
 
-	logrus.Infof("Opening replica: dir %s, size %d, sector size %d", s.dir, info.Size, sectorSize)
-	r, err := New(s.ctx, info.Size, sectorSize, s.dir, s.backing, s.revisionCounterDisabled, s.unmapMarkDiskChainRemoved, s.snapshotMaxCount, s.snapshotMaxSize)
+	logrus.Infof("Opening replica: dir %s, size %d, sector size %d, state: %v, upgrading: %v", s.dir, info.Size, sectorSize, state, isUpgrade)
+	expandEncryptDev := state == types.ReplicaStateClosed && s.encrypted && info.CLIAPIVersion < types.CLIAPIVersionForSupportingExtendLuks2HeaderSize && !isUpgrade
```
derekbit: Can't we check the version in the control plane?

pkg/replica/replica.go
```diff
@@ -227,6 +231,16 @@ func construct(ctx context.Context, readonly bool, size, sectorSize int64, dir,
...
+		if expandEncryptDev && cliAPIVersion == types.CLIAPIVersionForSupportingExtendLuks2HeaderSize {
+			expandingSize, err := util.GetActualBackendSize(r.info.Size, true, types.CLIAPIVersionForSupportingExtendLuks2HeaderSize)
+			if err != nil {
+				return nil, err
+			}
+			if err := r.Expand(expandingSize); err != nil {
+				return nil, err
+			}
+			r.info.CLIAPIVersion = cliAPIVersion
+		}
```
derekbit: Move the check to longhorn manager.
Here, we only need to check sizes. If they are mismatched, we can directly expand the replica.

pkg/replica/rpc/server.go
```diff
@@ -131,8 +131,8 @@ func (rs *ReplicaServer) ReplicaGet(ctx context.Context, req *emptypb.Empty) (*e
 	return &enginerpc.ReplicaGetResponse{Replica: rs.getReplica()}, nil
 }
 
-func (rs *ReplicaServer) ReplicaOpen(ctx context.Context, req *emptypb.Empty) (*enginerpc.ReplicaOpenResponse, error) {
-	if err := rs.s.Open(); err != nil {
+func (rs *ReplicaServer) ReplicaOpen(ctx context.Context, req *enginerpc.ReplicaOpenRequest) (*enginerpc.ReplicaOpenResponse, error) {
+	if err := rs.s.Open(req.IsUpgrade); err != nil {
```
derekbit: Can we use `Size` not `IsUpgrade` istead?

derekbit (commented):
As discussed, the PR needs to be updated.

pkg/replica/replica.go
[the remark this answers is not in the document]
mantissahz: Removed. Thanks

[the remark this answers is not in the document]
mantissahz: Removed. Thanks

go.mod
```diff
@@ -111,3 +111,7 @@ require (
 	k8s.io/utils v0.0.0-20251002143259-bc988d571ff4 // indirect
 	sigs.k8s.io/json v0.0.0-20250730193827-2d320260d730 // indirect
 )
+
+replace github.com/longhorn/go-common-libs => github.com/mantissahz/go-common-libs v0.0.0-20260414021535-de2c9d803168
+
+replace github.com/longhorn/types => github.com/mantissahz/types v0.0.0-20260416091717-7ac3200b0bf0
```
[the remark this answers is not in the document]
mantissahz: Wait for PR merged.

app/cmd/replica.go
```diff
@@ -83,6 +83,16 @@ func ReplicaCmd() cli.Command {
...
+			cli.BoolFlag{
+				Name:   "encrypted",
+				Hidden: false,
+				Usage:  "Volume is encrypted",
+			},
+			cli.IntFlag{
+				Name:  "engine-cli-api-version",
+				Value: 0,
+				Usage: "Engine CLI API version",
+			},
```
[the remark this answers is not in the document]
mantissahz: Modified. `CLIAPLIVersion` will not be saved in `volume.meta`

pkg/replica/replica.go
mantissahz: Modified. Thanks.

pkg/replica/server.go
mantissahz: Modified. Thanks

pkg/replica/rpc/server.go
mantissahz: No, as we discussed, it represents that the live upgrade is processed, and we should not expand the backend file at this moment.
And add another parameter `expectBackendSize`.

pkg/replica/server.go
```diff
@@ -57,27 +62,33 @@ func (s *Server) Create(size int64) error {
...
-	logrus.Infof("Opening replica: dir %s, size %d, sector size %d", s.dir, info.Size, sectorSize)
-	r, err := New(s.ctx, info.Size, sectorSize, s.dir, s.backing, s.revisionCounterDisabled, s.unmapMarkDiskChainRemoved, s.snapshotMaxCount, s.snapshotMaxSize)
+	logrus.Infof("Opening replica: dir %s, size %d, sector size %d, state: %v, upgrading: %v", s.dir, info.Size, sectorSize, state, isUpgrade)
+	exapdingEncryptedDevSize := int64(0)
+	// isUpgrade is true only when live upgrading the engine. Before live upgrade, the volume-head image file is expanded to the expected backend size.
+	// Therefore, we should not expand the replica size for live upgrade.
+	if state == types.ReplicaStateClosed && s.encrypted && s.cliAPIVersion == types.CLIAPIVersionForSupportingExtendLuks2HeaderSize && !isUpgrade && expectedBackendSize > 0 && expectedBackendSize > info.Size {
+		exapdingEncryptedDevSize = expectedBackendSize
+	}
+	r, err := New(s.ctx, info.Size, sectorSize, s.dir, s.backing, s.revisionCounterDisabled, s.unmapMarkDiskChainRemoved, s.snapshotMaxCount, s.snapshotMaxSize, exapdingEncryptedDevSize)
```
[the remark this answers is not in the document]
mantissahz: Thanks. Modified.

pkg/replica/server.go
```diff
@@ -57,27 +62,33 @@ func (s *Server) Create(size int64) error {
...
+	state, info := s.Status()
 	sectorSize := s.getSectorSize()
 
-	logrus.Infof("Opening replica: dir %s, size %d, sector size %d", s.dir, info.Size, sectorSize)
-	r, err := New(s.ctx, info.Size, sectorSize, s.dir, s.backing, s.revisionCounterDisabled, s.unmapMarkDiskChainRemoved, s.snapshotMaxCount, s.snapshotMaxSize)
+	logrus.Infof("Opening replica: dir %s, size %d, sector size %d, state: %v, upgrading: %v", s.dir, info.Size, sectorSize, state, isUpgrade)
+	exapdingEncryptedDevSize := int64(0)
+	// isUpgrade is true only when live upgrading the engine. Before live upgrade, the volume-head image file is expanded to the expected backend size.
+	// Therefore, we should not expand the replica size for live upgrade.
+	if state == types.ReplicaStateClosed && s.encrypted && s.cliAPIVersion == types.CLIAPIVersionForSupportingExtendLuks2HeaderSize && !isUpgrade && expectedBackendSize > 0 && expectedBackendSize > info.Size {
```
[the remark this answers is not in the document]
mantissahz: Thanks. Modified.

pkg/replica/replica.go
```diff
@@ -144,17 +144,19 @@ func ReadInfo(dir string) (Info, error) {
...
 	// size and sectorSize don't matter because they will be read from metadata
 	// snapshotMaxCount and SnapshotMaxSize don't matter because readonly replica can't create a new disk
-	return construct(ctx, true, 0, diskutil.ReplicaSectorSize, dir, head, backingFile, false, false, types.MaximumTotalSnapshotCount, 0)
+	return construct(ctx, true, 0, diskutil.ReplicaSectorSize, dir, head, backingFile, false, false, types.MaximumTotalSnapshotCount, 0, int64(0))
 }
 
-func construct(ctx context.Context, readonly bool, size, sectorSize int64, dir, head string, backingFile *backingfile.BackingFile, disableRevCounter, unmapMarkDiskChainRemoved bool, snapshotMaxCount int, snapshotMaxSize int64) (*Replica, error) {
+func construct(ctx context.Context, readonly bool, size, sectorSize int64, dir, head string, backingFile *backingfile.BackingFile,
+	disableRevCounter, unmapMarkDiskChainRemoved bool, snapshotMaxCount int, snapshotMaxSize int64, exapdingEncryptedDevSize int64) (*Replica, error) {
 	if size%sectorSize != 0 {
```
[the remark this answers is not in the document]
mantissahz: Thanks. Modified.

pkg/types/types.go
```diff
@@ -51,6 +53,8 @@ const (
 
 	DefaultRebuildSyncConcurrentLimit = 1
 	SyncRetryCount                    = 3
+
+	CLIAPIVersionForSupportingExtendLuks2HeaderSize = lhtypes.CliAPIVersionForSupportingExtendLuks2HeaderSize
```
derekbit, line 57: nit: Why do we need to define the same constant for `lhtypes.CliAPIVersionForSupportingExtendLuks2HeaderSize`?

pkg/replica/server.go
```diff
@@ -57,27 +62,33 @@ func (s *Server) Create(size int64) error {
...
-	_, info := s.Status()
+	state, info := s.Status()
 	sectorSize := s.getSectorSize()
 
-	logrus.Infof("Opening replica: dir %s, size %d, sector size %d", s.dir, info.Size, sectorSize)
-	r, err := New(s.ctx, info.Size, sectorSize, s.dir, s.backing, s.revisionCounterDisabled, s.unmapMarkDiskChainRemoved, s.snapshotMaxCount, s.snapshotMaxSize)
+	logrus.Infof("Opening replica: dir %s, size %d, sector size %d, state: %v, upgrading: %v", s.dir, info.Size, sectorSize, state, isUpgrade)
+	expandingEncryptedDevSize := int64(0)
+	// isUpgrade is true only when live upgrading the engine. Before live upgrade, the volume-head image file is expanded to the expected backend size.
+	// Therefore, we should not expand the replica size for live upgrade.
```
derekbit, line 87: Why do we need to compare the engine CLI version? Can't we only compare the size?

pkg/replica/replica.go
```diff
@@ -144,17 +144,19 @@ func ReadInfo(dir string) (Info, error) {
...
 func NewReadOnly(ctx context.Context, dir, head string, backingFile *backingfile.BackingFile) (*Replica, error) {
 	// size and sectorSize don't matter because they will be read from metadata
 	// snapshotMaxCount and SnapshotMaxSize don't matter because readonly replica can't create a new disk
-	return construct(ctx, true, 0, diskutil.ReplicaSectorSize, dir, head, backingFile, false, false, types.MaximumTotalSnapshotCount, 0)
+	return construct(ctx, true, 0, diskutil.ReplicaSectorSize, dir, head, backingFile, false, false, types.MaximumTotalSnapshotCount, 0, int64(0))
 }
 
-func construct(ctx context.Context, readonly bool, size, sectorSize int64, dir, head string, backingFile *backingfile.BackingFile, disableRevCounter, unmapMarkDiskChainRemoved bool, snapshotMaxCount int, snapshotMaxSize int64) (*Replica, error) {
+func construct(ctx context.Context, readonly bool, size, sectorSize int64, dir, head string, backingFile *backingfile.BackingFile,
+	disableRevCounter, unmapMarkDiskChainRemoved bool, snapshotMaxCount int, snapshotMaxSize int64, expandingEncryptedDevSize int64) (*Replica, error) {
```
derekbit, line 159: `expandingEncryptedDevSize` is not easy to understand. Can you use `expectedxxxx` or `desiredxxxx`?

pkg/replica/replica.go
```diff
@@ -227,6 +229,11 @@ func construct(ctx context.Context, readonly bool, size, sectorSize int64, dir,
 		if err := r.openLiveChain(); err != nil {
 			return nil, err
 		}
+		if expandingEncryptedDevSize > 0 {
+			if err := r.Expand(expandingEncryptedDevSize); err != nil {
```
derekbit, line 233: Add a INFO log message here.

pkg/meta/version.go
```diff
@@ -2,7 +2,7 @@ package meta
 
 const (
 	// CLIAPIVersion used to communicate with user e.g. longhorn-manager
-	CLIAPIVersion    = 11
+	CLIAPIVersion    = 12
 	CLIAPIMinVersion = 8
```
derekbit, line 6: nit:
Could you add a comment for explaining CLIAPIMinVersion?

pkg/replica/client/client.go
```diff
@@ -234,7 +234,7 @@ func (c *ReplicaClient) OpenReplica() error {
 	ctx, cancel := context.WithTimeout(context.Background(), GRPCServiceCommonTimeout)
 	defer cancel()
 
-	if _, err := replicaServiceClient.ReplicaOpen(ctx, &emptypb.Empty{}); err != nil {
+	if _, err := replicaServiceClient.ReplicaOpen(ctx, &enginerpc.ReplicaOpenRequest{IsUpgrade: false}); err != nil {
 		return errors.Wrapf(err, "failed to open replica %v", c.replicaServiceURL)
```
derekbit, line 238: nit:
```suggestion
		return errors.Wrapf(err, "failed to open replica %v with IsUpgrade false", c.replicaServiceURL)
```

pkg/replica/server.go
```diff
@@ -57,27 +62,33 @@ func (s *Server) Create(size int64) error {
...
+	if state == types.ReplicaStateClosed && s.encrypted && s.cliAPIVersion == types.CLIAPIVersionForSupportingExtendLuks2HeaderSize && !isUpgrade && expectedBackendSize > 0 && expectedBackendSize > info.Size {
```
derekbit, line 88: Can you explain why `state == types.ReplicaStateClose` is required and other conditions in the comment?

derekbit, line 88: refactor the check like
```
if isExpansionRequired(...) {
    expandingEncryptedDevSize = expectedBackendSize
}
```

derekbit (commented):
As we discussed, you mentioned that you will persistent the encrypted volume expansion information (a flag indicating if the volume is expanded) in the volume.meta. However, I don't see the logic. Don't we need it?

app/cmd/replica.go
```diff
@@ -116,14 +126,17 @@ func startReplica(c *cli.Context) (err error) {
 		}
 	}
 
+	encrypted := c.Bool("encrypted")
+	engineCLIAPIVersion := c.Int("engine-cli-api-version")
```
derekbit, line 130: Why do we need the engineCLIAPIVersion? When a new replica process is created, isn't it using the latest engine image?

mantissahz:
After the PR is merged:
https://github.com/longhorn/types/pull/91

pkg/types/types.go
mantissahz: Yes, we can remove the redefined constant

pkg/replica/server.go
mantissahz: From the original thought, we only expand the engine after version 12.
But yes, we can make it all the old versions of volumes.

mantissahz:
> As we discussed, you mentioned that you will persistent the encrypted volume expansion information (a flag indicating if the volume is expanded) in the volume.meta. However, I don't see the logic. Don't we need it?

After updating some implementations and tests, we don't need the flag, and we only need the fixed engine size and check if the volume is encrypted or not.

pkg/replica/server.go
derekbit: > But yes, we can make it all the old versions of volumes.

What does it mean?

app/cmd/replica.go
mantissahz: For old image volumes, they need this `CLIAPIVersion` parameter when they are detaching and attaching (not upgrading the engine to/after 12 yet)

pkg/replica/server.go
mantissahz: Such as volumes with the engine image 1.11.1 (CLIPAIVersion is 11), 
when they are detaching and reattaching, we can expand the replica image size.
Then we don't need to compare the engine CLI version here.
We don't need this `engineCLIAPIVersion := c.Int("engine-cli-api-version")` parameter as well.

derekbit: If it is confirmed and not required, let's remove it.

pkg/replica/server.go
```diff
@@ -57,27 +62,33 @@ func (s *Server) Create(size int64) error {
...
+		expandingEncryptedDevSize = expectedBackendSize
+	}
```
derekbit: Can we move the checks into  `New(...)`?
Another questions:
1. Is `info.Size` the volume size or the backend size?
2. Have you checked the backup/restore?
3. Do we have `encrypted` information in the metafile?

derekbit:
> > As we discussed, you mentioned that you will persistent the encrypted volume expansion information (a flag indicating if the volume is expanded) in the volume.meta. However, I don't see the logic. Don't we need it?
> 
> After updating some implementations and tests, we don't need the flag, and we only need the fixed engine size and check if the volume is encrypted or not.

Can you elaborate on the new design?

pkg/replica/server.go
mantissahz: We can move the check to New but it will need to send the necessary variables to New()

1. The backend size (image file size)
3. I havd checked a long time ago, and I will check it again. It needs some modification I believe.

mantissahz:
>Can you elaborate on the new design?

It's not a new design.
In the previous implementation, it will set the `engine.Status.CurrentSize` to the backend size gotten from the engine server, such as it will be `1G+16MiB` if the volume is encrypted and the 1 GiB size is requested. Then we need to handle the `engine.Status.CurrentSize` in the volume/engine controllers.
But now, I make the `engine.Status.CurrentSize` to the requested volume size (as unencrypted volumes) if the backend size from the engine server is expected. Then we could only focus on the backend size with `engine.Status.CurrentSize` when creating the instance processes.

pkg/types/types.go
mantissahz: Removed it. Thanks.

pkg/replica/server.go
mantissahz: `engine-cli-api-version`, `engineCLIAPIVersion` and `CLIAPIVersion` are removed. Thanks.

pkg/replica/replica.go
mantissahz: Added. Thanks

pkg/meta/version.go
mantissahz: Added. Thanks.

pkg/replica/client/client.go
mantissahz: Updated. Thanks.

pkg/replica/server.go
mantissahz: Add more descriptions about the parameters. Thanks.

mantissahz: Updated. Thanks.

app/cmd/replica.go
mantissahz: Removed. Thanks.

pkg/controller/control.go
```diff
@@ -180,7 +180,7 @@ func (c *Controller) addReplica(address string, snapshotRequired bool, mode type
 		return err
 	}
 
-	newBackend, err := c.factory.Create(c.VolumeName, address, c.DataServerProtocol, c.sharedTimeouts)
+	newBackend, err := c.factory.Create(c.VolumeName, address, c.DataServerProtocol, c.sharedTimeouts, c.isUpgrade, 0)
```
derekbit, line 183: The expected backend size is 0 here. It is a bit weird. Could we set it to the nominal volume size instead?

pkg/replica/backup_test.go
```diff
@@ -27,7 +27,7 @@ func (s *TestSuite) TestBackup(c *C) {
 	err = os.Chdir(dir)
 	c.Assert(err, IsNil)
 
-	r, err := New(context.Background(), 10*mb, bs, dir, nil, false, false, 250, 0)
+	r, err := New(context.Background(), 10*mb, bs, dir, nil, false, false, 250, 0, false, 0)
```
derekbit, line 30: Same question as above.

pkg/controller/control.go
mantissahz: `0` represents that we don't need to expand when constructing replicas.
If we set it to the nominal volume size, we will need to move the check from the function `Open()` of the replica server to the function `construct(...)`.

I'm fine either way. WDYT, @derekbit?

pkg/replica/replica.go
mantissahz: Modified. Thanks.

pkg/replica/server.go
```diff
@@ -57,34 +60,61 @@ func (s *Server) Create(size int64) error {
...
+	// encrypted: Expanding backend size only when the volume is encrypted.
+	if !encrypted {
+		return false
+	}
+	// isUpgrade: Before live upgrade (isUpgrade is true), the volume-head image file is expanded to the expected backend size. Therefore, we should not expand the replica size for live upgrade.
+	if isUpgrade {
+		return false
+	}
+	// expectedBackendSize: The expected size of the backend, and it should be 16NiB greater than the existing replica image file size, otherwise we don't need to expand the replica size.
```
derekbit, line 110:
```suggestion
	// expectedBackendSize: The expected size of the backend, and it should be 16MiB greater than the existing replica image file size, otherwise we don't need to expand the replica size.
```

pkg/replica/server.go
```diff
@@ -57,34 +60,61 @@ func (s *Server) Create(size int64) error {
...
+	// existingReplicaImageFileSize: The current size of the replica image file.
+	if expectedBackendSize <= 0 || existingReplicaImageFileSize <= 0 {
```
derekbit, line 112: Just curious. Is negative value possible?

pkg/replica/server.go
```diff
@@ -57,34 +60,61 @@ func (s *Server) Create(size int64) error {
...
+		return false
+	}
+	return expectedBackendSize > existingReplicaImageFileSize
```
derekbit, line 115: Do we need to make it more specific? I mean 
```
return expectedBackendSize == existingReplicaImageFileSize + 16MiB
```

pkg/replica/server.go
```diff
@@ -57,34 +60,61 @@ func (s *Server) Create(size int64) error {
...
-	logrus.Infof("Opening replica: dir %s, size %d, sector size %d", s.dir, info.Size, sectorSize)
-	r, err := New(s.ctx, info.Size, sectorSize, s.dir, s.backing, s.revisionCounterDisabled, s.unmapMarkDiskChainRemoved, s.snapshotMaxCount, s.snapshotMaxSize)
+	logrus.Infof("Opening replica: dir %s, size %d, sector size %d, state: %v, upgrading: %v", s.dir, info.Size, sectorSize, state, isUpgrade)
+
+	expandingEncryptedDevSize := int64(0)
+	if isExpandingEncryptedDevRequired(state, s.encrypted, info.Size, expectedBackendSize, isUpgrade) {
+		expandingEncryptedDevSize = expectedBackendSize
+	}
+
+	r, err := New(s.ctx, info.Size, sectorSize, s.dir, s.backing, s.revisionCounterDisabled, s.unmapMarkDiskChainRemoved, s.snapshotMaxCount, s.snapshotMaxSize, s.encrypted, expandingEncryptedDevSize)
```
derekbit, line 89: I'm thinking why not 
```suggestion
	r, err := New(s.ctx, info.Size, sectorSize, s.dir, s.backing, s.revisionCounterDisabled, s.unmapMarkDiskChainRemoved, s.snapshotMaxCount, s.snapshotMaxSize, s.encrypted, expectedBackendSize)
```

and move `isExpandingEncryptedDevRequired` into `New()`?

pkg/replica/server.go
```diff
@@ -57,34 +60,61 @@ func (s *Server) Create(size int64) error {
 	sectorSize := s.getSectorSize()
 
 	logrus.Infof("Creating replica %s, size %d/%d", s.dir, size, sectorSize)
-	r, err := New(s.ctx, size, sectorSize, s.dir, s.backing, s.revisionCounterDisabled, s.unmapMarkDiskChainRemoved, s.snapshotMaxCount, s.snapshotMaxSize)
+	r, err := New(s.ctx, size, sectorSize, s.dir, s.backing, s.revisionCounterDisabled, s.unmapMarkDiskChainRemoved, s.snapshotMaxCount, s.snapshotMaxSize, s.encrypted, int64(0))
```
derekbit, line 63: `int64(0)` is quite confused. Can we set it to `size`?

pkg/replica/server.go
mantissahz: Modified for the backup and restore.

pkg/controller/control.go
mantissahz: Modified to use the c.Size()

pkg/replica/backup_test.go
mantissahz: Modified. Thanks.

pkg/replica/server.go
mantissahz: Modified. Thanks.

pkg/replica/server.go
mantissahz: Modified. Thanks.

pkg/replica/server.go
mantissahz: Modified. Thanks.

pkg/replica/server.go
mantissahz: Modified. Thanks.

pkg/replica/server.go
mantissahz: No, they will not be negative with our code flow, but if someone uses the `engine-binary/longhorn` command, it might be negative.

go.mod
```diff
@@ -111,3 +111,5 @@ require (
 	k8s.io/utils v0.0.0-20251002143259-bc988d571ff4 // indirect
 	sigs.k8s.io/json v0.0.0-20250730193827-2d320260d730 // indirect
 )
+
```
COLDTURNIP: Remember to remove the local replacement.

derekbit:
Hello @mantissahz You can update go.mod. Thank you.

go.mod
mantissahz: Removed and updated to the latest `longhorn/types`.
Thanks.

mantissahz:
>Hello @mantissahz You can update go.mod. Thank you.

Updated. Thanks.

COLDTURNIP (dismissed):
LGTM

davidcheng0922 (dismissed):
LGTM

pkg/replica/client/client.go
```diff
@@ -234,8 +234,8 @@ func (c *ReplicaClient) OpenReplica() error {
 	ctx, cancel := context.WithTimeout(context.Background(), GRPCServiceCommonTimeout)
 	defer cancel()
 
-	if _, err := replicaServiceClient.ReplicaOpen(ctx, &emptypb.Empty{}); err != nil {
-		return errors.Wrapf(err, "failed to open replica %v", c.replicaServiceURL)
+	if _, err := replicaServiceClient.ReplicaOpen(ctx, &enginerpc.ReplicaOpenRequest{IsUpgrade: false}); err != nil {
+		return errors.Wrapf(err, "failed to open replica %v with IsUpgrade false", c.replicaServiceURL)
```
derekbit, line 238: Need to test the case.
1. Create 2 volumes in v1.11.1. One is attached and the other one is detached.
2. Upgrade longhorn system to master-head
3. Without any engine upgrade, test the two scenarios
  1. Detach the first volume and then attach again => The attachment should work
  2. Attach the second volume => The attachment should work

derekbit (commented):
After addressing and testing the last comment, we can merge the PR.

derekbit:
@mantissahz Please fix the conflict. Thank you.

pkg/replica/client/client.go
mantissahz: Verification `PASSED`:
The setting `concurrent-automatic-engine-upgrade-per-node-limit` is `0`.
1. Create two workloads with two LH volumes (A and B) at v1.11.1 (docker.io/longhornio/longhorn-engine:v1.11.1)
2. Write some data to two volumes and save the checksums.
3. Scale down the workload B and wait for the volume B to be detached.
4. Upgrade to the `master-head` with PR images(`longhorn-manager`, `longhorn-instance-manager`, and `longhorn-engine`)
5. Scale up the workload B and wait for the volume B to be attached.
6. The data in workload B is consistent.
7. The replica B image file size is not expanded.
8. Scale down the workload A and wait for the volume A to be detached.
9. Scale up the workload A and wait for the volume A to be attached.
6. The data in workload A is consistent.
7. The replica A image file size is not expanded.

derekbit (approved):

PR #1753, opened 2026-04-20
fix: reset file sync progress on retry to prevent exceeding 100%

mschneider82:

Issue longhorn/longhorn#12949                                            

When a file sync fails and is retried on an unstable network, the progress bytes were added again without subtracting the previous
attempt's contribution, causing rebuild progress to exceed 100%.
                                                                         
This PR wraps `SyncFileOperations` with a per-file tracker (`perFileProgressOps`) that remembers how much each file has             
contributed to the global progress. On retry, `reset()` subtracts the previous contribution before re-syncing, keeping the progress        
counter accurate and never exceeding 100%.    
                                                                         
- The fix is contained in `pkg/sync/rpc/server.go`                       
- Unit tests in `pkg/sync/rpc/progress_tracker_test.go` demonstrate:
  - The bug (progress exceeding 100% without the tracker)                
  - The fix (progress staying correct with the tracker)                  
  - Multi-file retry scenario                                            
  - Idempotent reset behavior                                            

The root cause is in `fileSyncRemote()`: on retry, the sparse sync compares checksums and transfers differing blocks again, but                  
`RebuildStatus.processedSize` is never decremented for the previous attempt's progress. This causes the same blocks to be counted multiple times. The `perFileProgressOps` wrapper intercepts `UpdateSyncFileProgress` calls, tracks per-file contributions, and subtracts them on `reset()` which is called in the existing retry path.

### DIFF
--- a/pkg/sync/rpc/server.go
+++ b/pkg/sync/rpc/server.go
@@ -580,7 +580,19 @@ func (s *SyncAgentServer) fileSyncRemote(ctx context.Context, req *enginerpc.Fil
 
 	logrus.Infof("Replica %s starts to sync files %+v to from remote replicas %+v", req.ToHost, fileList, fromClientMap)
 
-	var ops sparserest.SyncFileOperations
+	// Create a per-file progress tracker for each non-empty file. The tracker forwards
+	// UpdateSyncFileProgress calls to RebuildStatus but also remembers how much it has
+	// contributed. On retry a new tracker is created and the old one is detached so that
+	// late callbacks from the previous ssync attempt cannot contaminate the new attempt.
+	fileProgressTrackers := make(map[string]*perFileProgressOps, len(req.SyncFileInfoList))
+	for _, info := range req.SyncFileInfoList {
+		if info.ActualSize > 0 {
+			fileProgressTrackers[info.ToFileName] = &perFileProgressOps{
+				SyncFileOperations: s.RebuildStatus,
+			}
+		}
+	}
+
 	fileStub := &sparserest.SyncFileStub{}
 
 	syncFileLock := &sync.Mutex{}
@@ -621,6 +633,11 @@ func (s *SyncAgentServer) fileSyncRemote(ctx context.Context, req *enginerpc.Fil
 							break
 						}
 						done = len(unsyncedFileMap) == 0
+						// Do not count size for disk meta file or empty disk file.
+						var fileOps sparserest.SyncFileOperations = fileStub
+						if syncFileInfo != nil && syncFileInfo.ActualSize != 0 {
+							fileOps = fileProgressTrackers[syncFileInfo.ToFileName]
+						}
 						syncFileLock.Unlock()
 
 						if syncFileInfo == nil {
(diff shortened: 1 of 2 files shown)


pkg/sync/rpc/server.go
```diff
@@ -686,6 +701,35 @@ func (s *SyncAgentServer) fileSyncRemote(ctx context.Context, req *enginerpc.Fil
...
+// reset subtracts this file's accumulated progress from the underlying counter and
+// zeroes the local tracker so the next attempt starts from a clean slate.
+func (p *perFileProgressOps) reset() {
+	p.mu.Lock()
+	contributed := p.contributed
+	p.contributed = 0
+	p.mu.Unlock()
+	if contributed > 0 {
+		p.SyncFileOperations.UpdateSyncFileProgress(-contributed)
+	}
```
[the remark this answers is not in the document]
mschneider82: not necessary

mschneider82:
added a fix, copilot broke it.

mschneider82:
@yangchiu can you retest this as well?

pkg/sync/rpc/server.go
```diff
@@ -686,6 +699,35 @@ func (s *SyncAgentServer) fileSyncRemote(ctx context.Context, req *enginerpc.Fil
...
+	sparserest.SyncFileOperations
+	mu          sync.Mutex
+	contributed int64
+}
+
+func (p *perFileProgressOps) UpdateSyncFileProgress(size int64) {
+	p.mu.Lock()
+	p.contributed += size
+	p.mu.Unlock()
+	p.SyncFileOperations.UpdateSyncFileProgress(size)
```
derekbit, line 727: Can we put them in the same critical section
```
func (p *perFileProgressOps) UpdateSyncFileProgress(size int64) {
	p.mu.Lock()
    defer p.mu.Unlock()

	p.contributed += size
	p.SyncFileOperations.UpdateSyncFileProgress(size)
}
```

pkg/sync/rpc/server.go
```diff
@@ -686,6 +699,35 @@ func (s *SyncAgentServer) fileSyncRemote(ctx context.Context, req *enginerpc.Fil
...
+// zeroes the local tracker so the next attempt starts from a clean slate.
+func (p *perFileProgressOps) reset() {
+	p.mu.Lock()
+	contributed := p.contributed
+	p.contributed = 0
+	p.mu.Unlock()
+	if contributed > 0 {
+		p.SyncFileOperations.UpdateSyncFileProgress(-contributed)
+	}
+}
```
derekbit: Same question as above

pkg/sync/rpc/server.go
```diff
@@ -642,18 +653,20 @@ func (s *SyncAgentServer) fileSyncRemote(ctx context.Context, req *enginerpc.Fil
 							} else {
 								logrus.WithError(err).Warnf("Replica %s failed to sync file %v from remote replica %v for %d times, will requeue the file and let other remote replicas handle it", req.ToHost, syncFileInfo.ToFileName, fromAddress, unsyncedFileRetryMap[syncFileInfo.ToFileName])
 								fileRetryBackoff.Next(syncFileInfo.ToFileName, time.Now())
+								if tracker, ok := fileProgressTrackers[syncFileInfo.ToFileName]; ok {
+									tracker.reset()
+								}
```
davidcheng0922, line 670: How about creating a tracker per retry attempt to avoid this race condition?

Currently each file reuses the same tracker across retries. Even with `reset()`, late async `UpdateSyncFileProgress` callbacks from a previous attempt can still be applied and contaminate the new attempt.

WDYT

mschneider82: Ok

pkg/sync/rpc/server.go
```diff
@@ -580,7 +580,19 @@ func (s *SyncAgentServer) fileSyncRemote(ctx context.Context, req *enginerpc.Fil
 
 	logrus.Infof("Replica %s starts to sync files %+v to from remote replicas %+v", req.ToHost, fileList, fromClientMap)
 
-	var ops sparserest.SyncFileOperations
+	// Create a per-file progress tracker for each non-empty file. The tracker forwards
+	// UpdateSyncFileProgress calls to RebuildStatus but also remembers how much it has
+	// contributed. On retry a new tracker is created and the old one is detached so that
+	// late callbacks from the previous ssync attempt cannot contaminate the new attempt.
+	fileProgressTrackers := make(map[string]*perFileProgressOps, len(req.SyncFileInfoList))
```
davidcheng0922, line 587: I think we need a lock to protect reads and writes to `fileProgressTrackers`.

There may be multiple goroutines accessing it concurrently.

Like `syncFileLock` for `unsyncedFileMap` and `unsyncedFileRetryMap`

mschneider82: done

derekbit (approved):
LGTM

davidcheng0922 (approved):

derekbit:
@mergify backport v1.11.x

PR #1758, opened 2026-04-22
fix(dataserver): handles EOF & goroutine leaks

apoorvajagtap:
Issue https://github.com/longhorn/longhorn/issues/12964

This commit  handles the well-defined `EOF` semantics returned by `io.ReadFull` which are currently being discarded:
- `Wire.Read`: special-case only clean `EOF` at the header read; wrap all other errors with context.
- For the payload read, any `EOF` is unexpected, so wraps unconditionally.
- `readFromWire`: use `errors.Is` instead of `==` for correctness with wrapped errors.
- `listenAndServeTCP`: demote clean `EOF` to `Debug`, keep Warn for real errors.

Addtionally, also ensures to close the connection & loops to prevent leaks:
- closes net.Conn()

Thanks to @shuo-wu for highlighting this while debugging the JIRA.

### DIFF
--- a/pkg/dataconn/server.go
+++ b/pkg/dataconn/server.go
@@ -1,6 +1,7 @@
 package dataconn
 
 import (
+	"errors"
 	"io"
 	"net"
 
@@ -35,7 +36,7 @@ func (s *Server) Handle() error {
 
 func (s *Server) readFromWire(ret chan<- error) {
 	msg, err := s.wire.Read()
-	if err == io.EOF {
+	if errors.Is(err, io.EOF) {
 		ret <- err
 		return
 	} else if err != nil {
--- a/pkg/dataconn/server.go
+++ b/pkg/dataconn/server.go
@@ -140,6 +140,8 @@ func (s *Server) write() {
 			if err := s.wire.Write(msg); err != nil {
 				logrus.WithError(err).Warn("Failed to write")
 			}
+			// Returning on <-s.done is intentionally skipped,
+			// as in-flight handlers may still send responses after the close signal.
 		}
 	}
 }
--- a/pkg/dataconn/wire.go
+++ b/pkg/dataconn/wire.go
@@ -7,6 +7,8 @@ import (
 	"io"
 	"net"
 	"unsafe"
+
+	"github.com/cockroachdb/errors"
 )
 
 type Wire struct {
@@ -66,8 +68,11 @@ func (w *Wire) Read() (*Message, error) {
 
 	offset := 0
 
-	if _, err := io.ReadFull(w.reader, w.readHeader); err != nil {
-		return nil, err
+	if bytesRead, err := io.ReadFull(w.reader, w.readHeader); err != nil {
+		if bytesRead == 0 && errors.Is(err, io.EOF) {
+			return nil, err
+		}
+		return nil, errors.Wrap(err, "failed to read message header")
 	}
 
 	msg.MagicVersion = binary.LittleEndian.Uint16(w.readHeader[offset:])
@@ -91,8 +96,12 @@ func (w *Wire) Read() (*Message, error) {
 	length = binary.LittleEndian.Uint32(w.readHeader[offset:])
 	if length > 0 {
 		msg.Data = make([]byte, length)
+		// EOF here = connection dropped mid-payload — always unexpected.
 		if _, err := io.ReadFull(w.reader, msg.Data); err != nil {
-			return nil, err
+			if errors.Is(err, io.EOF) {
+				err = io.ErrUnexpectedEOF
+			}
+			return nil, errors.Wrapf(err, "failed to read message payload (seq=%d, size=%d)", msg.Seq, length)
 		}
 	}
 
--- a/pkg/replica/rpc/dataserver.go
+++ b/pkg/replica/rpc/dataserver.go
@@ -59,6 +59,9 @@ func (s *DataServer) listenAndServeTCP() error {
 		logrus.Infof("New connection from: %v", conn.RemoteAddr())
 
 		go func(conn net.Conn) {
+			defer func() {
+				_ = conn.Close()
+			}()
 			server := dataconn.NewServer(conn, s.s)
 			if err := server.Handle(); err != nil {
 				if errors.Is(err, io.EOF) {
@@ -91,6 +94,9 @@ func (s *DataServer) listenAndServeUNIX() error {
 		}
 		logrus.Infof("New connection from: %v", conn.RemoteAddr())
 		go func(conn net.Conn) {
+			defer func() {
+				_ = conn.Close()
+			}()
 			server := dataconn.NewServer(conn, s.s)
 			if err := server.Handle(); err != nil {
 				if errors.Is(err, io.EOF) {
--- a/pkg/replica/rpc/dataserver.go
+++ b/pkg/replica/rpc/dataserver.go
@@ -1,7 +1,9 @@
 package rpc
 
 import (
+	"errors"
 	"fmt"
+	"io"
 	"net"
 
 	"github.com/sirupsen/logrus"
@@ -58,8 +60,13 @@ func (s *DataServer) listenAndServeTCP() error {
 
 		go func(conn net.Conn) {
 			server := dataconn.NewServer(conn, s.s)
-			if err = server.Handle(); err != nil {
-				logrus.WithError(err).Warn("failed to handle data server")
+			if err := server.Handle(); err != nil {
+				if errors.Is(err, io.EOF) {
+					// Clean remote close: this is normal on detach, engine restart.
+					logrus.WithError(err).Info("Data server connection closed by remote")
+					return
+				}
+				logrus.WithError(err).Warn("Failed to handle data server")
 			}
 		}(conn)
 	}
@@ -85,8 +92,12 @@ func (s *DataServer) listenAndServeUNIX() error {
 		logrus.Infof("New connection from: %v", conn.RemoteAddr())
 		go func(conn net.Conn) {
 			server := dataconn.NewServer(conn, s.s)
-			if err = server.Handle(); err != nil {
-				logrus.WithError(err).Warn("failed to handle data server")
+			if err := server.Handle(); err != nil {
+				if errors.Is(err, io.EOF) {
+					logrus.WithError(err).Info("Data server connection closed by local peer")
+					return
+				}
+				logrus.WithError(err).Warn("Failed to handle data server")
 			}
 		}(conn)
 	}


pkg/replica/rpc/dataserver.go
```diff
@@ -86,6 +93,11 @@ func (s *DataServer) listenAndServeUNIX() error {
 		go func(conn net.Conn) {
 			server := dataconn.NewServer(conn, s.s)
 			if err = server.Handle(); err != nil {
+				if errors.Is(err, io.EOF) {
+					// Clean remote close: this is normal on detach, engine restart.
+					logrus.WithError(err).Debug("data server connection closed by remote")
```
shuo-wu: NIT: This is a UNIX domain socket connection, which is mainly considered as a local connection

shuo-wu (dismissed):
In general, LGTM.

Notice that the confusing `EOF` errors in the original JIRA ticket may not only be caused by `io.ReadFull` in this data connection server. For example, the metafile read during the replica server init can contribute as well. But we can use another Github ticket to track it

pkg/replica/rpc/dataserver.go
apoorvajagtap: Oh, my bad, I just copy pasted & forgot to edit the log message, corrected it now. Thanks!

pkg/dataconn/wire.go
```diff
@@ -91,8 +95,9 @@ func (w *Wire) Read() (*Message, error) {
 	length = binary.LittleEndian.Uint32(w.readHeader[offset:])
 	if length > 0 {
 		msg.Data = make([]byte, length)
+		// EOF here = connection dropped mid-payload — always unexpected.
 		if _, err := io.ReadFull(w.reader, msg.Data); err != nil {
```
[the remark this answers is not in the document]
apoorvajagtap: @shuo-wu I think it makes sense here, so I have pushed the latest commit that ensures that if EOF is received while reading payload, `wire.Read` returns `ErrUnexpectedEOF`, to handle the failure better instead of claiming it as "clean close". PTAL.

derekbit:
@apoorvajagtap Could you address all the comments from @copilot? Thanks

pkg/replica/rpc/dataserver.go
```diff
@@ -58,7 +60,12 @@ func (s *DataServer) listenAndServeTCP() error {
 
 		go func(conn net.Conn) {
 			server := dataconn.NewServer(conn, s.s)
-			if err = server.Handle(); err != nil {
+			if err := server.Handle(); err != nil {
+				if errors.Is(err, io.EOF) {
```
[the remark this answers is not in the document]
apoorvajagtap: @derekbit @shuo-wu I find this as a valid suggestion, but would prefer analysing the details & sending these changes in a separate PR to keep the intentions clear. Or let me know if you prefer to have another commit addressing this in this PR itself.

pkg/dataconn/wire.go
```diff
@@ -66,8 +67,11 @@ func (w *Wire) Read() (*Message, error) {
 
 	offset := 0
 
-	if _, err := io.ReadFull(w.reader, w.readHeader); err != nil {
-		return nil, err
+	if bytesRead, err := io.ReadFull(w.reader, w.readHeader); err != nil {
+		if bytesRead == 0 && errors.Is(err, io.EOF) {
+			return nil, err
+		}
+		return nil, fmt.Errorf("failed to read message header: %w", err)
```
derekbit:
```suggestion
		return nil, errors.Wrapf(err, "failed to read message header")
```

pkg/dataconn/wire.go
```diff
@@ -91,8 +95,12 @@ func (w *Wire) Read() (*Message, error) {
 	length = binary.LittleEndian.Uint32(w.readHeader[offset:])
 	if length > 0 {
 		msg.Data = make([]byte, length)
+		// EOF here = connection dropped mid-payload — always unexpected.
 		if _, err := io.ReadFull(w.reader, msg.Data); err != nil {
-			return nil, err
+			if errors.Is(err, io.EOF) {
+				err = io.ErrUnexpectedEOF
+			}
+			return nil, fmt.Errorf("failed to read message payload (seq=%d, size=%d): %w", msg.Seq, length, err)
```
derekbit:
```suggestion
			return nil, errors.Wrapf(err, "failed to read message payload (seq=%d, size=%d): %w", msg.Seq, length)
```

pkg/replica/rpc/dataserver.go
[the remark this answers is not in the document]
derekbit: Thanks @apoorvajagtap. You can create a new commit but in the same PR to close the connection leak. Thanks.

derekbit:
@apoorvajagtap The titles of the PR and commit should follow the conventional commit rules.
https://github.com/longhorn/longhorn-engine/actions/runs/25720549945/job/75529977081?pr=1758

pkg/dataconn/server.go
```diff
@@ -139,6 +140,8 @@ func (s *Server) write() {
 			if err := s.wire.Write(msg); err != nil {
 				logrus.WithError(err).Warn("Failed to write")
 			}
+			// Drains one done signal and return to avoid leaking the goroutine.
+			return
```
apoorvajagtap: @derekbit as [done](https://github.com/longhorn/longhorn-engine/blob/master/pkg/dataconn/server.go#L23) has a buffer of 5, there seems to be a possibility of goroutine leak here as well, so I have added a return to prevent that. Let me know in case this was intentional.

pkg/dataconn/server.go
```diff
@@ -139,6 +140,8 @@ func (s *Server) write() {
 			if err := s.wire.Write(msg); err != nil {
 				logrus.WithError(err).Warn("Failed to write")
 			}
+			// Drains one done signal and return to avoid leaking the goroutine.
```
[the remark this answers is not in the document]
derekbit: cc @apoorvajagtap Could you update the comment?

pkg/dataconn/wire.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"encoding/binary"
 	"fmt"
+	"github.com/pkg/errors"
 	"io"
 	"net"
 	"unsafe"
```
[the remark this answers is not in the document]
derekbit: @apoorvajagtap Please update the imports as the comment. Thanks.

pkg/dataconn/server.go
apoorvajagtap: Copilot highlighted a valid concern [here](https://github.com/longhorn/longhorn-engine/pull/1758#discussion_r3225717595). So, removing the return & keeping it as is, as the leak here seems preferred trade-off.

derekbit:
@apoorvajagtap 
CI failed. Could you fix it?
```
126.6 pkg/replica/rpc/dataserver.go:62:20: Error return value of `conn.Close` is not checked (errcheck)
126.6 			defer conn.Close()
```

apoorvajagtap:
> @apoorvajagtap CI failed. Could you fix it?
> 
> ```
> 126.6 pkg/replica/rpc/dataserver.go:62:20: Error return value of `conn.Close` is not checked (errcheck)
> 126.6 			defer conn.Close()
> ```

The latest commit has the changes. I have not logged the errors as that might add noise in the logs, but just handled the err returned. Let me know if we prefer otherwise.

pkg/replica/rpc/dataserver.go
```diff
@@ -57,8 +59,16 @@ func (s *DataServer) listenAndServeTCP() error {
...
 		go func(conn net.Conn) {
+			defer func() {
+				_ = conn.Close()
+			}()
 			server := dataconn.NewServer(conn, s.s)
-			if err = server.Handle(); err != nil {
+			if err := server.Handle(); err != nil {
+				if errors.Is(err, io.EOF) {
+					// Clean remote close: this is normal on detach, engine restart.
+					logrus.WithError(err).Debug("data server connection closed by remote")
```
derekbit:
```suggestion
					logrus.WithError(err).Info("Data server connection closed by remote")
```

pkg/replica/rpc/dataserver.go
```diff
@@ -84,8 +94,15 @@ func (s *DataServer) listenAndServeUNIX() error {
...
 		logrus.Infof("New connection from: %v", conn.RemoteAddr())
 		go func(conn net.Conn) {
+			defer func() {
+				_ = conn.Close()
+			}()
 			server := dataconn.NewServer(conn, s.s)
-			if err = server.Handle(); err != nil {
+			if err := server.Handle(); err != nil {
+				if errors.Is(err, io.EOF) {
+					logrus.WithError(err).Debug("data server connection closed by local peer")
```
derekbit:
```suggestion
					logrus.WithError(err).Info("Data server connection closed by local peer")
```

pkg/replica/rpc/dataserver.go
```diff
@@ -84,8 +94,15 @@ func (s *DataServer) listenAndServeUNIX() error {
...
+					return
+				}
 				logrus.WithError(err).Warn("failed to handle data server")
```
derekbit:
```suggestion
				logrus.WithError(err).Warn("Failed to handle data server")
```

pkg/replica/rpc/dataserver.go
```diff
@@ -57,8 +59,16 @@ func (s *DataServer) listenAndServeTCP() error {
...
+			}()
 			server := dataconn.NewServer(conn, s.s)
-			if err = server.Handle(); err != nil {
+			if err := server.Handle(); err != nil {
+				if errors.Is(err, io.EOF) {
+					// Clean remote close: this is normal on detach, engine restart.
+					logrus.WithError(err).Debug("data server connection closed by remote")
+					return
+				}
 				logrus.WithError(err).Warn("failed to handle data server")
```
derekbit:
```suggestion
				logrus.WithError(err).Warn("Failed to handle data server")
```

pkg/dataconn/wire.go
```diff
@@ -66,8 +68,11 @@ func (w *Wire) Read() (*Message, error) {
 
 	offset := 0
 
-	if _, err := io.ReadFull(w.reader, w.readHeader); err != nil {
-		return nil, err
+	if bytesRead, err := io.ReadFull(w.reader, w.readHeader); err != nil {
+		if bytesRead == 0 && errors.Is(err, io.EOF) {
+			return nil, err
+		}
+		return nil, errors.Wrapf(err, "failed to read message header")
```
c3y1huang:
```suggestion
		return nil, errors.Wrap(err, "failed to read message header")
```

c3y1huang (commented):
Might be worth checking if `Server.write` should return on `<-s.done` case.
https://github.com/longhorn/longhorn-engine/blob/0e43cac222ae0a98c27a8d40cab35a7b6b49cbfe/pkg/dataconn/server.go#L134-L142/

---

Update: It seems the `return` was originally targeted in this fix but was later removed. I couldn't find the Copilot concern mentioned in the [comment](https://github.com/longhorn/longhorn-engine/pull/1758/#discussion_r3225738106). Could you explain what the concern was? If it's valid, suggest adding an inline doc comment for future readers.

apoorvajagtap:
> Might be worth checking if `Server.write` should return on `<-s.done` case. https://github.com/longhorn/longhorn-engine/blob/0e43cac222ae0a98c27a8d40cab35a7b6b49cbfe/pkg/dataconn/server.go#L134-L142/
> 
> Update: It seems the `return` was originally targeted in this fix but was later removed. I couldn't find the Copilot concern mentioned in the [comment](https://github.com/longhorn/longhorn-engine/pull/1758/#discussion_r3225738106). Could you explain what the concern was? If it's valid, suggest adding an inline doc comment for future readers.

Hi @c3y1huang, yes I added it and then copilot raised the concern [here](https://github.com/longhorn/longhorn-engine/pull/1758#discussion_r3225717595), which seemed like a valid point given the scope I am aware of. Let me know your thoughts!

pkg/replica/rpc/dataserver.go
apoorvajagtap: @derekbit won't having these as `Info` increase noise in the logs each time the connection is closed willingly? 
cc @c3y1huang

c3y1huang:
> Hi @c3y1huang, yes I added it and then copilot raised the concern [here](https://github.com/longhorn/longhorn-engine/pull/1758#discussion_r3225717595), which seemed like a valid point given the scope I am aware of. Let me know your thoughts!

Got it. Suggest adding a short inline doc comment to make it clear this is intentional. Without context, the next reader might go through the same reasoning and propose the same change.

Something like: not returning on `<-s.done` is intentional, since an in-flight handler may still send a response after the close signal.

c3y1huang (approved):

derekbit:
@mergify backport v1.12.x

derekbit:
@mergify backport v1.11.x v1.10.x

PR #1762, opened 2026-04-28
fix: build failure

c3y1huang:

Issue https://github.com/longhorn/longhorn/issues/12877

https://github.com/longhorn/longhorn-engine/actions/runs/25027348196


derekbit (approved):
LGTM, but it is a temporarily solution. We can see how to improve it later.

derekbit:
@mergify backport v1.10.x v1.9.x

PR #1813, opened 2026-05-14
fix(client): prevent NPE in GetControllerReplicaInfo

derekbit:

Issue Longhorn/longhorn#13087

During replica rebuild storms, the gRPC ControllerReplicaCreate call may succeed but return a ControllerReplica response with a nil Address field. Accessing cr.Address.Address without a nil check causes a SIGSEGV panic, crashing the instance-manager process and triggering cascading volume detachments.

Change GetControllerReplicaInfo to return an error when the response is nil or missing its Address field, and update all call sites (ReplicaList, ReplicaGet, ReplicaCreate, ReplicaUpdate) to propagate the error instead of panicking.

### DIFF
--- a/pkg/controller/client/controller_client.go
+++ b/pkg/controller/client/controller_client.go
@@ -96,11 +96,18 @@ func GetVolumeInfo(v *enginerpc.Volume) *types.VolumeInfo {
 	}
 }
 
-func GetControllerReplicaInfo(cr *enginerpc.ControllerReplica) *types.ControllerReplicaInfo {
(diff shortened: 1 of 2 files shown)


derekbit:
@mergify backport v1.9.x v1.10.x v1.11.x v1.12.x

c3y1huang (approved):

liyimeng:
we hit on this issue as well, won't backport to 1.8.x?

derekbit:
> we hit on this issue as well, won't backport to 1.8.x?

Hello, 1.8.x is EOL. Please update to the supported versions. Thank you.

PR #1818, opened 2026-05-14
fix: return error when ControllerServer cannot find a replica based on the address

shuo-wu:

Issue https://github.com/longhorn/longhorn/issues/13087

Otherwise, this `ContollerServer` gRPC server will marshal the response as a nil `*ControllerReplica` pointer. Then, the client side will unmarshals the empty-byte result into a **zero-value** struct `cr *enginerpc.ControllerReplica`. This `cr` itself is non-nil but all its pointer fields like `cr.Address` are nil. Finally, it leads to some NPE issues when the callers of the client try to use `cr`

### DIFF
--- a/pkg/controller/control.go
+++ b/pkg/controller/control.go
@@ -524,7 +524,12 @@ func (c *Controller) RemoveReplica(address string) error {
 }
 
 func (c *Controller) ListReplicas() []types.Replica {
-	return c.replicas
+	c.RLock()
+	defer c.RUnlock()
+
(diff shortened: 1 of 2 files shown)


c3y1huang (dismissed):

derekbit (dismissed):

derekbit:
@mergify backport v1.9.x v1.10.x v1.11.x v1.12.x

c3y1huang (approved):

PR #1847, opened 2026-06-06
feat: support crash-safe WAL for replica snapshot chain operations

derekbit:

Issue longhorn/longhorn#13259

### DIFF
--- a/pkg/replica/replica.go
+++ b/pkg/replica/replica.go
@@ -408,15 +408,121 @@ func (r *Replica) RemoveDiffDisk(name string, force bool) error {
 	if name == r.info.Head {
 		return fmt.Errorf("cannot delete the active differencing disk")
 	}
-
-	if err := r.removeDiskNode(name, force); err != nil {
-		return err
+	if r.wal == nil {
+		return fmt.Errorf("RemoveDiffDisk: journal is not initialised")
+	}
+	if _, ok := r.diskData[name]; !ok {
+		return fmt.Errorf("cannot find disk %v", name)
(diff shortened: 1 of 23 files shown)


pkg/replica/replica.go
```diff
@@ -713,14 +724,56 @@ func (r *Replica) encodeToFile(obj interface{}, file string) (rollbackFunc func(
 	}
 
 	if err := json.NewEncoder(f).Encode(&obj); err != nil {
+		_ = f.Close()
```
c3y1huang: Suggest not swallowing the errors.

pkg/util/fsync.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"os"
+
+	"github.com/cockroachdb/errors"
+)
+
+// FsyncDir fsyncs a directory so that file create / rename / unlink
+// entries within it are durable. POSIX requires this after os.Rename to
+// guarantee crash recovery sees the new name.
+func FsyncDir(dir string) error {
```
c3y1huang: Can we move this into go-common-lib? It looks very similar to [SyncFile](https://github.com/longhorn/go-common-libs/blob/c237ce0a9cbfffd3f9476d4b293a07352c828d2e/io/file.go#L223).

derekbit: Done. Thanks.

pkg/replica/replica.go
derekbit: Done! Thanks.

PR #1873, opened 2026-06-12
fix: make engine runtime paths follow LONGHORN_DATA_PATH

davidcheng0922:

https://github.com/longhorn/longhorn/issues/9083

This change makes longhorn-engine runtime helper paths configurable through `LONGHORN_DATA_PATH` instead of assuming `/var/lib/longhorn`.

Previously, engine-side paths such as unix domain sockets, freeze points, and snapshot hash lock files were still tied to `/var/lib/longhorn`. To support environments where Longhorn is installed with a different data root, longhorn-engine now derives these paths from the configured Longhorn data path while keeping `/var/lib/longhorn` as the default.

### DIFF
--- a/pkg/replica/hash.go
+++ b/pkg/replica/hash.go
@@ -24,6 +24,7 @@ import (
 	xattrType "github.com/longhorn/sparse-tools/types"
 
 	"github.com/longhorn/longhorn-engine/pkg/types"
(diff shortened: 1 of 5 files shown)


derekbit (approved):
LGTM

PR #1890, opened 2026-06-25
fix: use UUID in expansion snapshot name to ensure uniqueness across volumes

### LINKED ISSUE #13386: [BUG] Volume expansion snapshot CR name expand-<size> collides across volumes
### Describe the Bug

The system snapshot created during volume expansion is named expand-<size> using only the target size, with no volume identifier or UUID. When two volumes are expanded to the same byte size only one Snapshot CR can exist for that name.

### To Reproduce

1. Two Longhorn volumes, both at size X.
2. Expand both to the same size Y at roughly the same time.
3. kubectl -n longhorn-system get snapshot | grep expand -> only one CR named expand-<Y> exists.
4. Both on-disk snapshots exist (visible per-volume in the UI as system/hidden). The volume that lost the create race has no Snapshot CR for it.

### Expected Behavior

Each volume's expansion snapshot gets its own Snapshot CR (name disambiguated by volume, as other system snapshots are by UUID).

### Support Bundle for Troubleshooting

no bundle

### Environment

- Longhorn version: v1.10.0
- Impacted volume (PV): any
- Installation method (e.g. Rancher Catalog App/Helm/Kubectl): kubectl/kustomize
- Kubernetes distro (e.g. RKE/K3s/EKS/OpenShift) and version: RKE2
  - Number of control plane nodes in the cluster: 3
  - Number of worker nodes in the cluster: 4
- Node config
  - OS type and version: ubuntu-22.04
  - Kernel version:
  - CPU per node:
  - Memory per node:
  - Disk type (e.g. SSD/NVMe/HDD):
  - Network bandwidth between the nodes (Gbps):
- Underlying Infrastructure (e.g. on AWS/GCE, EKS/GKE, VMWare/KVM, Baremetal): VMware
- Number of Longhorn volumes in the cluster: ~70

### Additional context

Observed during recovery (low-level-hacks) of stuck pv resize after an iscsid restart (many related issues like #10544).

Not sure if this conflict has any real impact, but at least causes confusion because:
- only single expand-y snapshot cr exists, referencing 1st PV
- 2nd PV shows the same system/hidden snapshot in UI, no CR exists referencing this PV
- same expand-y shows up in multiple engine.status.snapshots, volumeattachment..attachmenttickets etc for different volumes

### Workaround and Mitigation

Maybe change Replica.Expand to use UUID or something besides size in GenerateExpansionSnapshotName. Verified this is same from v1.10.0 to master.

kavirakesh14:
## Issue
When two volumes expand to the same size concurrently at the same time, both replicas call
`GenerateExpansionSnapshotName(size)` and get the same name (`expand-<size>`).
Since Snapshot CR names are cluster-scoped, only one CR is created — the
second volume's expansion snapshot exists on disk but has no CR.

## Fix
This PR fixes that by appending a UUID to the snapshot name so every
expansion event always gets a unique name.

```go
// before
return fmt.Sprintf(expansionSnapshotInfix, size)
// after
return fmt.Sprintf(expansionSnapshotInfix+"-%s", size, util.UUID())
```

## Test
Added a unit test in `pkg/util/disk/disk_test.go` that checks:
- Generated name starts with the correct `expand-<size>-` prefix
- Two calls with the same size always produce different names

Fixes: longhorn/longhorn#13386


shuo-wu (approved):

pkg/util/disk/disk.go
```diff
@@ -43,7 +45,7 @@ func GetSnapshotNameFromDiskName(diskName string) (string, error) {
 }
 
 func GenerateExpansionSnapshotName(size int64) string {
-	return fmt.Sprintf(expansionSnapshotInfix, size)
+	return fmt.Sprintf(expansionSnapshotInfix+"-%s", size, util.UUID())
```
shuo-wu, line 48: Wait. We cannot use random `util.UUID()` here. Otherwise, each replica of a volume will generate a different snap name during expansion.

shuo-wu (changes requested):

pkg/util/disk/disk.go
kavirakesh14: @shuo-wu Thanks for the review,   I used UUID initially since the issue suggested it, but you're right. I'm switching to volume name instead since all replicas of the same volume share the same name, so they'll always generate the same snapshot name while still being unique across different volumes. Does this approach work?

derekbit: @kavirakesh14 Sounds make sense to. Could you update your PR? Thanks.

derekbit:
@kavirakesh14 Thanks for your contribution, but the PR is pending for a while. Let's close it.
Replace the PR with https://github.com/longhorn/longhorn-engine/pull/1989 instead. 
Thank you.

PR #1970, opened 2026-08-12
fix: preserve OS TCP keepalive settings for sync agent

dnesting:

issue longhorn/longhorn#13703

Create the sync-agent listener without Go TCP keepalive timing overrides, then enable keepalive on each accepted connection. This prevents a single missed probe from canceling a long-running FileSend request and interrupting the replica rebuild.

I've created a companion test at https://github.com/longhorn/longhorn-tests/pull/3639.  I'm not sure the right way to run this test, but separately I built an engine image with the fix in this PR and re-ran my original reproduction on my own cluster and both reproduced the original failure and could not reproduce with this fix.  I also saw `ss` output confirming that the TCP keepalive timer was set to 120 minutes, which I guess is my OS default.

### DIFF
--- a/app/cmd/sync_agent.go
+++ b/app/cmd/sync_agent.go
@@ -82,7 +82,11 @@ func startSyncAgent(c *cli.Context) error {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	listen, err := net.Listen("tcp", listenPort)
+	// Preserve the OS keepalive timings. The listener enables keepalive on each
+	// accepted connection without letting the Go standard library replace them.
+	// https://github.com/grpc/grpc-go/issues/6250
+	listenConfig := net.ListenConfig{KeepAlive: -1}
+	listen, err := listenConfig.Listen(ctx, "tcp", listenPort)
 	if err != nil {
 		return errors.Wrap(err, "failed to listen")
 	}
@@ -91,7 +95,26 @@ func startSyncAgent(c *cli.Context) error {
 
 	logrus.Infof("Listening on sync %s", listenPort)
 
-	return server.Serve(listen)
+	return server.Serve(tcpKeepAliveListener{Listener: listen})
+}
+
+type tcpKeepAliveListener struct {
+	net.Listener
+}
+
+func (l tcpKeepAliveListener) Accept() (net.Conn, error) {
+	conn, err := l.Listener.Accept()
+	if err != nil {
+		return nil, err
+	}
+
+	if tcpConn, ok := conn.(*net.TCPConn); ok {
+		if err := tcpConn.SetKeepAlive(true); err != nil {
+			logrus.WithError(err).Warnf("Failed to enable TCP keepalive on connection from %v; using connection without TCP keepalive", conn.RemoteAddr())
+		}
+	}
+
+	return conn, nil
 }
 
 func doReset(c *cli.Context) error {


app/cmd/sync_agent.go
```diff
@@ -91,7 +94,35 @@ func startSyncAgent(c *cli.Context) error {
...
+
+	if err := tcpConn.SetKeepAlive(true); err != nil {
+		if errClose := tcpConn.Close(); errClose != nil {
+			logrus.WithError(errClose).Error("Failed to close TCP connection")
+		}
+		return nil, errors.Wrap(err, "failed to enable TCP keepalive")
+	}
+
+	return tcpConn, nil
 }
```
c3y1huang: Both error paths return non-temporary errors, if `Accept` retuns one of them, gRPC gives up and `Serve` [returns](https://github.com/grpc/grpc-go/blob/v1.83.0/server.go#L933-L964). Since that is the last thing `startSyncAgent` does, the entire sync-agent exits. So if `SetKeepAlive` ever hiccups on a connection, it could take down the server and every rebuilding in flight. Also, IIUC the non-TCP check can't actually trigger today, but if the listener ever gets wrapped later, it would start rejecting everything, and IIUC gRPC is fine with any `net.Conn` anyway.

Would it make more sense to log and move on instead? This would match what the Go lib did until [Go 1.12](https://github.com/golang/go/blob/go1.12/src/net/http/server.go#L3255-L3271) and later [moved to net](https://github.com/golang/go/blob/go1.26.5/src/net/tcpsock.go#L302). I am thinking something like:
```go
if tcpConn, ok := conn.(*net.TCPConn); ok {
    if err := tcpConn.SetKeepAlive(true); err != nil {
        logrus.WithError(err).Warnf("Failed to enable TCP keepalive on connection from %v", conn.RemoteAddr())
    }
}

return conn, nil
```

WDYT?

c3y1huang (commented):
I think we should land this first, but consider having a shared helper for other places with the same plain net.Listen pattern serving gRPC.

And, remember to sign-off your commit for the DCO check.

app/cmd/sync_agent.go
dnesting: Yeah this seems better.

c3y1huang (commented):
LGTM. Two commits (09fd5656d962b537056d46a96e4ee4a3868797ab & a87da45b1a3afeeb5b110e0225de05c4b1515db1) are review fixups; should fold them before merge for a clean history.

dnesting:
I'm assuming you mean squash the commits together, so I've done that.

c3y1huang (commented):
commit-lint is failing, please wrap the commit body.

app/cmd/sync_agent.go
```diff
@@ -91,7 +95,26 @@ func startSyncAgent(c *cli.Context) error {
...
+
+func (l tcpKeepAliveListener) Accept() (net.Conn, error) {
+	conn, err := l.Listener.Accept()
+	if err != nil {
+		return nil, err
+	}
+
+	if tcpConn, ok := conn.(*net.TCPConn); ok {
+		if err := tcpConn.SetKeepAlive(true); err != nil {
+			logrus.WithError(err).Warnf("Failed to enable TCP keepalive on connection from %v", conn.RemoteAddr())
```
derekbit:
```suggestion
			logrus.WithError(err).Warnf("Failed to enable TCP keepalive on connection from %v; using connection without TCP keepalive", conn.RemoteAddr())
```

shuo-wu (commented):
~~The file sync service is a pure HTTP server, even though its caller or the client caller relies on the gRPC connection to maintain this tmp file sync HTTP operation. In other words, this file sync connection itself should not be affected by the gRPC `TCP_USER_TIMEOUT` setting and failed by one keep-alive probe drop.~~ 

Never mind, your fix correctly targets to the sync agent gRPC service, which will cancel the file sync connection if its connection dies.

shuo-wu (approved):

derekbit (approved):

derekbit:
@mergify backport v1.12.x

PR #1976, opened 2026-08-18
test(integration): fix flaky test_metrics timing race

derekbit:

https://github.com/longhorn/longhorn/issues/13771

test_metrics issues a single IO burst and then polls metrics_get() for about one second. The controller reports metrics from the last completed 1-second bucket, so a single burst is only visible for one ~1s window that can end up to ~2s after the IO. The poll budget (10 iters x 0.1s) is the same magnitude as the metrics period, so when the IO lands early in a bucket the visible window opens after the poll budget is exhausted and the test reads 0, failing probabilistically on readThroughput.

Raise max_iters from 10 to 30 so the poll budget (~2.9s) spans more than two metrics periods and reliably captures the non-zero window. No product code change.

### DIFF
--- a/integration/data/test_basic_ops.py
+++ b/integration/data/test_basic_ops.py
@@ -61,8 +61,10 @@ def test_metrics(grpc_replica1, grpc_replica2, grpc_controller):  # NOQA
     data = random_string(length)
     verify_data(rw_dev, offset, data)
 
-    # metrics update period is 1 second
-    max_iters = 10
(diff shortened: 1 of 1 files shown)


mantissahz (approved):

COLDTURNIP (approved):

derekbit:
@mergify backport v1.12.x

derekbit:
@mergify backport v1.11.x

PR #1989, opened 2026-08-24
fix: include volume name in expansion snapshot name to avoid Snapshot CR collision

derekbit:

Issue Longhorn/longhorn#13386

The volume-expansion system snapshot was named "expand-<size>", derived solely from the target size. Because Snapshot CRs are cluster-scoped in longhorn-manager, two volumes expanded to the same size produced the same snapshot name and collided.

Generate the name as `expand-<size>-<volumeName>` instead. The volume name is plumbed from the replica Server through to the Replica so the name stays deterministic across all replicas of a volume while being unique across volumes.

### DIFF
--- a/app/cmd/replica.go
+++ b/app/cmd/replica.go
@@ -130,7 +130,9 @@ func startReplica(c *cli.Context) (err error) {
 		}
 	}()
 
-	s := replica.NewServer(ctx, dir, backingFile, diskutil.ReplicaSectorSize, disableRevCounter, unmapMarkDiskChainRemoved, snapshotMaxCount, snapshotMaxSize, encrypted)
+	volumeName := c.GlobalString("volume-name")
+
+	s := replica.NewServer(ctx, dir, backingFile, diskutil.ReplicaSectorSize, disableRevCounter, unmapMarkDiskChainRemoved, snapshotMaxCount, snapshotMaxSize, encrypted, volumeName)
 
 	address := c.String("listen")
 
@@ -149,7 +151,6 @@ func startReplica(c *cli.Context) (err error) {
 		}
 	}
 
-	volumeName := c.GlobalString("volume-name")
 	replicaInstanceName := c.String("replica-instance-name")
 	dataServerProtocol := c.String("data-server-protocol")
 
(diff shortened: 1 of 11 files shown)


derekbit:
- e2e without the fix
  - https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8953/ `Test Volume Expansion Snapshot Name Not Colliding Across Volumes`: FAILED
  - https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8955/ `Test Volume Expansion Snapshot Name Contains Volume Name`: FAILED

- e2e with the fix
  - https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8952/ `Test Volume Expansion Snapshot Name Not Colliding Across Volumes`: PASS
  - https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8954/ `Test Volume Expansion Snapshot Name Contains Volume Name`:   PASS

Test cases: https://github.com/longhorn/longhorn-tests/pull/3658

apoorvajagtap (commented):
LGTM!
Just some questions:
1. The only scenario when an expansion snapshot will be created with `volumeName == ""` is during integration tests, right?
2. Is there any possibility that the complete name `expand-<size>-<volumeName>` could exceed the allowed character limits per DNS-1123? If so, might make sense to fallback again to `expand-<size>` format with a warning log?

derekbit:
> The only scenario when an expansion snapshot will be created with volumeName == "" is during integration tests, right?

1. Yes, you are right. IIRC, volumeName should not be an empty string. The if volumeName == "" check is a safeguard to ensure the expansion can still work and to cover any edge cases I may have missed.

> Is there any possibility that the complete name expand-<size>-<volumeName> could exceed the allowed character limits per DNS-1123? If so, might make sense to fallback again to expand-<size> format with a warning log?

The test case 
```
c.Assert(GenerateExpansionSnapshotName("", size), Equals, fmt.Sprintf(expansionSnapshotInfix, size))
```
is not testing the allowed character limits. It aims to ensure that when `volumeName` is empty, the function falls back to the size-only format to avoid generating a name with a trailing dash.

c3y1huang (approved):

apoorvajagtap (approved):

derekbit:
@mergify backport v1.12.x

PR #2000, opened 2026-09-01
feat: wait outside the controller lock before reporting the block frontend ready

davidcheng0922:

https://github.com/longhorn/longhorn/issues/13606

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -108,3 +108,5 @@ require (
 	k8s.io/utils v0.0.0-20260626114624-be93311217bd // indirect
 	sigs.k8s.io/json v0.0.0-20250730193827-2d320260d730 // indirect
 )
+
+replace github.com/longhorn/go-iscsi-helper => ../go-iscsi-helper
--- a/pkg/controller/control.go
+++ b/pkg/controller/control.go
@@ -49,6 +49,11 @@ type Controller struct {
 	snapshotMaxCount   int
 	SnapshotMaxSize    int64
 
+	// frontendLifecycleLock serializes frontend start/shutdown flows while the
+	// controller lock is released for iSCSI device readiness. Data-path READs do
+	// not take this lock, so scan READs can still be served during startup.
+	frontendLifecycleLock sync.Mutex
+
 	rebuildSyncConcurrentLimit int
 
 	GRPCAddress string
@@ -589,9 +594,34 @@ func (c *Controller) startFrontend() error {
 	return nil
 }
 
-func (c *Controller) StartFrontend(frontend string) error {
+func (c *Controller) StartFrontend(frontend string) (err error) {
+	c.frontendLifecycleLock.Lock()
+	defer c.frontendLifecycleLock.Unlock()
+
 	c.Lock()
-	defer c.Unlock()
+	defer func() {
+		var waitFrontend types.Frontend
+
+		// StartFrontend() can only set the frontend type when replicas are not
+		// attached yet. Wait only after the block frontend has actually started.
+		if c.frontend != nil && len(c.replicas) > 0 && c.frontend.FrontendName() == types.EngineFrontendBlockDev {
+			waitFrontend = c.frontend
+		}
+
+		if err != nil || waitFrontend == nil {
+			c.Unlock()
+			return
+		}
+		c.Unlock()
+
+		// Wait outside the controller lock for iSCSI scan/device readiness. The
+		// scan can issue READs that need Controller.ReadAt() to take RLock().
+		if err = waitFrontend.WaitForDeviceReady(); err != nil {
+			// Roll back the partial frontend start so the next retry runs readiness again.
+			rollbackErr := c.rollbackFrontendStart(waitFrontend, false)
+			err = types.CombineErrors(err, rollbackErr)
+		}
+	}()
 
 	if c.isExpanding {
 		return fmt.Errorf("cannot start frontend during the engine expansion")
@@ -914,9 +944,27 @@ func isBackendServiceUnavailable(errorCodes map[string]codes.Code) bool {
 	return false
 }
 
-func (c *Controller) Start(volumeSize, volumeCurrentSize int64, addresses ...string) error {
+func (c *Controller) Start(volumeSize, volumeCurrentSize int64, addresses ...string) (err error) {
+	c.frontendLifecycleLock.Lock()
+	defer c.frontendLifecycleLock.Unlock()
+
 	c.Lock()
-	defer c.Unlock()
+	var waitFrontend types.Frontend
+	defer func() {
+		if err != nil || waitFrontend == nil {
+			c.Unlock()
+			return
+		}
+		c.Unlock()
+
+		// Wait outside the controller lock for iSCSI scan/device readiness. The
+		// scan can issue READs that need Controller.ReadAt() to take RLock().
+		if err = waitFrontend.WaitForDeviceReady(); err != nil {
+			// Roll back the partial frontend start so the next retry runs readiness again.
+			rollbackErr := c.rollbackFrontendStart(waitFrontend, true)
+			err = types.CombineErrors(err, rollbackErr)
+		}
+	}()
 
 	log := logrus.WithField("volume", c.VolumeName)
 
@@ -1071,7 +1119,31 @@ func (c *Controller) Start(volumeSize, volumeCurrentSize int64, addresses ...str
 		}
 	}
 
-	return c.startFrontend()
+	if err := c.startFrontend(); err != nil {
+		return err
+	}
+	// Start() may run without a block frontend configured. Wait only after the
+	// block frontend has actually started.
+	if len(c.replicas) > 0 && c.frontend != nil && c.frontend.FrontendName() == types.EngineFrontendBlockDev {
+		waitFrontend = c.frontend
+	}
+	return nil
+}
+
+// rollbackFrontendStart cleans up a frontend startup after device readiness
+// fails. StartFrontend() rolls back only the frontend because replicas/backend
+// already existed, while Start() also rolls back the backend state it created.
+func (c *Controller) rollbackFrontendStart(frontend types.Frontend, rollbackBackend bool) error {
+	var rollbackErr error
+	if err := frontend.Shutdown(); err != nil {
+		rollbackErr = errors.Wrap(err, "failed to roll back frontend startup")
+	}
+	if rollbackBackend {
+		if err := c.shutdownBackend(); err != nil {
+			rollbackErr = types.CombineErrors(rollbackErr, errors.Wrap(err, "failed to roll back backend startup"))
+		}
+	}
+	return rollbackErr
 }
 
 func (c *Controller) WriteAt(b []byte, off int64) (int, error) {
@@ -1435,6 +1507,9 @@ func (c *Controller) shutdownFrontend() error {
 }
 
 func (c *Controller) ShutdownFrontend() error {
+	c.frontendLifecycleLock.Lock()
+	defer c.frontendLifecycleLock.Unlock()
+
 	if err := c.shutdownFrontend(); err != nil {
 		return err
 	}
@@ -1452,6 +1527,9 @@ func (c *Controller) shutdownBackend() error {
 }
 
 func (c *Controller) Shutdown() error {
+	c.frontendLifecycleLock.Lock()
+	defer c.frontendLifecycleLock.Unlock()
+
 	/*
 		Need to shutdown frontend first because it will write
 		the final piece of data to backend
--- a/pkg/frontend/tgt/frontend.go
+++ b/pkg/frontend/tgt/frontend.go
@@ -69,8 +69,25 @@ func (t *Tgt) Startup(rwu types.ReaderWriterUnmapperAt) error {
 		return err
 	}
 
-	t.isUp = true
+	// The block frontend is not reported as up until the iSCSI device is
+	// discovered and /dev/longhorn/<volume> is created.
+	if t.frontendName != types.EngineFrontendBlockDev {
+		t.isUp = true
+	}
 
+	return nil
+}
+
+func (t *Tgt) WaitForDeviceReady() error {
+	if t.dev == nil {
+		return nil
+	}
+	if err := t.dev.WaitForDeviceReady(); err != nil {
+		return err
+	}
+	// Report the block frontend as up only after the deferred readiness wait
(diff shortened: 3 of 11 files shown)


davidcheng0922:
All passed
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/11967/

app/cmd/controller.go
```diff
@@ -228,6 +229,15 @@ func startController(c *cli.Context) error {
...
 		}
+
+		if frontendName == types.EngineFrontendBlockDev {
+			// Wait outside the controller lock for iSCSI scan/device readiness.
+			// The scan can issue READs that need Controller.ReadAt() to take RLock().
+			if err := longhorndev.WaitForDeviceReady(volumeName); err != nil {
+				logrus.Error(err.Error())
+				os.Exit(1)
+			}
+		}
```
shuo-wu: I prefer not to operate the Controller's inner components like Frontend in this pkg. Can we move this wait to a defer function in `Controller.Start`?

go.mod
```diff
@@ -108,3 +108,5 @@ require (
 	k8s.io/utils v0.0.0-20260626114624-be93311217bd // indirect
 	sigs.k8s.io/json v0.0.0-20250730193827-2d320260d730 // indirect
 )
+
+replace github.com/longhorn/go-iscsi-helper => ../go-iscsi-helper
```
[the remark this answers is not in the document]
davidcheng0922: After review and merge, it will be removed

pkg/controller/control.go
```diff
@@ -589,9 +589,26 @@ func (c *Controller) startFrontend() error {
...
+		// attached yet. Wait only after the block frontend has actually started.
+		if frontend == nil || frontend.FrontendName() != types.EngineFrontendBlockDev || frontend.State() != types.StateUp {
+			c.Unlock()
+			return
+		}
+		c.Unlock()
+
+		// Wait outside the controller lock for iSCSI scan/device readiness. The
+		// scan can issue READs that need Controller.ReadAt() to take RLock().
+		err = frontend.WaitForDeviceReady()
```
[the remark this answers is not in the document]
davidcheng0922: Address, thank you

pkg/controller/control.go
```diff
@@ -589,9 +589,27 @@ func (c *Controller) startFrontend() error {
...
+
+		if err != nil || waitFrontend == nil {
+			c.Unlock()
+			return
+		}
+		c.Unlock()
+
+		// Wait outside the controller lock for iSCSI scan/device readiness. The
+		// scan can issue READs that need Controller.ReadAt() to take RLock().
+		err = waitFrontend.WaitForDeviceReady()
```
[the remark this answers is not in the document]
davidcheng0922: Fix, add `frontendLifecycleLock` which is only for frontend life cycle: `Start` and `Shuddown`

pkg/controller/control.go
```diff
@@ -914,9 +940,23 @@ func isBackendServiceUnavailable(errorCodes map[string]codes.Code) bool {
...
+	defer func() {
+		if err != nil || waitFrontend == nil {
+			c.Unlock()
+			return
+		}
+		c.Unlock()
+
+		// Wait outside the controller lock for iSCSI scan/device readiness. The
+		// scan can issue READs that need Controller.ReadAt() to take RLock().
+		err = waitFrontend.WaitForDeviceReady()
```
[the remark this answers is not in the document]
davidcheng0922: Make sense, add roll back mechanism for transient error

app/cmd/controller.go
davidcheng0922: Sure, thank you

vendor/github.com/longhorn/go-iscsi-helper/iscsidev/iscsi.go
```diff
@@ -176,14 +176,21 @@ func (dev *Device) StartInitator() error {
...
-	if err := iscsi.UpdateScsiDeviceTimeout(dev.KernelDevice.Name, dev.ScsiTimeout, dev.nsexec); err != nil {
+
+	return nil
+}
+
+// WaitForDeviceReady completes the initiator-side readiness work that can block
+// on SCSI scan/device discovery. Call it after the engine/controller startup
+// lock is released so scan READs can be served.
+func (dev *Device) WaitForDeviceReady() (err error) {
+	localIP, err := util.GetIPToHost()
```
derekbit, line 187: Do we need to hold the file lock
```
	lock := lhns.NewLock(LockFile, LockTimeout)
	if err := lock.Lock(); err != nil {
		return errors.Wrap(err, "failed to lock")
	}
	defer lock.Unlock()
```

vendor/github.com/longhorn/go-iscsi-helper/iscsidev/iscsi.go
```diff
@@ -176,14 +176,21 @@ func (dev *Device) StartInitator() error {
...
+	if err != nil {
 		return err
 	}
 
-	return nil
+	dev.KernelDevice, err = iscsi.WaitForDeviceReady(localIP, dev.Target, TargetLunID, dev.ScsiTimeout, dev.nsexec)
```
derekbit, line 192:
```suggestion
kernelDev, err := iscsi.WaitForDeviceReady(...)
if err != nil {
    return err
}
dev.KernelDevice = kernelDev
```

pkg/frontend/tgt/frontend.go
```diff
@@ -69,11 +69,28 @@ func (t *Tgt) Startup(rwu types.ReaderWriterUnmapperAt) error {
...
+func (t *Tgt) WaitForDeviceReady() error {
+	if t.dev == nil {
+		return nil
+	}
+	if err := t.dev.WaitForDeviceReady(); err != nil {
+		return err
+	}
+	// Report the block frontend as up only after the deferred readiness wait
+	// succeeds outside the controller lock.
+	t.isUp = true
```
derekbit, line 90: Should we change the data type of `isUp` to atomic.Bool?
`isUp` is not really protected by a lock.

derekbit:
@davidcheng0922 FYR, the regressions found by Opus 5.0.

# Review — PR #2000 (`abd4e6d`)

`feat(controller): wait for block frontend readiness after unlock`
Target issue: [longhorn/longhorn#13606](https://github.com/longhorn/longhorn/issues/13606)

The overall direction — deferring iSCSI device readiness until after the controller
write lock is released — is correct. The findings below are split into behaviour
that works on `master` and breaks with this change (regressions), and problems that
are new but not regressions.

---

## Regressions

### R1 — Live upgrade destroys the in-use device node (data-path risk)

On `master`, `createDev()` is only called from the `startScsiDevice(true)` branch.
The reload/upgrade branch deliberately does **not** call it, because the device node
already exists and is very likely mounted and in use.

This PR moves `createDev()` into `LonghornDevice.WaitForDeviceReady()`
(`vendor/github.com/longhorn/go-iscsi-helper/longhorndev/dev.go:201`), where it runs
unconditionally for `FrontendTGTBlockDev`. Since `d.deviceReady` is never set on the
reload path, a live upgrade now reaches `createDev()` (`dev.go:333`), which does:

```go
if _, err := os.Stat(dev); err == nil {
    logrus.Warnf("Device %s already exists, clean it up", dev)
    if err := util.RemoveDevice(dev); err != nil { ... }
}
```

i.e. it unlinks and recreates `/dev/longhorn/<volume>` under a live workload.

**Fix:** set `d.deviceReady = true` in the reload branch of `startScsiDevice()` so
`WaitForDeviceReady()` short-circuits during upgrade.

---

### R2 — Loss of cross-process serialization for `iscsiadm`

`StartInitator`, `ReloadInitiator`, `StopInitiator` and `RefreshInitiator` all acquire
`lhns.NewLock(LockFile, LockTimeout)` (flock on `/var/run/longhorn-iscsi.lock`) before
touching `iscsiadm`. All engine processes on a node share the same iSCSI initiator DB
and the same portal (`GetIPToHost():3260`), so `sendtargets` discovery from one volume
rewrites node records for every other volume on that portal — see the
`CleanupScsiNodes()` workaround and the "iSCSI database failure" comment in
`LogoutTarget()`.

`Device.WaitForDeviceReady()` (`iscsidev/iscsi.go:186`) takes no lock, yet it now runs
`getIscsiNodeSessionScanMode`, `IsTargetLoggedIn`, `manualScanSession`
(`iscsiadm --rescan`, a mutation) and `findScsiDevice` — all of which were inside the
lock on `master`.

Worst case is silent: `getIscsiNodeSessionScanMode()` (`iscsi/initiator.go:271`)
returns `scanModeAuto` for *any* output it does not recognise. A torn read of a node
record being rewritten by a concurrent discovery makes the manual rescan get skipped,
the device never appears, and attach fails after the retry budget.

**Fix:** hold the flock around the `iscsiadm` calls only (not around the sleeps/probes,
or a busy node will hit the 120s `LockTimeout`), and make
`getIscsiNodeSessionScanMode()` return an error when the output contains no
`node.session.scan` line at all.

---

### R3 — Attach can now fail on volumes that previously attached fine

`waitForBlockDeviceReadable()` (`longhorndev/dev.go:233`) issues a real 4 KiB `pread`
against `/dev/longhorn/<vol>` and fails the whole startup after 30 attempts. Two
problems:

- The comment claims a "non-blocking read", but `O_NONBLOCK` does not make `read(2)`
  non-blocking on a block device; the I/O goes all the way to the backend and can hang
  for the SCSI timeout.
- If the volume is degraded or the backend transiently returns `EIO`, attach now fails
  and rolls back. On `master` it would have succeeded.

Using data-path I/O as a readiness gate is a behavioural change, not just a stricter
check.

---

### R4 — `isUp` startup transition is no longer covered by the controller write lock

On `master`, every `isUp = true` write (`Init`, `Startup`, `Upgrade`) happens inside
`c.Lock()` via `startFrontend()`, so readers holding `c.RLock()` — e.g.
`Controller.Snapshot()` (`pkg/controller/control.go:225-231`) — can never observe the
transition.

This PR writes it from `Tgt.WaitForDeviceReady()` (`pkg/frontend/tgt/frontend.go:90`),
which by design runs *after* `c.Unlock()`. `Tgt` has no mutex of its own.

Note: the shutdown-side race (`Tgt.Shutdown()` writing `isUp` under `c.RLock()`, plus
the unlocked reads in `getVolume()`, `pkg/controller/rpc/server.go:88`) already exists
on `master`. This PR does not create the race class, it just adds a third
unsynchronized writer and opens a startup path that did not exist before.

**Fix:** make `isUp` an `atomic.Bool`.

---

### R5 — Rollback mutates controller state without the controller lock

`rollbackFrontendStart()` (`pkg/controller/control.go:1136`) calls `frontend.Shutdown()`
directly, bypassing `c.shutdownFrontend()` (`control.go:1498`), which deliberately takes
`c.RLock()` first ("Make sure writing data won't be blocked"). The two shutdown paths
are now inconsistent, and in-flight `ReadAt()` can be running against the backend while
the frontend socket server is torn down.

---

### R6 — Detach can be blocked for the full readiness budget, with no cancellation

`Shutdown()` and `ShutdownFrontend()` now contend for `frontendLifecycleLock`, which
`Start()` holds across the entire readiness wait: up to 10s of login polling + 10s in
`GetDevice` + 30s in `waitForBlockDeviceReadable`, plus per-`pread` SCSI timeouts. There
is no `context`/deadline, so a stuck attach cannot be interrupted by a detach — which is
the attach/detach loop symptom from longhorn/longhorn#13606.

---

### R7 — `StartFrontend()` type-change guard weakened

```go
if c.frontend.FrontendName() != frontend && c.frontend.State() != types.StateDown {
    return fmt.Errorf("frontend %v is already started, cannot be set as %v", ...)
}
```

A blockdev frontend that has been started but has not yet passed readiness — and one
that failed readiness and was rolled back — now reports `StateDown`, so this guard no
longer rejects switching to a different frontend type while a tgt target/session may
still exist. `isUp` is being overloaded to mean both "started" and "device ready"; these
need to be separate states.

---

### R8 — `go.mod` local `replace`

```
replace github.com/longhorn/go-iscsi-helper => ../go-iscsi-helper
```

Breaks any non-vendored build. Must be removed together with the `go.sum` deletion and
the `=> ../go-iscsi-helper` marker in `vendor/modules.txt` before merge. (Acknowledged
by the author.)

---

## Not regressions, but should be fixed in this PR

- **The original deadlock is not actually closed.** `LoginTarget()` is still called from
  `StartInitator()` inside `startFrontend()`, i.e. under `c.Lock()`. `--no_wait` only
  helps when it is supported; the synchronous fallback reproduces the exact deadlock
  this PR is meant to fix. `isNoWaitUnsupportedError()` also relies on matching
  `iscsiadm` error text, which varies by version and locale.
- **`dev.KernelDevice` is clobbered to `nil` on failure** in `iscsidev/iscsi.go:192`
  (`dev.KernelDevice, err = iscsi.WaitForDeviceReady(...)`). Assign only on success.
- **`Tgt.WaitForDeviceReady()` returns `nil` when `t.dev == nil`** without setting
  `isUp`, so `Start()` reports success while the frontend stays `Down` forever. This is
  a bug state and should return an error.
- **`Expand()` does not take `frontendLifecycleLock`.** `StartFrontend()` checks
  `c.isExpanding` under the lock and then releases it, so an expansion (and its iSCSI
  rescan) can interleave with the readiness scan.
- **`types.CombineErrors` uses `fmt.Errorf("%v, %v")`**, which breaks the error chain for
  `errors.Is/As`. Worth confirming the wrapped message cannot accidentally match the
  `ControllerErrorNoBackendReplicaError` string checks that longhorn-manager relies on.
- **Misleading comment** at `control.go:602`: "StartFrontend() can only set the frontend
  type when replicas are not attached yet" contradicts the `len(c.replicas) > 0`
  condition immediately below it.
- **No tests.** `pkg/controller/control_test.go` could cover, with a fake frontend:
  (a) a readiness failure rolls back and leaves the controller retryable, and
  (b) `ReadAt()` is not blocked while readiness is pending.

---

## Suggested order of work

1. Move `StartInitator()` (or at least `LoginTarget()`) out of the controller lock, or
   drop the synchronous fallback — this is what actually closes the deadlock.
2. R1: `deviceReady = true` on the reload branch.
3. R4: `isUp` → `atomic.Bool`.
4. R2: restore the flock around the `iscsiadm` calls.
5. `dev.KernelDevice`: assign only on success.
6. R6: thread a `context`/deadline through the readiness wait.
7. R8: drop the local `replace`; add tests.

PR #2001, opened 2026-09-01
chore(vendor): update longhorn dependencies

shuo-wu:

Issue longhorn/longhorn#13714

Although the ticket is for v2 engine only, it's better to update the vendors so that things are consistent in
longhorn-instance-manager level.


derekbit (approved):

derekbit:
@mergify backport v1.13.x

PR #2014, opened 2026-09-08
fix(replica): persist snapshot's own name in its disk meta file

KR-Ravindra:
Issue longhorn/longhorn#13728

## Problem

After taking a snapshot, the persisted `volume-snap-<name>.img.meta` file carries the previous head's filename in its `Name` field instead of the snapshot's own filename:

```
$ cat volume-snap-000.img.meta
{"Name":"volume-head-000.img","Parent":"","Removed":false,"UserCreated":true,...}
```

Every snapshot meta file written by `createDisk()` is affected.

## Root cause

`pkg/replica/replica.go`, `createDisk()` (line numbers on current master):

- L1019: `r.diskData[newSnapName] = r.diskData[oldHead]` reuses the old head's `disk` struct (pointer copy) for the new snapshot and updates `UserCreated`, `Created` and `Labels` on it, but not `Name`.
- L1030: `r.encodeToFile(r.diskData[newSnapName], newSnapName+".meta")` persists that struct while `Name` still reads `volume-head-NNN.img`.
- L1037: `r.activeDiskData[len(r.activeDiskData)-1].Name = newSnapName` corrects `Name` on the same struct, but only after the meta file has already been written.

The in-memory state ends up correct (the `activeDiskData` tail and `r.diskData[oldHead]` point at the same struct, and `readDiskData()` re-derives `Name` from the filename on reload), so only the on-disk meta file is wrong.

## Fix

Set `r.diskData[newSnapName].Name = newSnapName` together with the other field updates, right after the pointer copy and before `encodeToFile()`. The later `activeDiskData[...].Name = newSnapName` assignment becomes redundant (same struct) and is removed.

Because that struct is still the current head's until the snapshot completes, `createDisk()` now keeps a copy of it (`oldHeadDisk`) and restores it in the existing rollback closure that removes the new map keys. If the snapshot meta write fails (for example ENOSPC), the head keeps its own `Name` instead of the just-removed `volume-snap-<name>.img`, and the `UserCreated`/`Created`/`Labels` values set for the snapshot no longer leak onto the head either.

Existing meta files written by earlier versions are not rewritten by this change; they are still loaded correctly because `readDiskData()` takes the name from the filename.

## How tested

Added two tests in `pkg/replica/replica_test.go`:

- `TestSnapshotDiskMetaName`: takes two snapshots on a temp-dir replica, then for each snapshot disk and the head reads back `<disk>.meta` and asserts `Name` equals the disk filename (and that `r.diskData[<disk>].Name` matches).
- `TestSnapshotMetaWriteFailureRollback`: creates a directory at `volume-snap-001.img.meta.tmp` so `encodeToFile()` fails for the snapshot meta, asserts `Snapshot()` returns the tmp-file error, that the head's in-memory `disk` struct is byte-for-byte what it was before the call (`Name`, `UserCreated`, `Created`, `Labels`), that `activeDiskData`'s tail and `ListDisks()` are unchanged and the failed snapshot's files are gone, and that a following `Snapshot("002")` succeeds with correct `Name` values in every meta file. Without the restore in the rollback this test fails on the head's `Name` (`volume-snap-001.img` instead of `volume-head-001.img`).

Before the fix:

```
FAIL: replica_test.go:203: TestSuite.TestSnapshotDiskMetaName

replica_test.go:229:
    c.Assert(d.Name, Equals, name)
... obtained string = "volume-head-000.img"
... expected string = "volume-snap-000.img"

OOPS: 0 passed, 1 FAILED
FAIL	github.com/longhorn/longhorn-engine/pkg/replica	2.015s
```

After the fix:

```
ok  	github.com/longhorn/longhorn-engine/pkg/replica	5.019s   (-check.f 'TestSnapshotDiskMetaName$|TestSnapshotMetaWriteFailureRollback$')
ok  	github.com/longhorn/longhorn-engine/pkg/replica	43.529s  (full package)
```

`go build ./...`, `go vet ./pkg/replica/`, `gofmt -l` and `golangci-lint run ./pkg/replica/...` are clean.

- Per the discussion on the issue, the old (wrong) `Name` value incidentally recorded which head the snapshot was taken from. This PR only corrects the field; persisting snapshot creation-order information explicitly, and back-filling existing replicas' meta files, are left as follow-ups if wanted.
- No behavior change for reads/writes or for reload: `readDiskData()` already overrides `Name` with the filename-derived value.
- The rollback restore is new behavior only on the failure path: previously a failed snapshot meta write left the head struct with the snapshot's `UserCreated`/`Created`/`Labels`; now it is restored completely.

## Links

- https://github.com/longhorn/longhorn/issues/13728

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.

### DIFF
--- a/pkg/replica/replica.go
+++ b/pkg/replica/replica.go
@@ -1016,7 +1016,11 @@ func (r *Replica) createDisk(name string, userCreated bool, created string, labe
 	r.diskData[newHeadDisk.Name] = &newHeadDisk
 	if newSnapName != "" {
 		r.addChildDisk(newSnapName, newHeadDisk.Name)
+		// The old head's disk struct is reused for the new snapshot. Keep a copy so the rollback
+		// can restore the head if a later step fails.
+		oldHeadDisk := *r.diskData[oldHead]
 		r.diskData[newSnapName] = r.diskData[oldHead]
+		r.diskData[newSnapName].Name = newSnapName
 		r.diskData[newSnapName].UserCreated = userCreated
 		r.diskData[newSnapName].Created = created
 		r.diskData[newSnapName].Labels = labels
@@ -1024,6 +1028,7 @@ func (r *Replica) createDisk(name string, userCreated bool, created string, labe
 			delete(r.diskData, newHeadDisk.Name)
 			delete(r.diskData, newSnapName)
 			delete(r.diskChildrenMap, newSnapName)
+			*r.diskData[oldHead] = oldHeadDisk
 			return nil
 		})
 
@@ -1034,7 +1039,6 @@ func (r *Replica) createDisk(name string, userCreated bool, created string, labe
 		rollbackFuncList = append(rollbackFuncList, snapMetaEncodeRollbackFunc)
 
 		r.updateChildDisk(oldHead, newSnapName)
-		r.activeDiskData[len(r.activeDiskData)-1].Name = newSnapName
 	}
 	delete(r.diskData, oldHead)
 
--- a/pkg/replica/replica_test.go
+++ b/pkg/replica/replica_test.go
@@ -200,6 +200,95 @@ func (s *TestSuite) TestSnapshot(c *C) {
 	c.Assert(disks["volume-head-002.img"].Size, Equals, "0")
 }
 
+func (s *TestSuite) TestSnapshotDiskMetaName(c *C) {
+	dir, err := os.MkdirTemp("", "replica")
+	c.Assert(err, IsNil)
+	defer func() {
+		errRemove := os.RemoveAll(dir)
+		c.Assert(errRemove, IsNil)
+	}()
+
+	r, err := New(context.Background(), 9, 3, dir, nil, false, false, 250, 0, false, false, types.ReplicaStateInitial, 9, "")
+	c.Assert(err, IsNil)
+	defer func() {
+		errClose := r.Close()
+		c.Assert(errClose, IsNil)
+	}()
+
+	err = r.Snapshot("000", true, getNow(), nil)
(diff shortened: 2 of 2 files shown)


KR-Ravindra:
Round 1 self-review (AI agent operated by me; I review every revision). Staying a draft until these are done:

1. Keep the existing `r.activeDiskData[len(r.activeDiskData)-1].Name = newSnapName` line rather than deleting it. It is only redundant if the tail of `activeDiskData` and `r.diskData[oldHead]` are always the same pointer; that holds for `createDisk` and `openLiveChain`, but `diskData` entries are reassigned elsewhere (replica.go:867, :1089) and the assignment costs nothing. Removing it widens the change without helping the fix.
2. Drop the `Claude-Session:` and `Co-Authored-By:` trailers from the commit; keep only the `Longhorn 13728` trailer and the DCO sign-off.

Verified against `master`: the pointer copy at :1019, the meta write at :1030 and the late `Name` fix at :1037 are as described, and `readDiskData` overrides `Name` from the filename on reload. No other PR references longhorn/longhorn#13728; shuo-wu's note about back-filling existing meta files and recording creation order is left as a follow-up, as the body says.

KR-Ravindra:
Self-review before marking ready (an AI agent I operate prepared this; I reviewed the diff and the tests). Two things worth a look: the rollback path now restores the old head's disk struct if the snapshot meta write fails, with a test that forces that failure; and the PR deliberately does not back-fill `Name` in existing replicas' meta files nor persist creation order, which shuo-wu raised on the issue and which I would handle as follow-ups if you want them.

PR #2019, opened 2026-09-10
fix(rebuild): race condition when creating a rebuilding goroutine

mantissahz:

Issue # longhorn/longhorn#13914

Check if replica is in mode WO after using the API `ReplicaCreate(address, true, types.WO)`.

### DIFF
--- a/integration/core/test_cli.py
+++ b/integration/core/test_cli.py
@@ -272,6 +272,116 @@ def test_replica_add_after_rebuild_failed(bin, grpc_controller_client,  # NOQA
         assert r.mode == 'RW'
 
 
+def add_replica_and_wait(bin, grpc_controller_client, replica_url):  # NOQA
+    cmd = [bin, '--debug', '--url', grpc_controller_client.address,
+           'add-replica',
+           '--size', SIZE_STR,
+           '--current-size', SIZE_STR,
+           replica_url]
+    subprocess.check_call(cmd)
+    wait_for_rebuild_complete(grpc_controller_client.address)
+
+
+def get_snapshot_names(bin, grpc_controller_client):  # NOQA
+    cmd = [bin, '--debug', '--url', grpc_controller_client.address,
+           'snapshot', 'info']
+    return sorted(json.loads(subprocess.check_output(cmd)).keys())
+
+
+def test_replica_add_rw_replica_is_inert(bin,  # NOQA
+                                         grpc_controller_client,  # NOQA
+                                         grpc_replica_client,  # NOQA
+                                         grpc_replica_client2):  # NOQA
+    # longhorn/longhorn#13914: longhorn-manager can leave a stale
+    # rebuilding goroutine running after the rebuild it was spawned for
+    # already succeeded, so it issues add-replica against a replica that
+    # is now RW.
+    #
+    # The engine rejects that request rather than reporting success,
+    # because nothing was rebuilt -- a success would make the engine
+    # controller emit a "has been rebuilt" event and run a post-rebuild
+    # snapshot purge for a rebuild that never happened. What has to hold
+    # is that the rejection is inert: the healthy replica survives it
+    # untouched. longhorn-manager recognises this specific message in
+    # isReplicaNotInWORWMode() and skips handleRebuildFailure(), which
+    # is what stops the replica from being removed and marked failed.
+    open_replica(grpc_replica_client)
+    open_replica(grpc_replica_client2)
+
+    add_replica_and_wait(bin, grpc_controller_client,
+                         grpc_replica_client.url)
+    add_replica_and_wait(bin, grpc_controller_client,
+                         grpc_replica_client2.url)
+    verify_replica_mode(grpc_controller_client,
+                        grpc_replica_client2.url, "RW")
+
+    snapshots_before = get_snapshot_names(bin, grpc_controller_client)
+
+    # the stale rebuild request
+    cmd = [bin, '--debug', '--url', grpc_controller_client.address,
+           'add-replica',
+           '--size', SIZE_STR,
+           '--current-size', SIZE_STR,
+           grpc_replica_client2.url]
+    proc = subprocess.run(cmd, capture_output=True, encoding='utf-8')
+    assert proc.returncode != 0
+    # longhorn-manager matches on this wording to tell a stale rebuild
+    # apart from a real failure, so it is part of the engine's contract
+    assert "is not in mode WO: RW" in proc.stderr
+
+    volume = grpc_controller_client.volume_get()
+    assert volume.replicaCount == 2
+
+    replicas = grpc_controller_client.replica_list()
+    assert len(replicas) == 2
+    assert sorted(r.address for r in replicas) == \
+        sorted([grpc_replica_client.url, grpc_replica_client2.url])
+    for r in replicas:
+        assert r.mode == 'RW'
+
+    # the rejected add must not have started a rebuild, which would
+    # leave behind another system-generated snapshot
+    assert get_snapshot_names(bin, grpc_controller_client) == snapshots_before
+
+
+def test_replica_create_existing_rw_replica(bin,  # NOQA
+                                            grpc_controller_client,  # NOQA
+                                            grpc_replica_client,  # NOQA
+                                            grpc_replica_client2):  # NOQA
+    # Pins the controller behaviour the longhorn/longhorn#13914 fix in
+    # sync.AddReplica depends on: ControllerReplicaCreate for an address
+    # the controller already holds is a no-op that reports the replica's
+    # *current* mode rather than the requested one. sync.AddReplica uses
+    # that reported mode as its only signal that the rebuild it was
+    # asked to start has already happened. If this ever returns WO, or
+    # errors, or re-adds the replica, that guard stops working and the
+    # stale-rebuild path in #13914 silently comes back.
+    open_replica(grpc_replica_client)
+    open_replica(grpc_replica_client2)
+
+    add_replica_and_wait(bin, grpc_controller_client,
+                         grpc_replica_client.url)
+    add_replica_and_wait(bin, grpc_controller_client,
+                         grpc_replica_client2.url)
+    verify_replica_mode(grpc_controller_client,
+                        grpc_replica_client2.url, "RW")
+
+    snapshots_before = get_snapshot_names(bin, grpc_controller_client)
+
+    r = grpc_controller_client.replica_create(
+        address=grpc_replica_client2.url, snapshot_required=True, mode='WO')
+    assert r.address == grpc_replica_client2.url
+    assert r.mode == 'RW'
+
+    replicas = grpc_controller_client.replica_list()
+    assert len(replicas) == 2
+    for replica in replicas:
+        assert replica.mode == 'RW'
+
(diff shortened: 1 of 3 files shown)


pkg/sync/sync.go
```diff
@@ -462,11 +462,16 @@ func (t *Task) AddReplica(volumeSize, volumeCurrentSize int64, address, instance
...
 
 	logrus.Infof("Adding replica %s in WO mode", address)
-	_, err = t.client.ReplicaCreate(address, true, types.WO)
+	controllerReplicaInfo, err := t.client.ReplicaCreate(address, true, types.WO)
 	if err != nil {
 		return err
 	}
 
+	if controllerReplicaInfo != nil && controllerReplicaInfo.Mode == types.RW {
+		logrus.Infof("Replica %s is aleady in RW mode", address)
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks

pkg/sync/sync.go
```diff
@@ -462,11 +462,16 @@ func (t *Task) AddReplica(volumeSize, volumeCurrentSize int64, address, instance
...
+		return nil
```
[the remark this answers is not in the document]
mantissahz: The error will be handled by the engine controller in longhorn-manager.
And if the replica is still in mode WO, the rebuilding goroutine will fail at the purging snapshots stage.

derekbit:
@mantissahz Is it ready for review?

pkg/types/error.go
```diff
@@ -21,6 +21,10 @@ const (
 	CannotRequestHashingSnapshotPrefix = "cannot request hashing snapshot"
 
 	ErrorStringNoSpaceLeftOnDevice = "no space left on device"
+
+	ErrorStringReplicaAddress     = "replica %s"
+	ErrorStringNotInModeWO        = "is not in mode WO:"
+	ErrorStringReplicaNotInWOMode = ErrorStringReplicaAddress + " " + ErrorStringNotInModeWO + " %s"
```
derekbit: How about
```
const (
	ErrorStringReplicaAddress     = "replica %s"
	ErrorStringNotInModeWO        = "is not in mode WO:"
)

func ErrReplicaNotInWOMode(replicaAddr string, currentMode string) error {
	return fmt.Errorf("replica %s is not in mode WO: %s", replicaAddr, currentMode)
}
```

integration/core/test_cli.py
```diff
@@ -272,6 +272,100 @@ def test_replica_add_after_rebuild_failed(bin, grpc_controller_client,  # NOQA
...
+    add_replica_and_wait(bin, grpc_controller_client,
+                         grpc_replica_client2.url)
+    verify_replica_mode(grpc_controller_client,
+                        grpc_replica_client2.url, "RW")
+
+    snapshots_before = get_snapshot_names(bin, grpc_controller_client)
+
+    # the stale rebuild request
+    add_replica_and_wait(bin, grpc_controller_client,
+                         grpc_replica_client2.url)
```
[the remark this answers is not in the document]
mantissahz: Updated. Thanks.

pkg/types/error.go
mantissahz: Modified. Thanks.

davidcheng0922 (dismissed):
LGTM

pkg/sync/sync.go
```diff
@@ -462,11 +462,15 @@ func (t *Task) AddReplica(volumeSize, volumeCurrentSize int64, address, instance
...
 	logrus.Infof("Adding replica %s in WO mode", address)
-	_, err = t.client.ReplicaCreate(address, true, types.WO)
+	controllerReplicaInfo, err := t.client.ReplicaCreate(address, true, types.WO)
 	if err != nil {
 		return err
 	}
 
+	if controllerReplicaInfo != nil && controllerReplicaInfo.Mode == types.RW {
+		return types.ErrReplicaNotInWOMode(address, string(types.RW))
+	}
```
derekbit: Why do we need to check WO earlier? Isn't the error message returned in the following operations?

derekbit: I meant the check is redundant because it is already enforced in `getTransferClients`.

c3y1huang (commented):
Comment moved to https://github.com/longhorn/longhorn-engine/pull/2019#pullrequestreview-5174980707. Deferring auto-merge to discuss the direction of the fix.

c3y1huang (changes requested):
IIUC, the cause is a timing mismatch in longhorn-manager:
1. Engine controller spawns goroutine 1 to rebuild replica
1. Goroutine 1 blocks in the pre-rebuild snapshot purge before it calls `ReplicAdd` -> wait 30s for replica to appear in the engine -> timed out -> return err -> requeued.
1. Next reconcile -> still no replica in engine -> engine controller spawns goroutine 2
1. Goroutine 2 -> wait for snapshot purge
1. purge completes -> both goroutine 1 & 2 called `ReplicaAdd` for the same replica.
1. The one that arrives second finds the replica already exists -> report failure -> removed replica the first one rebuilt.

If that is the case, this PR seems to be treating the symptom (step 6) not the cause (step 4), so goroutine 2 is still created. Also, it only covers RW case, I believe WO has the same trigger and ending.

IMO the fix belongs in longhorn-manager at step 4. The controller should skip when the rebuild is already inflight instead of spawning a second+ goroutine. That removes the duplications rather than trying to survive it.

mantissahz:
> If that is the case, this PR seems to be treating the symptom (step 6) not the cause (step 4), so goroutine 2 is still created. Also, it only covers RW case, I believe WO has the same trigger and ending.

Yes, it might trigger it again, but it should finally complete the rebuild.

> IMO the fix belongs in longhorn-manager at step 4. The controller should skip when the rebuild is already inflight instead of spawning a second+ goroutine. That removes the duplications rather than trying to survive it.

Yes, my first idea is to prevent the second goroutine from being created, and that might be a big scale to modify.
Therefore, we leave it as an improvement https://github.com/longhorn/longhorn/issues/13992 after discussing with @derekbit.
This PR mitigates the issue.

pkg/sync/sync.go
mantissahz: Removed it. Thanks.

c3y1huang:
> Yes, it might trigger it again, but it should finally complete the rebuild.

That is actually the concern. This is what I think it looks like:
1. Goroutine 1 copying data into new replica. Replica is WO.
1. Goroutine 2 arrives -> asks engine to add replica -> continues because replica is WO.
1. Goroutine 2 calls PrepareRebuildReplica on the replica goroutine 1 is halfway through filling.
1. Goroutine 1 finishes, flips replica to RW.
1. Goroutine 2 tries to continue -> fails because replica is now RW -> reports failure.
1. The replica goroutine 1 just finished rebuilding gets deleted
1. Rebuild starts over.

So it does get there in the end, but only after throwing away a finished one. Isn't that the loop longhorn/longhorn#13914 describes?

> Yes, my first idea is to prevent the second goroutine from being created, and that might be a big scale to modify.

Maybe I am missing something, but I pictured it fairly small:
- a new map EngineController to track which engines have build in flight
- a check in startRebuilding before prepareRebuildContext that returns early if the engine is already in the map
- a defer at the top of runRebuild to remove it from the map.

Just my 2 cents.

mantissahz:
After discussing with @shuo-wu and @derekbit, we decided to return an error directly in the function `canAdd` on the backend when creating a replica using the API `ControllerReplicaCreate`.

Regression tests look good for basic and HA tests:
- https://ci.longhorn.io/job/private/job/longhorn-tests-regression/12503/
- https://ci.longhorn.io/job/private/job/longhorn-tests-regression/12515/

pkg/controller/rpc/server.go
```diff
@@ -217,7 +217,7 @@ func (cs *ControllerServer) ReplicaGet(ctx context.Context, req *enginerpc.Repli
 }
 
 func (cs *ControllerServer) ControllerReplicaCreate(ctx context.Context, req *enginerpc.ControllerReplicaCreateRequest) (*enginerpc.ControllerReplica, error) {
-	if err := cs.c.AddReplica(req.Address, req.SnapshotRequired, types.GRPCReplicaModeToReplicaMode(req.Mode)); err != nil {
+	if err := cs.c.AddReplica(req.Address, req.SnapshotRequired, true, types.GRPCReplicaModeToReplicaMode(req.Mode)); err != nil {
```
derekbit, line 220: When is `AddReplica` with `creatingReplica=false`?

pkg/controller/control.go
```diff
@@ -147,8 +147,8 @@ func (c *Controller) WaitForShutdown() error {
 	return c.lastError
 }
 
-func (c *Controller) AddReplica(address string, snapshotRequired bool, mode types.Mode) error {
-	return c.addReplica(address, snapshotRequired, mode)
+func (c *Controller) AddReplica(address string, snapshotRequired, creatingReplica bool, mode types.Mode) error {
```
derekbit, line 150: I don't see any case that `AddReplica` with `creatingReplica=false` in this PR. Why do we need the new parameter?

mantissahz: line 462 will use `AddReplica` with `creatingReplica=false` in this PR.

pkg/controller/rpc/server.go
mantissahz: Line 462 in this PR will use `AddReplica` with `creatingReplica=false`.

pkg/controller/control.go
```diff
@@ -160,8 +160,11 @@ func (c *Controller) hasWOReplica() bool {
 	return false
 }
 
-func (c *Controller) canAdd(address string) (bool, error) {
+func (c *Controller) canAdd(address string, creatingReplica bool) (bool, error) {
 	if c.hasReplica(address) {
+		if creatingReplica {
+			return false, fmt.Errorf(types.ErrorStringReplicaAddressExist+" %v", address)
+		}
```
derekbit, line 167: Why do we need creatingReplica? Could we always return `fmt.Errorf(types.ErrorStringReplicaAddressExist+" %v", address)` and let the caller decide how to proceed?

mantissahz: Now `canAdd` is called by `addReplica` and `addReplicaNoLock`.
If we decide to call `canAdd` without `creatingReplica`, we need to handle it in `addReplicaNoLock` and check if the returned error is a false alarm (a replica might change the mode from WO to RW).
It will be a bit more complicated than adding a parameter.
Also, for other functions that will call `canAdd` in the future.

But I am fine with either way.

pkg/controller/control.go
```diff
@@ -160,8 +160,11 @@ func (c *Controller) hasWOReplica() bool {
...
 		return false, nil
 	}
```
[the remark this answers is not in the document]
mantissahz: `creatingReplica==True` will only be used when creating a replica.
And this replica should not be in ERR Mode.

mantissahz:
The rebuilding replica map in the control plane, as we discussed before, can eliminate the root cause that recreates unnecessary rebuild goroutines. I have updated the longhorn-manager PR.
We can keep this engine-gouard PR because it already solves the issue and will prevent similar scenarios from using the `ReplicaAdd` API in the future.

WDYT @derekbit, @c3y1huang, @shuo-wu?

